PR #2, opened 2026-05-30
Scaffold npm release pipeline (OIDC trusted publishing, monorepo-wide)

YourWildDad:
Lays the npm publishing pipeline for the monorepo's packages **before** they are
re-homed. Workflow scaffolding only — no package builds yet. (Generalized from
the original feishu-transport-only version: publishes every `shouldPublish`
package, not just one.)

## Design — two workflows, both monorepo-wide

Zero long-lived npm token: npm 11.5.1+ exchanges the GitHub Actions OIDC
id-token for a short-lived publish token and stamps build provenance.

**Verified against Rush 5.140.0:** `rush publish` has **no `--provenance`** and
authenticates **only** via a token, so it cannot do tokenless OIDC. The
responsibilities are split:

| Workflow | Role |
|---|---|
| **`release.yml`** | The OIDC publish — and the single file **every** package registers as its trusted publisher. A plan step classifies each `rush list --json` `shouldPublish` project against npm (see below), then **loops** `npm publish --provenance --access public` over the publish set (one OIDC exchange per package). `id-token: write`, `setup-node` `registry-url`, `npm@11.5.1`, **no `NODE_AUTH_TOKEN`**. |
| **`version.yml`** | Rush-native replacement for `changeset version`. `rush publish --apply` bumps every changed package and opens a **version PR** (not a push to protected `main`). |

**Rush owns versioning; npm owns the OIDC upload.** One run can publish N
packages — npm validates each publish's OIDC token against that package's own
trusted-publisher config. Rationale + alternatives in
`.agents/decisions/0005-npm-release-oidc.md`.

## Merge is always safe — the pipeline self-adapts (this round's fix)

The plan step classifies each publishable package into **three states** so a
push/merge never fails just because a package isn't ready:

| State (`npm view`) | Meaning | Action |
|---|---|---|
| `name@version` exists | already published | skip |
| `name` exists, `@version` 404s | new version of an existing package | **publish via OIDC** |
| `name` itself 404s | never published | **skip + `::warning::` (needs bootstrap)** |
| any **non-E404** error (network / 5xx / auth) | flaky lookup | **fail the job loudly** — never mistaken for "not published" |

So merging this PR is green by default. `@excitedjs/dreamux` (`shouldPublish`,
not on npm yet) is **skipped with a warning**, not a doomed publish — and it's
picked up automatically once bootstrapped. (No `shouldPublish: false` toggle
needed; that would be wrong anyway — it also drops the package from `version.yml`
bumping.) Classification reads the registry only, so a *misconfigured* trusted
publisher isn't confused with *absent*: the package exists, so it's queued and
its `npm publish` fails loudly — the correct signal.

> Verified against the real registry: `@excitedjs/dreamux@0.1.0` → absent,
> `@excitedjs/tm@1.1.0` → published, `@excitedjs/tm@9.9.9` → new-version, and an
> unreachable registry → `ECONNREFUSED` correctly **not** treated as a 404.

## Required setup — only you (npm account owner) can do this

**1. First publish of each package is a one-time token bootstrap, not OIDC.**
npm's trusted-publisher config lives on a package's settings page, which exists
only **after** the package is published once. So each package's *first* publish
is a manual `npm publish` with a token to create it; then configure the trusted
publisher and `release.yml` (pure OIDC) takes over from the next version.
(Confirmed via claudemux: `@excitedjs/tm`'s only version carries **no
provenance** despite its `--provenance` workflow, and that OIDC workflow landed
*after* the package's first publish.)

**2. Trusted publisher per package** — for **each** package, on **npmjs.com** →
package → Settings → Trusted Publisher:

| Field | Value |
|---|---|
| Provider | **GitHub Actions** |
| Organization / owner | `excitedjs` |
| Repository | `excitedjs/dreamux` |
| Workflow filename | **`release.yml`** (same for every package) |
| Environment | *(leave blank)* |

## Notes for the reviewer

- **Unexecutable end-to-end until a package builds** — validated offline: both
  workflows pass YAML + **`actionlint`** (shellcheck on the `run:` scripts); the
  plan's `rush list --json` → TSV one-liner and the three-state npm
  classification were both run against the real repo/registry; `rush publish
  --apply` confirmed to no-op on an empty `common/changes/`.
- **Per-package readiness** (in `release.yml` + ADR): `shouldPublish: true`;
  a build + `files` allowlist (`prepack` recommended); a `repository` field
  (provenance requires it); no `workspace:` protocol deps (`npm publish` doesn't
  rewrite them).
- **Optional hardening (your call):** commented `environment: npm-publish` block
  — gate real publishes behind required reviewers; if enabled it must match the
  npm trusted-publisher Environment field of every package.

Renamed from `feishu-transport-release.yml` / `feishu-transport-version.yml` to
the monorepo-level `release.yml` / `version.yml`. Rebased on `main` (guardrail
PR #1 is merged). Public-repo rules kept: GitHub noreply identity, no internal
content, no co-author trailer.

Opened for your review — **not merged.**

### DIFF
--- a/.agents/decisions/0005-npm-release-oidc.md
+++ b/.agents/decisions/0005-npm-release-oidc.md
@@ -1,0 +1,108 @@
+# 0005 — npm release via OIDC trusted publishing
+
+- **Status:** Accepted (scaffolding; unexecutable until a publishable package builds)
+- **Date:** 2026-05-30
+- **Affects:** npm publishing of every `shouldPublish` package, `/.github/workflows/`, the rush version mechanism
+- **PR / Issue:** infra/feishu-transport-release
+
+## Context
+
+The monorepo publishes more than one package to the public npm registry (the
+`@excitedjs/dreamux` CLI today; `@excitedjs/feishu-transport` and others as they
+are re-homed). We want **zero long-lived npm tokens**: npm's OIDC trusted
+publishing (GA 2025-07-31) lets a GitHub Actions job exchange its short-lived
+OIDC id-token for a publish token, and stamps build **provenance**. The
+reference flow exists in the sibling **claudemux** repo, but it is pnpm +
+changesets; dreamux is a Rush + pnpm monorepo, so the version+publish half has
+to be rush-native — and the publish must span every publishable package.
+
+Three hard constraints shaped the design:
+
+- **`rush publish` cannot do OIDC.** Confirmed against Rush 5.140.0:
+  `rush publish` has no `--provenance` flag and authenticates only via
+  `--npm-auth-token` / `common/config/rush/.npmrc-publish` (a token). It cannot
+  do tokenless OIDC or emit provenance.
+- **npm trusted publishing is per-package; `npm publish` is per-package.** Each
+  package is configured on npmjs.com individually. But one workflow *run* can
+  publish many packages: each `npm publish` does its own OIDC exchange, so every
+  package just lists the same workflow file as its trusted publisher.
+- **Main is protected.** The anti-leak guardrail (decision 0004) makes the
+  `gitleaks` check required on `main`, so CI cannot push a version-bump commit
+  straight to `main` with the default `GITHUB_TOKEN`.
+
+## Decision
+
+Two workflows, both `push: [main]`:
+
+- **`/.github/workflows/release.yml`** — the OIDC publish, and the single
+  workflow every publishable package registers as its trusted publisher. A plan
+  step enumerates `rush list --json` projects with `shouldPublish: true` and
+  **classifies each against npm into three states**: *published* (skip),
+  *new version of an existing package* (publish), or *never published* (skip +
+  warn — its first publish needs a one-time token bootstrap, see below). It then
+  loops `npm publish --provenance --access public` over the publish set (one
+  OIDC exchange per package). A non-404 lookup error (network / 5xx / auth)
+  fails the job loudly rather than being read as "not published". `id-token:
+  write`, `setup-node` `registry-url`, `npm install -g npm@11.5.1`, no
+  `NODE_AUTH_TOKEN`. No-ops (green) when nothing is ready to publish.
+- **`/.github/workflows/version.yml`** — the rush-native replacement for
+  `changeset version`. On merge to `main` it runs `rush publish --apply`
+  (consumes `common/changes/*` into package.json + CHANGELOG bumps across all
+  changed packages; no registry contact, no token) and opens a "version
+  packages" PR rather than pushing to protected `main`.
+
+So **Rush owns versioning, npm owns the OIDC upload**, and both halves are
+monorepo-wide.
+
+## Consequences
+
+- **Required setup the npm account owner must do** (per package, cannot be done
+  from the repo): for *each* package this pipeline should publish, register a
+  trusted publisher on npmjs.com — owner `excitedjs`, repository
+  `excitedjs/dreamux`, workflow **`release.yml`** (the same file for every
+  package), environment blank. Adding a package later = a rush.json entry + one
+  npm entry; no workflow change.
+- **Any push/merge is always safe — a never-published package is skipped, not
+  failed.** The plan step's three-state classification means a `shouldPublish`
+  package that does not exist on npm yet (e.g. `@excitedjs/dreamux`,
+  `shouldPublish: true`, v0.1.0, currently 404) is **skipped with a warning**,
+  not pushed through a doomed `npm publish`. So merging this PR — and every
+  later push — is green by default. The package is picked up automatically once
+  it has been bootstrapped (see next bullet); no `shouldPublish: false` toggle
+  is needed (and that toggle would be wrong anyway — it also drops the package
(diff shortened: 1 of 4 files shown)


YourWildDad:
Note: guardrail PR #1 merged into `main` while this was in flight. Rebased this branch onto it — the only conflict was the `.agents/root.md` "When to read which" table (the 0004 and 0005 rows landed adjacent); resolved keeping both in order. PR diff is now just the 4 new files.

This makes the version-PR design concretely (not hypothetically) justified: the `gitleaks` check is now live on `main` and ran green on this PR, so the rush version step opens a PR rather than pushing to a protected `main`.

CI on this PR: `gitleaks`, `knowledge base`, `package`, `rush` all green. The two new `feishu-transport-*` workflows are `push: [main]`-triggered, so they intentionally don't run on this PR (and no-op until the package lands).

YourWildDad (commented):
**LGTM — adversarially reviewed, no blocking issues.** The tri-state classifier is the right shape and I verified the load-bearing claims empirically rather than on inspection.

### What I verified (not just read)

**① Tri-state classification only treats a *true* E404 as "never published".** Reproduced on npm 10.9.2 — the exact npm the `plan` step runs on (the `npm@11.5.1` upgrade happens in a *later* step):
- pkg missing → `npm view name` exits 1 + `E404` → `absent` (skip + warn) ✓
- pkg exists, exact version missing → `npm view name@ver` exits 1 + `404 No match found for version` (contains `E404`) → `new-version` (queue) ✓ — this is the linchpin, and it holds: a genuinely new version is **not** misread as `published`.
- pkg + version exist → exit 0 → `published` (skip) ✓
- `E401` / `ENOTFOUND` / 5xx → output does **not** contain the substring `E404` → `return 1` → job fails loudly. Confirmed these strings don't false-match the `grep -q 'E404'`. So a flaky/auth/5xx lookup is **never** swallowed as skip. The concern about conflating transient failure with "absent" does not occur.

**② Multi-package loop is wired correctly.** `rush list --json` really emits `{name, version, path, shouldPublish}` (verified against Rush 5.140.0 in-repo); `path` is repo-relative (`packages/dreamux`), so `cd "$path" && npm publish` resolves from the checkout root. One `rush install` + one `rush build`, then per-package `npm publish --provenance --access public` — `--access public` is required for a scoped package's first public publish, and one OIDC exchange per `npm publish` is correct.

**③ `id-token: write` scope is minimal, not enlarged.** Job-level `contents: read` + `id-token: write` only, scoped to the `publish` job; `version.yml` carries no `id-token` (correct — it doesn't publish). Good separation.

**④ Workflow rename is consistent at HEAD.** At `bd42fde` the files are `release.yml` / `version.yml` with **zero** residual `feishu-transport-*.yml` references anywhere in the tree, and every cross-reference (decision 0005, both workflow headers, the trusted-publisher instruction, the `gh pr create` body) agrees.

**⑤ No internal content.** `.npmrc` pins the public registry (which actively *helps* OIDC); no `ou_`/`oc_`/`cli_` ids, tokens, or internal hostnames in the diff. "feishu" appears only as the public package name `@excitedjs/feishu-transport`.

**Bonus — the cold-runner stdout question.** CI runners are ephemeral, so the `plan` step cold-bootstraps rush via `npx -y @microsoft/rush@X` on *every* run, and `2>/dev/null` only silences stderr. I cold-installed (empty npm cache) to check where the bootstrap chatter lands: **all** of it (npm notices, install summary, rush banner) goes to **stderr**; stdout carries **only** the JSON. So `JSON.parse` gets clean input — the "fails red on every run / breaks the always-green claim" failure mode does **not** happen.

Also nice: `count=$((count+1))` (not `((count++))`, which would trip `set -e` at 0), `state="$(classify)" || exit 1`, the `git status --porcelain -- packages common/changes` scoping that ignores lockfile byproducts, `concurrency.cancel-in-progress: false`, and the self-breaking version loop.

### Non-blocking notes (future / optional, none gate this merge)

1. **Cold rush bootstrap on every push to main.** The custom `install-run-rush.js` shells to `npx -y @microsoft/rush@5.140.0`, so even a no-op `plan` pays a full cold rush install each run (the wrapper's own docstring owns this tradeoff). Not a correctness bug — stdout stays clean and the 20-min timeout covers it — but consider caching `~/.npm` (or the `_npx` dir) once this fires frequently.
2. **Publish ordering once a 2nd package lands.** The publish loop iterates `rush list` order. If two co-bumped packages have an intra-repo dependency on each other's *new* version, publishing the dependent first fails (loudly — which is fine, but avoidable). Moot today (one package); worth a topological-order pass when `feishu-transport` re-homes.
3. **Empty `_authToken=` from `setup-node`.** `registry-url` writes `//registry.npmjs.org/:_authToken=${NODE_AUTH_TOKEN}` to `~/.npmrc`; with no `NODE_AUTH_TOKEN` it expands empty. This is the canonical npm trusted-publishing GHA recipe (OIDC overrides it) and is consistent with the public-registry `.npmrc` — but it's the single most fragile real-world point and can only be confirmed at the first actual OIDC publish (package not landed yet). Worth an explicit check at bootstrap time.

PR #5, opened 2026-05-30
Settle the install model (monorepo-only) — finish the #4 channel refactor

leo2128:
## Why

The WIP that merged the channel refactor (#4) left `main` half-finished. `@excitedjs/dreamux` now depends on the newly-extracted `@excitedjs/feishu-transport` via the pnpm `workspace:*` protocol, the per-package `packages/dreamux/package-lock.json` was deleted (it can't be regenerated while a `workspace:*` dep is present), and the contradiction was left unresolved:

- the CI `package` job (`npm ci`) is **broken** — `npm` can't resolve `workspace:*`;
- `CLAUDE.md`'s "two install paths, both must keep working" rule is **violated**.

Both core packages are now published to npm (`@excitedjs/feishu-transport@0.0.1`, `@excitedjs/dreamux@0.1.1`), so the publish blocker #4 was waiting on is gone. The only open question was the install model.

## Decision — keep one install path (the monorepo/rush path)

Recorded in [`.agents/decisions/0006-install-model.md`](../blob/feat/complete-channel-install-model/.agents/decisions/0006-install-model.md). Decision 0001 already made the monorepo path "required once a second package lands" and explicitly left the per-package path to be retired "by a future decision"; three packages now exist, so this is the natural close-out.

`workspace:*` stays — it's the correct rush `useWorkspaces: true` idiom, and pnpm rewrites it to a real version in the **published** manifest, so external `npm install @excitedjs/dreamux` is unaffected (verified against `@excitedjs/dreamux@0.1.1`, which pins `@excitedjs/feishu-transport: 0.0.1`).

## What changed

- **CI** (`.github/workflows/ci.yml`): drop the broken `package` (`npm ci`) job; the `rush` job now runs `update → build → test` (`DREAMUX_SKIP_LIVE_CODEX=1`), so typecheck/build/test coverage is preserved on the supported path.
- **Docs**: `CLAUDE.md`, `README.md`, `packages/dreamux/README.md`, `.agents/root.md`, `.agents/components/repo-structure.md` now describe one install path and the three packages (dreamux host + `feishu-transport` core + `feishu-channel` placeholder). The stale `src/feishu/` "content/render copied verbatim" note is corrected to the thin transport adapter.
- **KB**: new `decisions/0006-install-model.md`; decision 0001's "two paths" consequence marked superseded. `.gitignore` lockfile comment updated.

No source code changed — this is the config/docs decision that makes the merged #4 coherent and CI-green.

## Verification

[...]

PR #7, opened 2026-05-30
fix(release): extract JSON from rush list stdout (banner crashed the Plan step)

leo2128:
## Symptom

Since #5 merged, the **`release` workflow fails on every push to main** at the *Plan — classify each publishable package against npm* step:

```
SyntaxError: Unexpected token 'T', "The rush.j"... is not valid JSON
    at JSON.parse (<anonymous>)
```

(`ci` and `version packages` are green; only `release` is red.)

## Root cause

The Plan step pipes the project list into a JSON parser:

```sh
node common/scripts/install-run-rush.js list --json 2>/dev/null \
  | node -e '... JSON.parse(stdin) ...'
```

Its comment assumes the rush bootstrap banner goes to **stderr**. It doesn't — `install-run-rush.js` logs `"The rush.json configuration requests Rush version 5.140.0"` (and rush's own `Invoking "rush list --json"` preamble) to **stdout**, ahead of the JSON. So `2>/dev/null` doesn't strip it, and `JSON.parse` dies on the first word.

This is **latent breakage from #4**, not a regression from #5: #4 swapped the minimal `install-run-rush.js` for the full rushstack bootstrap (which prints that banner). It never fired before because #4's merge tip commit carried `[skip ci]`, so the release workflow's first real run was the #5 merge.

## Fix

Extract the JSON object from the stream (`indexOf('{')` … `lastIndexOf('}')`) instead of parsing the whole thing, and fail loudly with a clear message if no object is found. Comment corrected to describe the actual stdout shape. One-step change, no other logic touched.

## Verification

Replayed the exact pipeline locally against the real polluted stdout:

```
@excitedjs/dreamux        0.1.1  packages/dreamux
@excitedjs/feishu-transport 0.0.1 packages/channel/feishu-transport
```

- `@excitedjs/feishu-channel` (`shouldPublish: false`) correctly excluded; exit 0.
- Both emitted packages are already at their published versions, so classify → `published → skip` → `count=0` → "nothing to publish" → the job goes green.
- `release.yml` still parses as valid YAML; diff clean of internal markers.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -84,12 +84,16 @@ jobs:
       - name: Plan — classify each publishable package against npm
         id: plan
         run: |
-          # Enumerate every Rush project with shouldPublish=true (rush list emits
-          # clean JSON on stdout; its banner goes to stderr). The plan is written
-          # to a TSV file so the loop below runs in this shell (a pipe would put
-          # the counter in a subshell and lose it).
+          # Enumerate every Rush project with shouldPublish=true. The
+          # install-run-rush.js bootstrap prints a banner ("The rush.json
+          # configuration requests Rush version ...") plus rush's own
+          # 'Invoking "rush list --json"' preamble to STDOUT before the JSON, so
+          # we extract the JSON object (first '{' .. last '}') instead of parsing
+          # the whole stream. The plan is written to a TSV file so the loop below
+          # runs in this shell (a pipe would put the counter in a subshell and
+          # lose it).
           node common/scripts/install-run-rush.js list --json 2>/dev/null \
-            | node -e 'let s="";process.stdin.on("data",d=>s+=d).on("end",()=>{for(const p of JSON.parse(s).projects){if(p.shouldPublish)process.stdout.write([p.name,p.version,p.path].join("\t")+"\n")}})' \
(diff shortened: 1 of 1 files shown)


LanTingShuXu (approved):
Approved.

I reviewed this against the failing release symptom from #5's merge and the original #4 release/install context. The change is scoped to the Plan step, and the local replay confirms the polluted `install-run-rush.js list --json` stdout is now reduced to the two publishable packages:

- `@excitedjs/dreamux`
- `@excitedjs/feishu-transport`

`@excitedjs/feishu-channel` remains excluded via `shouldPublish: false`, as expected. `git diff --check` passes, and a red-line grep over the changed workflow found no internal markers. Local PyYAML is unavailable here, but the YAML diff is structurally limited to the shell block/comment, and GitHub's workflow parser will be the authoritative check once the maintainer allows it.

Non-blocking follow-up to track separately: the release workflow header still says package publish requirements include no `workspace:` protocol deps, while the current source packages intentionally use `workspace:*` after the install-model decision. This PR fixes the immediate Plan crash/count=0 path, but before the next real package version publish, the release path should be validated end-to-end for manifest rewriting vs `npm publish` behavior.

Issue #6, opened 2026-05-30
Wire the Feishu host onto the shared @excitedjs/feishu-transport core package


YourWildDad:
## Context

The Feishu channel connection is being split into **3 layers**: an engine-agnostic core package `@excitedjs/feishu-transport` (`transport` / `parse` / `policy` / `render` / `contract`), a standing daemon that owns the single Feishu connection, and a thin stdio proxy per session. Full architecture rationale, the daemon-vs-direct-HTTP reasoning, and the experimental backing live in the claudemux tracking issue: https://github.com/excitedjs/claudemux/issues/10

dreamux should **consume the same core package** instead of carrying its own copy of the engine-agnostic logic.

## Scope for dreamux

- Replace dreamux's own `content.ts` / `render.ts` / `types.ts` with the core package.
- Adapt `bot.ts` / `feishu.ts` to call into core (`transport` / `parse` / `policy` / `render` / `contract`).
- Update `package.json` + lockfile to depend on `@excitedjs/feishu-transport`.

## Blocked on

- `@excitedjs/feishu-transport` is **not yet published to public npm** (currently E404). This is the gating prerequisite — the host port cannot merge until core is published. Any port PR stays **WIP / Draft** until then.

## Notes

- This is **not** a mechanical cherry-pick from earlier work — reconcile against dreamux's current `main`; current repo state wins.
- Regenerate the lockfile via a **fresh install against the public npm registry** (`registry.npmjs.org`); do not carry over any lockfile from older history.
- Keep the architecture/experiment details in the claudemux issue and cross-link rather than duplicate here.

## Remaining steps

- [ ] (prereq) `@excitedjs/feishu-transport` published to public npm.
- [ ] Port dreamux host to consume core (`bot.ts` / `feishu.ts` / deps).
- [ ] Fresh lockfile from the public registry.
- [ ] Open the host port as WIP / Draft until core is published.

PR #8, opened 2026-05-30
feat: thread Feishu replies through shared outbound targets

LanTingShuXu:
## Summary

- upgrade the shared Feishu transport `send(OutboundTarget, text)` path to honor `replyToMessageId` via `im.message.reply`
- render `mentionUserIds` as leading card mentions, so dreamux can @-back the original sender
- add a dreamux-local channel outbound contract so dispatcher/turn-manager code does not depend directly on Feishu types
- map durable inbound metadata (`source_chat_id`, `source_message_id`, `sender_id`) into outbound targets for normal replies, retry, and crash-recovery notifications
- add a compatibility-default Feishu inbound gate: normal user messages still deliver, while missing sender id, self-sent bot-loop messages, Feishu `bot`/`app` sender types, and ambiguous group messages before `botOpenId` resolution are dropped before durable enqueue

## Scope

This is the first dreamux implementation slice for issue #6 / the Feishu-channel isomorphism work. It keeps the access model compatible with today's behavior: no pairing or allowlist requirement is enabled. The gate is intentionally thin and defensive; the stricter `gate()` / AccessStore policy remains a pluggable follow-up.

## Verification

- `DREAMUX_SKIP_LIVE_CODEX=1 node common/scripts/install-run-rush.js test`
  - local Codex is `0.135.0`; the live `0.134.x` compatibility test is intentionally skipped by that env var
- `node common/scripts/install-run-rush.js change --verify --no-fetch`
- red-line scan over changed files for internal domains/package registries/long Feishu ids/secrets: no new hit; only existing source comments mention generic token handling

### DIFF
--- a/.agents/CONTRIBUTING.md
+++ b/.agents/CONTRIBUTING.md
@@ -22,7 +22,7 @@ being terse.
 
 - new package added / removed under `packages/`
 - a CLI subcommand added / removed / renamed (issue #4 made `dreamux` the
-  unified surface — see decision 0002)
+  unified surface — see [the CLI naming decision](decisions/cli-and-package-naming.md))
 - a settled design decision after debate (write a decision record)
 - a new external dependency that materially shapes the runtime (a
   database engine swap, a new IPC mechanism)
@@ -52,7 +52,7 @@ being terse.
 | Kind | When to use | Naming |
 |---|---|---|
 | `components/<thing>.md` | A piece big enough to have its own contract / mental model (server, CLI, codex-client, feishu-bot, repo-structure) | kebab-case, no number |
-| `decisions/NNNN-<slug>.md` | A choice that was debated and settled; future maintainers need the *why* | 4-digit zero-padded number, kebab-case slug |
+| `decisions/<slug>.md` | A choice that was debated and settled; future maintainers need the *why* | topic slug, kebab-case, no sequence number |
 | `domains/<area>.md` | A cross-cutting contract that spans multiple components | kebab-case |
 | `proposals/<slug>.md` | An active design under discussion | kebab-case |
 | `research/<slug>.md` | A frozen investigation snapshot; must end with an explicit "disposition" section (Promoted / Deferred / Out of scope) | kebab-case |
@@ -61,7 +61,7 @@ being terse.
 ## Decision record template
 
 ```markdown
-# NNNN — <decision title>
+# <decision title>
 
 - **Status:** Accepted | In progress | Superseded by [link]
 - **Date:** YYYY-MM-DD
--- a/.agents/components/repo-structure.md
+++ b/.agents/components/repo-structure.md
@@ -2,7 +2,7 @@
 
 Rush + pnpm monorepo since issue #4. Three packages today, all wired
 through pnpm `workspace:*` and installed via the rush path only (see
-[decision 0006](../decisions/0006-install-model.md)):
+[the install-model decision](../decisions/install-model.md)):
 
 | Package | Folder | Role |
 |---|---|---|
@@ -20,7 +20,7 @@ sibling claudemux repo import one implementation instead of drifting copies.
 |---|---|
 | `/rush.json` | Rush project list + pnpm/Node version pins |
 | `/common/config/rush/` | Rush command definitions (`command-line.json`), pnpm `.npmrc`, version policies, generated `pnpm-lock.yaml` |
-| `/common/scripts/install-run-rush.js` | Bootstrap that shells out to `npx @microsoft/rush@<version>` (see [decision 0001](../decisions/0001-rush-pnpm-monorepo.md)) |
+| `/common/scripts/install-run-rush.js` | Bootstrap that shells out to `npx @microsoft/rush@<version>` (see [the Rush + pnpm decision](../decisions/rush-pnpm-monorepo.md)) |
 | `/common/temp/` | Rush working dir (gitignored) |
 | `/packages/dreamux/` | The `@excitedjs/dreamux` package |
 | `/bin/` | Thin shims that forward to `/packages/dreamux/bin/` so pre-monorepo PATH entries keep working |
@@ -61,8 +61,8 @@ The per-package `cd packages/dreamux && npm install` path is **retired**:
 pnpm `workspace:*` protocol, which `npm` cannot resolve. There is no
 committed per-package `package-lock.json`. External consumers are
 unaffected — pnpm rewrites `workspace:*` to a real version at publish time.
-See [decision 0006](../decisions/0006-install-model.md) (which retires the
-two-paths consequence of [decision 0001](../decisions/0001-rush-pnpm-monorepo.md)).
+See [the install-model decision](../decisions/install-model.md) (which retires
+the two-paths consequence of [the Rush + pnpm decision](../decisions/rush-pnpm-monorepo.md)).
 
 ## Rush change files
 
@@ -114,7 +114,7 @@ multi-package release notes precise while still using Rush as the validator.
 
 - npm package: `@excitedjs/dreamux`
 - CLI binaries installed by the package:
-  - `dreamux` (preferred, see [decision 0002](../decisions/0002-cli-and-package-naming.md))
+  - `dreamux` (preferred, see [the CLI naming decision](../decisions/cli-and-package-naming.md))
   - `dreamux-server` (legacy alias)
   - `server-ctl` (legacy alias)
 
@@ -126,4 +126,4 @@ multi-package release notes precise while still using Rush as the validator.
 | `~/.codex-host/` | Server-owned runtime state: SQLite (`state.db`), admin socket, per-dispatcher codex sockets and logs. | The server |
 
 The split is load-bearing: a `rm -rf ~/.codex-host` recovery never loses
-user-edited settings. See [decision 0003](../decisions/0003-global-config-dir.md).
+user-edited settings. See [the global-config decision](../decisions/global-config-dir.md).
--- a/.agents/components/repo-structure.md
+++ b/.agents/components/repo-structure.md
@@ -64,6 +64,52 @@ unaffected — pnpm rewrites `workspace:*` to a real version at publish time.
 See [decision 0006](../decisions/0006-install-model.md) (which retires the
 two-paths consequence of [decision 0001](../decisions/0001-rush-pnpm-monorepo.md)).
 
+## Rush change files
+
+Release-facing changes need Rush change files under
+`/common/changes/<package>/`. Always validate them before pushing:
+
+```bash
+node common/scripts/install-run-rush.js change --verify --no-fetch
+```
+
+Rush can generate change files non-interactively when every changed package
+uses the same release type and message:
+
+```bash
+node common/scripts/install-run-rush.js change \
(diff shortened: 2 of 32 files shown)


leo2128 (approved):
Cross-review (claudemux side). **Approve** — both slices are correct and well-scoped; I independently checked out the branch and ran `rush build` (exit 0) and `rush test` (exit 0, `DREAMUX_SKIP_LIVE_CODEX=1`). A few non-blocking comments + one small test-coverage request.

### Outbound contract

1. **Reply-routing invariant changed — please re-document (core, affects claudemux too).** The old `send` doc explicitly said it routes by `chat_id` *"never by a message_id, so a forged reply target cannot redirect the message into an unrelated conversation."* With `replyToMessageId` now routed through `im.message.reply`, the message goes to *that message's* chat and `chatId` is effectively ignored. It's safe **here** because `outboundTargetForInbound` derives both fields from the same trusted durable row — but the core contract dropped the guarantee and its comment. Since claudemux will consume the same `send`, please document that `replyToMessageId` must belong to `chatId`'s conversation (caller's responsibility), or assert it.

2. **Channel abstraction shape (isomorphism).** `ChannelOutboundTarget` / `OutboundSink` living dreamux-local is the right call for this slice — agree with your two-layer plan (abstract interface per-host now, extract `@excitedjs/channel-contract` later). The one ask: let's freeze the *shape* (`conversationId` / `replyToMessageId` / `mentionUserIds` / `conversationKey`) as the agreed cross-repo contract now, so claudemux's daemon adapter mirrors it exactly and we don't drift before the package split.

3. **Nits.** (a) Multi-card + reply: every card replies to the same source `message_id` (not chained) — fine, just confirming it's intended; the @-back prefix correctly lands on the first card only. (b) `client.im.message as unknown as MessageApiWithReply` — worth a one-line comment that the lark SDK doesn't surface `reply`'s type. (c) `@`-back is applied to every reply including p2p/DM, where mentioning the sole counterpart is redundant; consider gating @-back to group chats (low priority).

### Compatible inbound gate

4. **Test the loop-prevention branches (small, recommend before merge).** `feishu-gate.test.ts` covers deliver / missing-sender / group-unknown-botid, but **not** the two most security-relevant drops: `senderId === botOpenId` (self/bot-loop) and `isBotSenderType(senderType)`. Those are the branches that stop a channel self-amplifying — cheap to lock with two more cases.

5. **`mentions` is dead input.** `CompatibleFeishuGateInput.mentions` is accepted (and passed from `server.ts`) but never read by the gate. Either drop it or comment that it's reserved for the future allowlist/mention-gating extension.

6. **Local gate vs core `gate()` (alignment, not a blocker).** Writing a dreamux-local `compatibleFeishuGate` instead of calling core `gate()` with a permissive policy is reasonable for the compat default, but it means core's hazard logic isn't the single source of truth — a drift risk as the pluggable extension lands. Worth making the extension able to swap in core `gate()`. I'll hit the same fork on claudemux, so let's keep these aligned.

Nice that the unresolved-`bot open_id` group case is now an **explicit, logged** drop rather than the silent drop from the #4 notes. 👍

LanTingShuXu:
Addressed the cheap follow-ups from review:

- documented the reply routing precondition in the shared Feishu transport: when `replyToMessageId` is used, callers must only pass a message id observed in the same `target.chatId` conversation;
- added direct unit coverage for `compatibleFeishuGate` self-loop (`senderId === botOpenId`) and Feishu `bot`/`app` sender-type drops.

Re-ran:
- `DREAMUX_SKIP_LIVE_CODEX=1 node common/scripts/install-run-rush.js test`
- `node common/scripts/install-run-rush.js change --verify --no-fetch`

LanTingShuXu:
Added the requested core increment for the claudemux daemon/proxy work:

- exported `toChannelInbound(parsed: ParsedInbound)` from `@excitedjs/feishu-transport`;
- extended `ParsedInbound` with optional `meta?: Record<string, unknown>` so hosts can attach raw-event narrow metadata before conversion;
- `toChannelInbound` returns `{ text, meta: Record<string,string> }`, keeps text explicit (`(empty message)` when empty), drops protocol-unsafe keys such as hyphenated names, and drops nested / non-string meta instead of blind stringification;
- added tests for string-only underscore metadata, media placeholder preservation, and empty-text degradation.

Important I/O clarification for claudemux: the current `ParsedInbound` only contains text, so `toChannelInbound(parsed)` cannot derive `message_id/chat_id/sender_id` by itself. The daemon should construct `parsed.meta` from the Feishu raw event normalizer, then call `toChannelInbound(parsed)`. That keeps core engine-agnostic and avoids passing raw Feishu event structure into this pure helper.

Re-ran:
- `DREAMUX_SKIP_LIVE_CODEX=1 node common/scripts/install-run-rush.js test`
- `node common/scripts/install-run-rush.js change --verify --no-fetch`

LanTingShuXu:
Follow-up to the narrow inbound helper: I chose option A from the claudemux coordination thread to avoid host-level drift.

Added:
- `narrowMetaFromEvent(rawEvent) -> Record<string, unknown>` in core `parse/`;
- export from `@excitedjs/feishu-transport`;
- tests for wrapped Feishu event payloads, already-unwrapped payloads, and dropping missing / empty / nested / non-string envelope values.

Canonical keys extracted today: `message_id`, `chat_id`, `chat_type`, `sender_id`, `sender_type`, `root_id`, `parent_id`, `create_time`.

Intended host call shape:

```ts
const parsed = parseInbound({ message_type, content, mentions })
const payload = toChannelInbound({
  ...parsed,
  meta: narrowMetaFromEvent(rawEvent),
})
```

Re-ran:
- `DREAMUX_SKIP_LIVE_CODEX=1 node common/scripts/install-run-rush.js test`
- `node common/scripts/install-run-rush.js change --verify --no-fetch`

YourWildDad (commented):
## Independent design-conformance review — verdict: **CONFORMS-WITH-NITS**

Third-party gate against issue #6 (dreamux) + issue #10 (claudemux). Not a general code review — only checking the implementation conforms to the agreed channel-agnostic design. Nothing here blocks merge.

### Pass — the core decoupling point holds (items 1 & 2)
- **No Feishu field leakage into the dispatcher-facing contract.** `ChannelOutboundTarget` (`packages/dreamux/src/channel/outbound.ts`) exposes only abstract concepts — `conversationId` / `replyTo` / `mentionUsers` / `conversationKey`. The Feishu names (`chatId` / `replyToMessageId` / `mentionUserIds`) and the durable-row names (`source_message_id` / `sender_id`) never surface on the neutral contract. This is a real decouple, not a rename.
- **Dependency direction is correct.** `TurnManager` and `DispatcherRuntime` import only `OutboundSink` / `outboundTargetForInbound` from `../channel/outbound.js`; neither pulls a Feishu type. The neutral→Feishu mapping is isolated at the composition root (`server.ts`) and core transport `send()` (`replyToMessageId` → `im.message.reply`, `mentionUserIds` → leading `<at>` mentions). Chain `dispatcher → neutral contract → Feishu adapter → core` is intact.

### Pass — gate hookup (item 3)
- `compatibleFeishuGate(...)` runs inside `bot.start(...)` and `return`s before `runtime.enqueueInbound(...)`, i.e. before durable enqueue. Drop reasons (missing sender, self-loop, bot/app sender type, ambiguous group pre-`botOpenId`) match the described compatibility-first gate; smoke tests assert no enqueue / no turn / no outbound on drop.

### Pass — red-line / supply-chain (item 5)
- Diff grep for internal mirrors / domains / system names / long Feishu ids / corporate emails / absolute paths: **no hits**. No `pnpm-lock.yaml` change — correct, since this PR adds no new external dependency (`@excitedjs/feishu-transport` is an in-repo `workspace:*`), so there is no lockfile-residue surface to regenerate.

### Item 4 — N/A
Dedup / idempotency key is not touched by this PR (the gate is loop-prevention, not dedup). Skipped per scope.

---

### Nits (non-blocking)

1. **(nit, item 3) Core increment is added + well-tested but not yet consumed by the host.** `narrowMetaFromEvent` / `toChannelInbound` land in `packages/channel/feishu-transport/src/parse/content.ts`, yet `bot.ts` `normalizeInboundEvent` still hand-extracts the same envelope fields (`senderId`, `chatId`, `chatType`, …) — the exact copy-drift the function's own docstring says it exists to prevent. Acceptable for an explicit "first slice"; the natural follow-up is wiring the host onto `narrowMetaFromEvent` so the field mapping has one owner. **Positive note:** the increment itself conforms to #10's spec *exactly* — meta keys (`message_id`/`chat_id`/`chat_type`/`sender_id`/`sender_type`/`root_id`/`parent_id`/`create_time`), the `/^[A-Za-z0-9_]+$/` key guard, drop-don't-stringify for nested/non-string values, and test-locked media/empty → explicit placeholder degradation.

2. **(nit) Neutral→Feishu mapping lives inline in `server.ts`.** The `ChannelOutboundTarget → OutboundTarget` field translation is an object literal in the `outbound: { send: ... }` wiring. Defensible (server.ts is the composition root), but moving it into the `feishu/` adapter would keep the Feishu field vocabulary fully contained in one module.

3. **(nit) Dead gate input field.** `CompatibleFeishuGateInput.mentions` is passed from `server.ts` (`mentions: event.mentions`) but never read in `compatibleFeishuGate`. Drop it or use it.

4. **(FYI, cross-repo) Field-name reconcile, later.** claudemux #10 anticipated mirroring `replyToMessageId` / `mentionUserIds`; this PR (correctly, for decoupling — those are core's Feishu names) uses neutral `replyTo` / `mentionUsers`. claudemux has not shipped `ChannelOutboundTarget` yet, so nothing is locked into divergence; just reconcile names when `@excitedjs/channel-contract` is extracted. dreamux is the source of truth here, and its choice is the more-neutral one — no change requested in #8.

LanTingShuXu:
Addressed the three design-conformance nits in `606f58a`:

- `normalizeInboundEvent` now consumes the core helpers directly: `parseInbound(...)` + `narrowMetaFromEvent(raw)` + `toChannelInbound(...)`, so envelope meta extraction and sanitization no longer drift in the host.
- Removed the dead `mentions` field from `CompatibleFeishuGateInput` and its call site.
- Moved the neutral `ChannelOutboundTarget` -> Feishu `OutboundTarget` mapping into the Feishu adapter via `channelOutboundToFeishuTarget(...)`; `server.ts` now only composes the adapter.

Verification run locally:

```text
node common/scripts/install-run-rush.js build
node common/scripts/install-run-rush.js typecheck
DREAMUX_SKIP_LIVE_CODEX=1 node common/scripts/install-run-rush.js test
node common/scripts/install-run-rush.js change --verify --no-fetch
```

All passed. The test run skipped only the live Codex smoke path because the local Codex version is intentionally not the pinned live-test version.

YourWildDad:
Superseded by #10 (merged as 88e8b1b). Rebasing this branch onto current `main` pulled in `.github/workflows/ci.yml` (merged via #9), and the fork's GitHub token lacks the `workflow` scope, so the fork branch could not be force-pushed. The conflict-free rebased equivalent was pushed straight to `excitedjs/dreamux` and merged as #10. Work and review (one APPROVE + an independent conformance review CONFORMS-WITH-NITS, all 3 nits fixed) carry over unchanged.

PR #9, opened 2026-05-31
ci: strengthen dreamux pull request gates

leo2128:
Authored by **@LanTingShuXu (x教授)**; applied + pushed by @leo2128 because modifying `.github/workflows/ci.yml` requires a GitHub token with `workflow` scope, which the author's token lacks. The original local commit's author identity used a non-public email and was **intentionally dropped on re-apply** so nothing internal lands in this public repo — the exact thing the new author-email gate guards against.

## What

- **`ci.yml`**
  - **`rush-change-status`** — PR gate: release-surface changes must declare Rush change files (`rush change --verify` against the base branch).
  - **`commit-metadata`** — reject commit *author* emails that would leak an internal / machine identity into the public repo.
  - **`shellcheck`** lane (ubuntu + macOS) over `.agents/scripts/check.sh`, `common/git-hooks/pre-commit`, and `bin/{dreamux,server,server-ctl}`.
  - **`rush`** lane → ubuntu + macOS matrix, and runs `rush typecheck` explicitly before build/test.
- **README** + **decision 0006 (install-model)** document the expanded CI surface.

## Verification (local, on this applied state)

`.agents/scripts/check.sh`, `rush change --verify`, `rush typecheck`, `rush build`, `rush test` — all pass. `shellcheck` isn't installed locally; the new CI lane exercises it.

## Note for review

The author-email gate **names the internal domains it blocks**. This repo accepts public contributor emails (gmail / 163 / github-noreply are already in history), so an allowlist would reject legitimate contributors — a denylist must name what it blocks. These are the only such strings introduced, and they live inside a security control rather than as leaked content. Flagging it explicitly since the repo had none before; happy to adjust the approach if the owner prefers.

> Heads-up: because this PR touches `.github/workflows/`, GitHub gates fork-PR CI behind a maintainer "approve and run workflows" click before the checks execute.

### DIFF
--- a/.agents/decisions/0006-install-model.md
+++ b/.agents/decisions/0006-install-model.md
@@ -38,8 +38,9 @@ which install model to keep.
   version (`@excitedjs/feishu-transport: 0.0.1`) in the **published** manifest.
 - No per-package `package-lock.json` is committed.
 - CI's `package` job (`npm ci`) is removed; its typecheck/build/test coverage
-  moves into the `rush` job, which now runs `rush update` → `rush build` →
-  `rush test` (`DREAMUX_SKIP_LIVE_CODEX=1`, no codex binary on the runner).
+  moves into the `rush` job, which now runs `rush update` → `rush typecheck` →
+  `rush build` → `rush test` (`DREAMUX_SKIP_LIVE_CODEX=1`, no codex binary on
+  the runner).
 
 Path 1 was always framed as "pre-monorepo muscle memory." Decision 0001 itself
 made the monorepo path "required once a second package lands"; three packages
@@ -56,7 +57,9 @@ now exist, so retiring path 1 is the natural close-out, not a new constraint.
 - **Foot-gun:** don't re-add a per-package `package-lock.json` or a `package`
   CI job — `npm` cannot lock a `workspace:*` graph, so either would reintroduce
   the exact breakage this record removes.
-- **Guards:** the `rush` CI job is the authoritative install/build/test gate;
+- **Guards:** the `rush` CI job is the authoritative
+  install/typecheck/build/test gate; PRs also run `rush change --verify` against
+  the base branch so release-surface changes declare Rush change files.
   `CLAUDE.md`, `README.md`, and `components/repo-structure.md` all describe the
   single path; this record supersedes the "two paths" consequence of 0001.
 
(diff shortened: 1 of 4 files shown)


LanTingShuXu (changes requested):
Request changes.

The direction is right, but this PR cannot merge in its current state because the newly added gates are already red on the PR run.

Blocking findings:

1. The new `shellcheck` job fails on `.agents/scripts/check.sh` with SC2295 at line 43. Once a lint lane is introduced, the repository has to pass it on the introducing PR. Fix the script (`rel="${f#"$KB_ROOT"/}"`) or keep the lane out until the checked files are clean.

2. The expanded `rush` lane fails on both ubuntu and macOS during `rush typecheck`. The current log only reports that `@excitedjs/dreamux` and `@excitedjs/feishu-channel` returned exit code 2, without enough diagnostic output to review the cause. Either fix the underlying typecheck failure on GitHub-hosted Node 22, or adjust the workflow to surface package-level diagnostics (`--verbose` or equivalent) and then fix what it reports. A CI-strengthening PR has to be green under the CI environment it defines.

3. The author-email gate hardcodes company-specific domains directly in the public workflow. Given the explicit open-source red line, I would not commit those strings into the repository as the default mechanism. Move the private denylist into repository configuration instead, e.g. a GitHub Actions variable such as `AUTHOR_EMAIL_DENYLIST_REGEX`, and have the workflow enforce it when present while keeping only generic public checks in the committed file (`localhost`, `.local`, invalid email shape, etc.). That preserves the guardrail without making the public repo carry company-domain policy data.

The `rush change` gate, KB check, gitleaks, and commit metadata mechanics otherwise look structurally reasonable. Re-run CI after the above changes; I’ll re-review once the new gates are green.

LanTingShuXu (approved):
APPROVE.

Re-reviewed the latest head after the three blocking fixes:

- `shellcheck` is now green on both ubuntu and macOS; the SC2295 issue in `.agents/scripts/check.sh` is fixed.
- The Rush lane now runs `update -> build -> typecheck -> test`, which matches the fresh-runner dependency on generated `dist/*.d.ts`; both ubuntu and macOS lanes are green.
- The author-email gate no longer commits company-domain strings. The committed workflow only contains generic checks plus the optional `AUTHOR_EMAIL_DENYLIST_REGEX` repo variable hook, which is the right boundary for private policy.

I also rechecked the current PR checks: all 8 CI jobs are green, including gitleaks, rush change, shellcheck, KB, and the OS matrix rush lane.

No remaining blockers from me.

PR #15, opened 2026-06-02
docs(agents): clarify dispatcher tm boundary

YourWildDad:
## Summary
- record that dreamux server hosts dispatcher lifecycles only
- remove the server-owned Codex teammate runtime slice from this PR
- keep tm-launched work behind the dispatcher command boundary

## Validation
- Not run (docs-only scope after architecture correction).


YourWildDad:
## Heterogeneous review — server-owned Codex teammates (Claude reviewing the Codex-authored backend)

Scope: the **dreamux server backend only**. The `@excitedjs/tm` adapter half is explicitly a follow-up in the sibling repo and out of scope here. I did not flag the one unskipped `codex-0134-live` failure — that's the Codex 0.134.x pin vs the locally installed 0.135.0 (environment, not code).

### Verified premises — the design works as intended

- **The core "survive across turns" claim holds.** The daemon is spawned by the long-running server via `CodexProcess` (`detached: true`, its own process group, stdio redirected to log files), so it lives in the *server's* process tree, not in the short-lived `dreamux teammate send` CLI's tree. A Codex shell sandbox tearing down its children (die-with-parent) cannot reach it. The reuse test (`counter.count === 1`, a single `initialize`, two `turn/start`) confirms one daemon is reused across sends. ✅
- **The migration is correct.** `CURRENT_USER_VERSION` 1→2, the `row < 2` guard is idempotent and transactional, and the `0002` CHECK set (`declared|starting|ready|degraded|stopping|stopped`) matches `CodexTeammateStatus = DispatcherStatus` exactly, so no row can violate it. ✅
- **Path safety.** `TEAMMATE_NAME_RE` (`^[A-Za-z0-9][A-Za-z0-9_-]{0,63}$`) forbids `.` and `/`, so `name` is safe to interpolate into `runtime_dir/teammates/<name>/…`. `mustDirectory` validates the cwd exists, and `createCwd: false` correctly refuses to invent a user worktree. ✅
- **Turn serialization** via `turnQueue` (chained through both fulfil and reject) is correct — Codex threads are single-turn-at-a-time, and a failed turn doesn't break the chain. Start coalescing via `startingTeammates` mirrors the proven dispatcher pattern. ✅

### Findings

**[High] A teammate has no recovery path out of `degraded` — once the daemon connection drops, every later `send` fails permanently.**

`teammate-runtime.ts` installs an `onClose` handler that sets `degraded` but leaves the runtime in `Server.teammateSlots`. After that:
- `Server.sendTeammate` → `startTeammate` short-circuits on `this.teammateSlots.has(name)` and returns the *dead* runtime;
- `runOneTurn` throws `codex teammate '…' is not ready` because `status !== 'ready'`.

So a `degraded` teammate is bricked until `teammate.kill` + `teammate.spawn`. Nothing evicts or rebuilds the slot. Note the divergence from `DispatcherRuntime`, which installs **no** `onClose` at all (it never auto-degrades on a mid-life disconnect) — the two runtimes now handle connection loss inconsistently. Minimal fix: have `onClose` evict the slot (`teammateSlots.delete(name)`) so the next `send` rebuilds, and/or have `startTeammate` tear down a non-`ready` slot before re-creating.

**[Med] `thread/resume` failure is unhandled and feeds the same dead end — and this is exactly the PR's headline path.**

`CodexTeammateRuntime.resolveThread` calls `thread/resume` with no try/catch. `DispatcherRuntime.resolveThread` wraps it and falls back to a fresh thread + `recordLostThread`. For a teammate, a failed resume throws out of `start()` → `degraded` → (per the High finding) stuck. Restart is precisely when resume is most likely to fail: Codex thread store rotated, codex home moved, version drift.

I'm deliberately **not** recommending you copy the dispatcher's silent fresh-thread fallback: a teammate is driven by a programmatic adapter that passed an explicit `thread_id` and expects continuity — silently swapping to a fresh thread would make the adapter believe it resumed while quietly losing all context. Also `CodexTeammateRow` has no `last_lost_thread_id` column, so there is nowhere to surface the loss the way the dispatcher does. This is a semantics decision to make on purpose:
- **fail-loud-but-recoverable** — surface the resume error and keep the row usable for a `kill`+`resume` with a corrected thread; or
- **fresh-thread-with-an-explicit-loss-signal** — needs a new column + a status the adapter can read.

Either is fine; the current "throw and brick" is not.

> The High finding and this one share one root: there is no route out of `degraded`. Mid-life disconnect and resume-failure are just two triggers into the same dead end. Worth noting the cross-restart resume guarantee is only covered by the fake-codex smoke test — there is no live-codex test for the teammate resume path.

**[Med] Status is reported as falsely healthy after a non-graceful restart.**

`summarizeTeammates` / `teammate.status` return `runtime?.getStatus() ?? row.status`. After a crash (no graceful `shutdown()`), the DB row keeps its last value (`ready`), but no runtime exists, so `teammate.list` / `teammate.status` report `ready` while nothing is running — until the next `send` lazily restarts it. A consumer that gates on status before sending gets a false-healthy reading. (Graceful shutdown is fine — it persists `stopped`.) Dispatchers avoid this because boot eagerly restarts enabled rows (the `listEnabled()` loop) and reconciles inbound `running → unknown`; teammates have neither. Consider reconciling persisted teammate status to `stopped`/`unknown` at boot for rows with no live runtime.

**[Med/Low] `teammate.spawn` is not atomic: a start failure leaves an orphan row that then blocks re-spawn with CONFLICT.**

`spawnTeammate` does `create()` then `startTeammate()`. If start fails (invalid `codex_args_json`, missing/relocated cwd, codex bin not found), the row persists, and the next `teammate.spawn` hits `SQLITE_CONSTRAINT_PRIMARYKEY` → `CONFLICT`. The operator must `teammate.kill` first to clear it. Consider rolling the row back on start failure, or making `spawn` restart an existing non-running row instead of conflicting.

**[Low] `kill` and `send` are not mutually exclusive.** `removeTeammate` → `stopTeammate` awaits the `turnQueue` snapshot it captured, but a concurrent `sendTeammate` can append a new turn after that snapshot, or re-enter `startTeammate` before `teammateSlots.delete`. So a racing turn can run against a closing client, or briefly re-spawn a daemon for a row about to be removed. Single-user admin socket, so low severity — but the teammate lifecycle has no per-name mutex the way starts do.

**[Info — pre-existing, not introduced here] Notification handlers accumulate on the long-lived client.** Each `runTurn` registers a new `client.onNotification` handler (`rpc.ts` keeps them in an append-only array) that is never removed — only short-circuited by a `done` flag — and its closure retains that turn's `itemsByTurn`. This lives in `events.ts` / `rpc.ts` and affects `DispatcherRuntime` equally; flagging only as a follow-up to deregister the collector on turn completion.

**[Info — pre-existing, local-trust] `extraArgs` can override the sandbox allowlist.** `codexArgsToCli` appends raw `extraArgs` after `-c sandbox_mode=…`, and Codex is last-write-wins, so `extraArgs: ["-c","sandbox_mode=danger-full-access"]` bypasses `ALLOWED_SANDBOX_MODES`. Same for dispatchers, and the admin socket is same-user — noting only.

Minor nit: there is no standalone `teammate.start` / restart admin verb — `send` is the only way to (re)start a row after a server reboot, since `spawn`/`resume` both `create()` and CONFLICT on an existing row. Fine if the adapter always drives via `send`.

### Net

The backend achieves its core goal, and the migration / path-safety / serialization fundamentals are solid. The one thing I'd want resolved before this leaves draft is the **single shared `degraded` dead end (the High finding + the resume-failure Med)**: give a teammate a way back to `ready` without a manual kill, and decide the resume-failure semantics deliberately.

— Claude (heterogeneous reviewer)

YourWildDad:
## Re-review of `18f460d` (fix: recover degraded Codex teammates)

Verified both by static trace and empirically: `rush build` (incl. `tsc --noEmit`) passes, and `smoke.test.ts` is **26/26**, including the 5 new teammate cases. Verdict: **both root causes are fixed, three of the Med findings are cleared too, and no serious new holes were introduced.**

### The two roots

**[High → fixed] Degraded teammate is now rebuildable.**
`startTeammate` short-circuits only when `getStatus() === 'ready'`; `doStartTeammate` tears down a non-ready slot (`stopTeammate`) before building a fresh runtime. The `rebuilds after a degraded connection` test confirms the full path: drop the connection → `degraded` → next `send` rebuilds (two `initialize`s, spawn counter 2, back to `ready`). The discarded runtime's `suppressCloseDegrade = true` is harmless — each runtime is single-use after `stop()`, never restarted.

**[Med → fixed] `thread/resume` failure is loud and recoverable.**
`resolveThread` now wraps `thread/resume`, re-throws a wrapped error, and crucially does **not** touch `thread_id` and does **not** start a fresh thread. On failure the slot is never set, so the row stays `degraded` with `last_error` and the original `thread_id` intact, and the next `send` retries the *same* thread. The `resume failure is loud but retryable` test confirms: after failure `getTeammateRuntime` is null, the row is `degraded` with `thread_id` preserved, then a retry resumes the same thread. This is the fail-loud-but-recoverable semantics, not a silent swap — correct call. `teammate.resume` additionally moved from create-or-CONFLICT to `updateForResume`, so a saved-but-unresumable thread id can be explicitly replaced and restarted.

### Bonus: three Med findings also cleared

- **Falsely-healthy status after a crash:** `reconcileStoppedOnBoot()` resets `starting`/`ready`/`stopping` rows to `stopped` (leaving `degraded` as-is, which is correct — that state is real). `list`/`status` no longer report a phantom live daemon after a non-graceful restart.
- **Non-atomic spawn:** `spawnTeammate` now rolls the new row back (`stopTeammate` + `remove`) when startup throws, so a corrected re-spawn no longer hits a stale CONFLICT. Covered by the bogus-`approvalPolicy` test.

### New-hole sweep: nothing serious — two minor smells, non-blocking

1. **`reconcileStoppedOnBoot()` runs in the constructor, before the admin pidfile lock is acquired** (the lock lives in `start()` → `admin.start()`). If a second server is *constructed* against a DB a live server already owns, it would reset the live server's `ready` rows to `stopped` before its own `start()` is rejected by the lock. Practical impact is near-nil — it's masked by `runtime?.getStatus() ?? row.status` (the in-memory runtime status wins for the live server), it requires two servers on one DB (exactly what the lock prevents), and the admin API is the only reader. Cheap tidy-up: move `reconcileStoppedOnBoot()` into `start()` after the lock is held.
2. **`kill` racing a `send`-triggered rebuild** — the original Low (`kill`/`send` have no per-name mutex), now with a slightly wider window because rebuild adds another teardown→restart sequence. A concurrent `kill` + `send` on the same name can in principle orphan a daemon (running, in `teammateSlots`, with its DB row removed). Single-user admin socket → still Low, still non-blocking.

The two pre-existing Info items (per-turn `onNotification` handlers accumulating on the long-lived client in `events.ts`/`rpc.ts`; `extraArgs` overriding the `sandbox_mode` allowlist via last-write-wins) are untouched by this fix and correctly out of scope here.

Nice turnaround — the `degraded` dead end is gone and the resume semantics are now a deliberate, tested choice.

— Claude (heterogeneous reviewer)

Issue #36, opened 2026-06-03
dreamux MVP implementation: Feishu-channel Codex dispatcher server


YourWildDad:
## Summary

This issue tracks implementation of the converged dreamux MVP architecture.

Design source of truth:

- `.agents/decisions/top-level-design.md`
- `.agents/decisions/dispatcher-tm-packaging.md`

The MVP is a local `dreamux serve` process that hosts dispatcher instances. Each dispatcher receives Feishu long-connection events, gates and coalesces inbound messages, injects accepted work into one Codex thread, and allows Feishu outbound only through an explicit dispatcher-scoped MCP tool path. The Feishu MCP process is a stdio shim that proxies outbound requests back to the live server over the local admin socket, so Feishu credentials and channel-owned reaction state stay in `dreamux serve`.

## Key decisions

- D1: One dispatcher is one shared-context trust domain; per-chat isolation is out of scope for the MVP.
- D2: Codex uses a dispatcher-scoped Feishu MCP stdio shim; the shim proxies to `dreamux serve` over local IPC and does not hold Feishu credentials.
- D3: Consecutive inbound messages from the same chat are coalesced into one Codex turn.
- D4: Inbound messages and reaction ownership are not persisted; stale received reactions after restart are an accepted MVP limitation.
- D5: There is no turn timeout; only a real Codex app-server child or child-WebSocket failure triggers backoff restart and thread resume.
- D6: `uninstall` removes dreamux-owned config, state, logs, and service integration; workspace-local dispatcher skills are reported but not deleted by default.
- D7: Feishu inbound uses the SDK long-connection WebSocket path; webhook delivery is out of scope for the MVP.

Compatibility note: dreamux does not pin, bundle, or host Codex. Compatibility is guarded through `doctor`, live tests, and version diagnostics.

## Implementation checklist

- [x] Task 1: Runtime paths/state/log contract. Deps: none. Implement `~/.dreamux/state`, `~/.dreamux/logs`, `admin.sock`, and per-dispatcher `codex.sock` path builders in `src/runtime/paths.ts`, including Unix socket path budget tests.
- [x] Task 2: Config schema, validation, and secret hygiene. Deps: Task 1. MVP Feishu config contains only `app_id` and `app_secret`; all declared dispatcher app ids must be unique; config is owner-only; secret values are redacted from config display, status, doctor output, and logs.
[...]

PR #38, opened 2026-06-03
Task 2: validate config schema and secrets

YourWildDad:
## Summary

- Enforce the MVP Feishu config schema: dispatcher bot config accepts only `app_id` and `app_secret`, requires non-empty values, and rejects duplicate `app_id` declarations.
- Require private `0600` config files on config reads, make `dreamux serve` fail loudly when config is missing, and keep `config show` redacted with no raw mode.
- Validate dispatcher ids before they become path segments, add `serverJsonPath()`, and update docs that still described legacy runtime paths or raw config display.

## Tests

- `.agents/scripts/check.sh`
- `git diff --check`

Tracking: #36 Task 2.

### DIFF
--- a/.agents/decisions/global-config-dir.md
+++ b/.agents/decisions/global-config-dir.md
@@ -32,11 +32,12 @@ We needed a user-editable global config that is:
 
 ## Decision
 
-Create `~/.dreamux/config.json` at server startup if absent. Subsequent boots
-merge missing keys with built-in defaults in memory. `dreamux onboard` writes
-the file with mode `0600` and upserts only the current
-`feishu.bots.<dispatcher-id>` entry when the JSON file already exists; existing
-global settings and other bot secrets are preserved.
+Historically this decision created `~/.dreamux/config.json` at server startup
+if absent. The current MVP, superseded by top-level-design, fails loudly when
+the file is missing from `dreamux serve`; `dreamux onboard` creates the file
+with mode `0600` and upserts only the current `feishu.bots.<dispatcher-id>`
+entry when the JSON file already exists. Existing global settings and other bot
+secrets are preserved.
 
 Older installs that have only `~/.dreamux/config.toml` fail fast with an
 explicit migration message. dreamux does not silently create default JSON over
@@ -82,8 +83,8 @@ a same-key global default. See `src/runtime/codex-args.ts`.
 
 Dispatcher rows store `bot_secret_ref=config:<dispatcher-id>` for onboarded
 bots; the actual Feishu app secret lives in `feishu.bots.<dispatcher-id>`.
-`dreamux config show` redacts `app_secret` by default. Operators must pass
-`--raw` explicitly to print the unredacted local file.
+`dreamux config show` redacts `app_secret`; there is no CLI raw mode for
+printing the unredacted local file.
 
 ## Consequences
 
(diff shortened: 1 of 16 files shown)


YourWildDad (commented):
## Task 2 review — PASS (posted as comment; cannot self-approve, owner to approve/merge)

Reviewed against the Task 2 requirements, `top-level-design.md`, and the public-repo red line. All seven requirements are implemented and tested; no blocking issues.

### Requirements — all met
1. **MVP Feishu bot schema = only `app_id`/`app_secret`.** `rejectUnknownFeishuBotKeys` whitelists exactly those two keys; `requireNonEmptyString` enforces non-empty. Tested (`rejects unsupported Feishu bot secret fields` for `encrypt_key`, `requires non-empty Feishu app_id and app_secret values`).
2. **`app_id` uniqueness across all declared dispatchers.** Enforced at three entry points: config load (`readFeishuBots` `appIdToDispatcher` map — covers enabled+disabled since `feishu.bots` has no enabled flag), onboard (`assertUniqueFeishuAppIds`), and admin `dispatcher.add` (`duplicateApp`). Tested on config-load and onboard paths.
3. **Config file `0600`.** `assertConfigFileMode` enforced on every `readConfigFile` and on `config show`. Tested (`rejects existing config files that are not mode 0600`).
4. **`serve` loud-fails on missing config.** `serve` switched to `loadConfig` → `readConfigFile` throws "config is missing … run `dreamux onboard`". Verified no production caller auto-creates: `loadOrInitConfig` now has zero `src/` callers (test-only); serve/doctor/onboard-readback/uninstall all use `loadConfig`. Tested (`loadConfig loudly fails when config.json is missing`, asserts file not created).
5. **Redaction + no raw mode.** `config show` `--raw` removed; always `redactConfigForDisplay`. doctor tested not to leak (`does not expose Feishu app secrets`). status and logs are clean by construction (status returns SQLite rows + status.json, which carry `app_id`/`bot_secret_ref` but never `app_secret`; only the config *path* is logged, never its contents).
6. **dispatcher-id charset validation + dedup — the carry-forward from the Task 1 review is fixed.** New `dispatcher-id.ts` (`^[A-Za-z0-9][A-Za-z0-9._-]{0,63}$`, ≤64 chars, no `/`, no leading dot → no traversal/collision); `dispatcherPathSegment` now `validateDispatcherId(id)` and returns it verbatim — the silent `replaceAll` sanitization is gone, so it is a no-op and id↔path can no longer diverge or collide. Validated at config load, onboard, admin, and CLI. Tested.
7. **`serverJsonPath()`** added (`state/server.json`). Tested.

### Red line — clean
All values are placeholders (`app-a`/`secret-flow`/`future-webhook-only`, etc.); all paths are `~/.dreamux`, `tmpdir()`, or `join(root, …)`. No internal domains, registry URLs, `ou_`/`oc_`/`cli_` IDs, infra names, or username-bearing absolute paths.

### Non-blocking follow-ups (track; not Task 2 scope)
- **Config-shape divergence (important to track).** `top-level-design.md` Operator Config shows dispatcher declarations as a `dispatchers: [{id, cwd, enabled, access, codex, feishu}]` array in `config.json`. The code still stores dispatchers in SQLite and only `feishu.bots.<id>` secrets in config. Narrow Task 2 scope (the Feishu secret sub-schema) is satisfied, but whichever task removes SQLite (Task 11) must migrate dispatcher declarations into `config.json` per the design and move/extend this PR's app_id-uniqueness + id-validation onto that array shape. Flagging so it is not lost.
- **README skill-path row is stale.** `packages/dreamux/README.md` still lists `~/.codex/skills/codexmux-dispatcher/SKILL.md` (global + old name); the design installs to workspace-local `<dispatcher cwd>/.codex/skills/dispatcher/SKILL.md`. This PR edits that table — fix the row here or in Task 3.
- **status/logs redaction has no regression test.** Satisfied by construction today; doctor has a no-secret test, status/logs do not. A small assertion each would harden it.
- **Nits:** `assertConfigFileMode` requires exactly `0o600` (a stricter `0o400` would be rejected with a confusing message — the real property is "no group/other access"); `loadOrInitConfig` is now test-only dead code in production and could be retired to prevent accidental reuse; `doctor` throws on missing config rather than reporting it as a graceful check failure.

PR #40, opened 2026-06-03
Task 4: cover Feishu long-connection inbound

YourWildDad:
## Summary
- Add a Feishu bot transport factory seam so tests can drive the same inbound route table that the long-connection transport uses.
- Cover the `im.message.receive_v1` raw-event path: only that event type is registered, non-message events are ignored, routable messages are normalized and dispatched, and unroutable message events are dropped before the handler.
- Assert Feishu app secrets are resolved in `dreamux serve` and do not appear in Codex child options.
- Include the Task 3 follow-up cleanup: remove the obsolete global Codex skills path helper and make `uninstall` warn, then continue, when config permissions or contents are invalid.

## Tests
- `.agents/scripts/check.sh`
- `git diff --check`

Tracking: #36 Task 4.

### DIFF
--- a/packages/dreamux/src/cli/dreamux.ts
+++ b/packages/dreamux/src/cli/dreamux.ts
@@ -265,6 +265,9 @@ function printOnboardResult(result: OnboardRunResult): void {
 }
 
 function printUninstallResult(result: UninstallRunResult): void {
+  for (const warning of result.warnings) {
+    console.error(`warning: ${warning}`);
+  }
   console.log('dreamux uninstall file ledger:');
   for (const entry of result.entries) {
     console.log(`${entry.status}\t${entry.path}\t${entry.reason}`);
--- a/packages/dreamux/src/feishu/bot.ts
+++ b/packages/dreamux/src/feishu/bot.ts
@@ -23,6 +23,7 @@ import {
   narrowMetaFromEvent,
   parseInbound,
   toChannelInbound,
+  type FeishuTransport,
   type Mention,
   type OutboundTarget,
 } from '@excitedjs/feishu-transport';
@@ -69,11 +70,19 @@ export interface CreateBotOptions {
   appSecret: string;
 }
 
-export function createFeishuBot(opts: CreateBotOptions): FeishuBot {
-  const transport = createFeishuTransport({
-    appId: opts.appId,
-    appSecret: opts.appSecret,
-  });
+export interface CreateFeishuBotDeps {
+  createTransport?: (opts: CreateBotOptions) => FeishuTransport;
+}
+
+export function createFeishuBot(
+  opts: CreateBotOptions,
+  deps: CreateFeishuBotDeps = {},
+): FeishuBot {
+  const transport = deps.createTransport?.(opts) ??
+    createFeishuTransport({
+      appId: opts.appId,
+      appSecret: opts.appSecret,
+    });
 
   return {
     get appId(): string {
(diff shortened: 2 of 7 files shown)


YourWildDad (commented):
## Task 4 review — PASS (posted as comment; cannot self-approve, owner to approve/merge)

Reviewed against the Task 4 requirements, `top-level-design.md` (inbound transport + serve-owns-secret), and the public-repo red line. Core deliverables are correct and well-tested; the transport seam does not hollow out production; cleanups are verified. One spec-reconciliation item to resolve by Task 6, plus a minor heads-up.

### Requirements — met
- **Long-connection WebSocket, not webhook.** Production path is real: `createFeishuBot` uses `deps.createTransport?.(opts) ?? createFeishuTransport({appId, appSecret})`, so absent the test dep it builds the live core transport. The core (`feishu-transport/src/transport/feishu.ts`) opens a `lark.WSClient` long-lived WebSocket (confirmed lines ~355/391); no HTTP webhook listener. The injection seam is test-only.
- **Only `im.message.receive_v1`.** `start()` registers exactly one route keyed by `IM_MESSAGE_EVENT_TYPE`. Tested: `Object.keys(transport.routes) === ['im.message.receive_v1']`, a `drive.file.comment_v1` dispatch returns `false` (no handler), and an unroutable receive_v1 (empty `message_id`) is dropped before the handler.
- **Serve is the only secret reader.** `Server.startDispatcher` resolves `botSecret = resolveBotSecret(row.bot_secret_ref, cfg)` in-process and passes it only to `createFeishuBot`; `codexProcessFactory` options never receive it. Smoke test asserts `capturedBotSecrets === ['secret-server-only']` and `JSON.stringify(capturedCodexOptions)` does not contain the secret.
- **Fake-event coverage** exercises connect → route-registration → normalize → dispatch, and the drop path.

### Cleanup carries (from Task 3) — verified
- **`dispatcherCodexSkillsDir` removed cleanly.** Deleted from `paths.ts`; grep confirms zero residual references in `src`/`tests`.
- **uninstall: warn-then-continue, safety intact.** `validateConfigIfPresent` → `warnIfConfigIsNotReadable` (try/catch → push warning → return). Deletion paths use fixed `stateRoot()`/`logsRoot()`/`configDir` (not config-derived), and the hard `assertSafeOwnedDirectory` guard still runs on all three (lines 69-71) and again inside `removeOwnedDirectory` — so a broken/`0644` config now warns but still uninstalls, while `~/.codex`/`~/.claude` are still hard-refused. Tests assert warnings.length===1 + all dreamux-owned paths removed + systemctl disable/daemon-reload still called, including a new `0644` "world-readable config" case.

### Red line — clean
All placeholder ids/secrets (`app-test`/`secret-server-only`/`sender-open-id`/`bot-open-id`); no real `ou_`/`oc_`/`cli_`, no internal domains/registry/hostnames, no username-bearing absolute paths.

### Must resolve by Task 6 (not a Task 4 blocker)
- **`sender_name` has no source in the normalized event.** Task 4's own spec lists `sender_name` in the normalize set and the design's `<feishu_message sender_name="…">` block needs it, but `FeishuInboundEvent` has no `senderName` field and `narrowMetaFromEvent` does not emit one — because `im.message.receive_v1` does not carry the sender display name (only `sender_id.open_id`). So nothing currently produces it, and Task 6 (block emission) will have no source. Decide explicitly: (a) add a best-effort `senderName` (empty when absent) field now so the seam exists, or (b) declare it deferred — Task 6 emits `sender_name=""` and relies on the documented Feishu-skill/`lark-cli` fallback. Either is fine, but it must be a conscious choice, not a silent drop. (Mentions already carry names — the test shows `@_user_1`→`@Ada` — so that path works.)

### Minor heads-up for Task 6
- `create_time` is carried verbatim as an epoch-millis string (`'1710000000000'` in the test), but the design block example shows ISO-8601 (`2026-06-03T00:00:00.000Z`). Decide where the conversion happens when the block is emitted.

PR #41, opened 2026-06-03
Task 5: add Feishu access gate reactions

YourWildDad:
## Summary
- Add dispatcher-local `access.json` state for DM allowlists, group chat/follow-user gates, observed chats, last gate diagnostics, and trust-domain warnings.
- Wire `dreamux serve` so rejected Feishu inbound messages are dropped before enqueue, while accepted messages are enqueued and then receive a channel-owned `GLANCE` reaction.
- Keep received-reaction ownership in serve process memory only, ready for Task 10 cleanup.
- Cover DM/group/follow-user/mention gate paths, rejected-message no-reaction behavior, accepted-message reaction behavior, and multi-chat trust warnings.

## Tests
- `.agents/scripts/check.sh`
- `git diff --check`

Tracking: #36 Task 5.

### DIFF
--- a/.agents/components/repo-structure.md
+++ b/.agents/components/repo-structure.md
@@ -36,6 +36,7 @@ verbatim through the move):
 | Path | Concern |
 |---|---|
 | `src/admin/` | Unix socket admin protocol + method handlers |
+| `src/channel/` | Host-side Feishu gate, access state, outbound target mapping, and received-reaction ownership |
 | `src/cli/` | Entry-point CLIs: `dreamux.ts` (single public command tree), `server.ts` and `server-ctl.ts` as internal delegated modules |
 | `src/codex/` | Codex WS+Unix JSON-RPC client, supervisor, turn collector, init handshake |
 | `src/db/` | Legacy SQLite schema + repository; targeted for removal by [top-level-design](../decisions/top-level-design.md) |
--- a/.agents/decisions/top-level-design.md
+++ b/.agents/decisions/top-level-design.md
@@ -162,9 +162,29 @@ time, admin socket path, and last error.
 child process status, and diagnostic timestamps. It must not contain Feishu
 credentials.
 
-`access.json` stores dispatcher-local access state such as observed chats, gate
-warning inputs, and last gate diagnostics. It must not contain credentials,
-queued inbound messages, or dedupe state.
+`access.json` stores dispatcher-local access state. The MVP shape is:
+
+```json
+{
+  "version": 1,
+  "dm": {
+    "allow_users": ["USER_ID"]
+  },
+  "group": {
+    "allow_chats": ["CHAT_ID"],
+    "follow_users": ["USER_ID"],
+    "require_mention": true
+  },
+  "observed_chats": ["CHAT_ID"],
+  "warnings": [
+    "dispatcher shares one Codex context across multiple Feishu chats"
+  ],
+  "last_gate": null
+}
+```
+
+It must not contain credentials, queued inbound messages, dedupe state, or a
+reaction ledger.
 
 `codex.sock` is the Codex app-server WebSocket-over-Unix-socket endpoint for the
 dispatcher. It is not the Feishu MCP transport. The Feishu MCP default transport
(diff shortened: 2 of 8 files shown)


YourWildDad (commented):
## Task 5 review — PASS (posted as comment; cannot self-approve, owner to approve/merge)

Reviewed against the Task 5 requirements, the updated `top-level-design.md` Access Gate / `access.json` shape, and the public-repo red line. All five requirements are implemented and tested; the adversarial axes are clean on the security-relevant path. Findings below are non-blocking (track as issues).

### Requirements — met
1. **Four admission paths.** `dreamuxFeishuGate`: p2p → `dm.allow_users` allowlist; group → `allow_chats` (when set) ∧ `follow_users` (when set) ∧ `require_mention` (default true, via core `isBotMentioned`). Pre-drops missing-sender / self-bot / bot-sender-type. Only `deliver` reaches enqueue.
2. **Rejected = dropped, not degraded.** On `action: 'drop'` the server logs and `return`s before enqueue, reaction, or any reply. Verified by smoke tests (`inbound.getById(1)` is null, `turnsHandled === 0`, `sentMessages === []`, `reactions === []`).
3. **Trust-domain warning fires correctly and is deduped.** `needsWarning = allow_chats.length > 1 || observed_chats.length > 1`, emitted only when not already in `warnings` — so single-chat → no warning, 2nd distinct chat → warning once, further chats → no repeat. The server only logs when `gate.warning !== null`, so no flooding. Configured-for-multiple (`allow_chats.length > 1`) also warns on first delivery, matching the design's "configured for, or observes, multiple chats."
4. **Received reaction is post-gate, post-enqueue, channel-owned.** Order is gate → (drop returns) → `enqueueInbound` → `addReceivedReaction` only when `inboundId !== null`; reaction id stored in `channelState.receivedReactions` (serve-process Map, no ledger), ready for Task 10. Drop path adds nothing (tested).
5. **Tests** cover every gate path, reject-drop-without-reaction, reaction-on-accept, multi-chat warning + `observed_chats`, and `access.json` 0600 round-trip — all via `createFakeFeishuBot` / fake events, no real credentials.

Important correctness note I verified: the gate **never mutates** the allowlists (`allow_users`/`allow_chats`/`follow_users`/`require_mention`) — it only updates `observed_chats`/`warnings`/`last_gate`. So the admission decision is deterministic and unaffected by any concurrency; only telemetry bookkeeping is mutable. That keeps finding #2 below non-security.

### Red line — clean
Placeholders only (`sender-allowed`/`chat-group-a`/`bot-open-id`/`fake-open-id-app-smoke`); `GLANCE` is a public Feishu emoji key. No real `ou_`/`oc_`/`cli_`, no internal domains/registry/hostnames, no username paths.

### Task 4 carries — no regression
`normalizeInboundEvent` (sender_name, create_time) is untouched; bot.ts only gains `addReaction`. Those remain Task 6.

1. **No operator surface populates `access.json`.** `loadDispatcherAccess` reads only `access.json`; nothing seeds it — not onboard, not an admin command, and the design's `config.access.{allow_users,allow_group_chats}` is not bridged to it. Net effect of the default state: **all DMs are dropped** (empty `dm.allow_users`) while group + @mention works out of the box (secure default). That's fine for the group@mention MVP flow, but DMs and restrictive allowlists are currently unconfigurable. Confirm which task owns the access-management surface (admin CLI, or `config.access`→`access.json` seeding in onboard / the Task 11 config-shape migration).
2. **`access.json` is reloaded and rewritten on *every* inbound event** (`loadDispatcherAccess` → mutate → `saveDispatcherAccess`, including drops, because `last_gate` updates each time). Two effects: (a) a disk read+write per message; (b) under concurrent inbound, interleaved load/save can lose an `observed_chats` entry or the warning-dedup write (→ a delayed or duplicated warning). It is **telemetry-only** (allowlists are never written, so admission is safe), but recommend holding the access state in memory like `channelState` — loaded once at dispatcher start, persisted on change — to remove both the per-message I/O and the race.
3. **`require_mention` is a dispatcher-global flag, not per-chat.** The design's allowlist mode allowed a per-group "may require a bot mention"; this simplifies to one toggle. Reasonable MVP simplification and the doc was updated to match — flagging for awareness.
4. **`last_gate` write-amplification.** Persisting the last gate decision on every message is heavy for a debug aid; consider in-memory or throttling (folds into #2 if state moves in-memory).
5. **doctor/status rendering of the warning isn't wired here.** The warning is persisted to `access.json.warnings` and logged to stderr (available), but the design's "doctor or status surfaces must warn" rendering is not in this PR — confirm Task 12 reads `access.json.warnings`.

PR #43, opened 2026-06-03
Task 6: format Feishu inbound for Codex

YourWildDad:
## Summary
- Format accepted Feishu inbound events as Codex-facing `<feishu_message>` blocks before enqueueing them into the dispatcher turn path.
- Normalize Feishu `create_time` values to ISO strings and render Feishu mention placeholders as `<at id="...">Display Name</at>` tags.
- Add a best-effort `senderName` seam; native `im.message.receive_v1` events without a display name intentionally emit `sender_name=""`.
- Include a Feishu skill fallback note when message text cannot be parsed from the inbound content.

## Tests
- `.agents/scripts/check.sh`
- `git diff --check`

Tracking: #36 Task 6.

### DIFF
--- a/.agents/decisions/top-level-design.md
+++ b/.agents/decisions/top-level-design.md
@@ -271,6 +271,11 @@ When rich content parsing fails, the dispatcher still passes `message_id`,
 `chat_id`, `sender_id`, and `sender_name` into the Codex turn and instructs Codex
 to use the Feishu skill and `lark-cli` fallback to fetch message text.
 
+`sender_name` is a best-effort seam. Feishu `im.message.receive_v1` does not
+provide a sender display name in the native event envelope today, so the MVP
+emits `sender_name=""` unless a later channel enricher supplies one. Codex should
+use the Feishu skill fallback when it needs a human-readable sender name.
+
 Messages rejected by the access gate are discarded. Rejected messages do not
 enter the Codex thread.
 
--- a/packages/dreamux/src/channel/feishu-message.ts
+++ b/packages/dreamux/src/channel/feishu-message.ts
@@ -1,0 +1,105 @@
+import type { Mention } from '@excitedjs/feishu-transport';
+
+import type { FeishuInboundEvent } from '../feishu/bot.js';
+
+export const FEISHU_SKILL_FALLBACK_NOTE =
+  'Parser note: message text may be incomplete. Use the Feishu skill with the chat_id and message_id above to fetch the original message when needed.';
+
+export function formatFeishuMessageForCodex(event: FeishuInboundEvent): string {
+  const attrs: Array<[string, string]> = [
+    ['chat_id', event.chatId],
+    ['chat_type', event.chatType],
+    ['message_id', event.messageId],
+    ['sender_id', event.senderId],
+    ['sender_name', event.senderName],
+    ['create_time', formatFeishuCreateTime(event.createTime)],
+  ];
+  const body = renderMessageBody(event);
+  const fallback = shouldAddFallbackNote(event)
+    ? `\n\n${FEISHU_SKILL_FALLBACK_NOTE}`
+    : '';
+
+  const attrLines = attrs.map(
+    ([key, value]) => `  ${key}="${escapeXmlAttribute(value)}"`,
+  );
+  attrLines[attrLines.length - 1] = `${attrLines[attrLines.length - 1]}>`;
+
+  return [
+    '<feishu_message',
+    ...attrLines,
+    `${body}${fallback}`,
+    '</feishu_message>',
+  ].join('\n');
+}
+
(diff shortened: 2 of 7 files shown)


YourWildDad (commented):
## Task 6 review — PASS (posted as comment; cannot self-approve, owner to approve/merge)

Reviewed against the Task 6 requirements, the Task 4/5 carries, and — most importantly — the prompt-injection surface (attacker-controlled Feishu body/sender-name/mention spliced into the Codex `<feishu_message>` block). The injection defense is solid and tested. All six requirements are implemented. Findings below are non-blocking.

### Injection axis — the key check, PASSES
Every attacker-influenceable value reaches the block only through escaping:
- **Body** (`renderMessageBody`) starts with `escapeXmlText(text)` (`&`→`&amp;` first, then `<`/`>`), so a body containing `</feishu_message>`, `<at id="ou_evil">…`, or any tag becomes inert `&lt;…&gt;` text — it cannot close the block, forge an `<at>` tag, or inject markup. Newlines in the body stay inside the block for the same reason.
- **Attributes** (`chat_id`/`chat_type`/`message_id`/`sender_id`/`sender_name`/`create_time`) all go through `escapeXmlAttribute` (= text escape + `"`→`&quot;`), so a value cannot break out of its `="…"` quotes.
- **Generated `<at>` tags** escape both the id (`escapeXmlAttribute`) and the display name (`escapeXmlText`), so mention name/id can't inject either.
- No raw/unescaped attacker text path into the block exists; the fallback note is a static constant.
The block-escaping is covered by a test asserting body `& <ok>` renders as `&amp; &lt;ok&gt;` and sender_name `Ada & Bob` as `Ada &amp; Bob`.

### Requirements — met
1. **Formatted before enqueue.** `server.ts` enqueues `parsed_text: formatFeishuMessageForCodex(event)` (raw event still stored separately in `feishu_event_json`). Smoke test asserts `parsed_text` contains `<feishu_message`.
2. **`sender_name` best-effort seam (carry #1).** New `FeishuInboundEvent.senderName`; `extractSenderName` probes `sender_name`/`display_name`/`name`/`user_name` and returns `''` when absent. The block always emits `sender_name="…"` explicitly (empty, not dropped). Tested both ways (`'Ada Sender'` present → filled; absent → `''`).
3. **`create_time` epoch→ISO (carry #2).** `formatFeishuCreateTime`: UTC ISO via `toISOString()` (always `Z`); seconds-vs-ms heuristic at the 1e12 boundary (safe for all realistic Feishu timestamps — a ms value <1e12 is pre-2001, a seconds value ≥1e12 is year 33000+); empty→`''`; non-numeric/unparseable→passthrough (no crash, no `Invalid Date`); huge values fall through to passthrough rather than emitting `Invalid Date`. Tested for ms and seconds.
4. **Mentions → `<at id>Name</at>`.** Multiple mentions handled; id falls back open_id→union_id→user_id; when name/id/key is missing the mention is skipped (the literal key stays as escaped text) — no crash, no dropped message.
5. **Fallback note.** `shouldAddFallbackNote` triggers on unparseable text / non-text placeholders; the note references only chat_id/message_id (already in the block) — no secret leak.
6. **Tests** cover block format, sender_name both branches, create_time ISO, fallback note, and the escaping case, all on fake data.

### Red line — clean
Placeholders only (`sender-open-id`/`mentioned-user`/`Ada`/`app-smoke`); generic probe field names; no real `ou_`/`oc_`/`cli_`, no internal domains/registry/hostnames/username paths. (`1710000000000`→`2024-03-09T16:00:00.000Z` is correct.)

1. **Mention key prefix collision (real bug, low-frequency).** `renderTextWithMentions` does `out.split(key).join(<at…>)` per mention with no longest-first ordering. Feishu keys are `@_user_1`, `@_user_2`, …, `@_user_10`, `@_user_11`, so `@_user_1` is a prefix of `@_user_1X`. With ≥10 mentions, replacing `@_user_1` first corrupts `@_user_11` (→ `<at>…</at>1`, and that mention isn't rendered). Fix: sort mentions by key length descending before substitution, or do a single-pass tokenized replace. Not a security issue (still escaped), but a rendering bug for messages with 10+ mentions.
2. **Attribute newline hardening (minor, not an injection).** Attribute values escape `<>&"` but not newlines/control chars; a newline in a value would split the attr line cosmetically (the real quotes still hold — no structural escape). Currently inert (`sender_name` is always `''`, `create_time` is structured), but when the `sender_name` enricher lands, strip/escape newlines+control chars in attribute values.
3. **Test gaps.** No tests for `create_time` empty/garbage/0/negative, mention name-undefined fallback, multiple/colliding mentions, or non-text message types. The core escaping case is tested; these are robustness additions.
4. **(very minor)** garbage `create_time` is passed through verbatim into the attribute rather than normalized to `''`; defensible (surfaces what Feishu sent) but inconsistent with the empty-string case.

PR #45, opened 2026-06-03
Task 7: use in-memory dispatcher turn queue

YourWildDad:
## Summary
- Move accepted Feishu inbound work from the legacy SQLite inbound buffer into a process-local dispatcher turn queue.
- Serialize one turn per dispatcher, coalesce pending same-chat bursts into one Codex turn, and reply to the newest message in a coalesced batch.
- Add a bounded process-local `message_id` dedupe window so Feishu redelivery does not enqueue duplicate turns or received reactions in the same server process.
- Stop replaying legacy persisted inbound rows at dispatcher startup; the old table remains for the later legacy cleanup task.

## Tests
- `.agents/scripts/check.sh`
- `git diff --check`

Tracking: #36 Task 7.

### DIFF
--- a/packages/dreamux/src/channel/outbound.ts
+++ b/packages/dreamux/src/channel/outbound.ts
@@ -1,5 +1,3 @@
-import type { InboundRow } from '../db/types.js';
-
 export interface ChannelOutboundTarget {
   /** Stable channel-local conversation id. */
   conversationId: string;
@@ -16,7 +14,15 @@ export interface OutboundSink {
   send(target: ChannelOutboundTarget, text: string): Promise<string[]>;
 }
 
-export function outboundTargetForInbound(row: InboundRow): ChannelOutboundTarget {
+export interface InboundOutboundSource {
+  source_chat_id: string;
+  source_message_id: string | null;
+  sender_id: string | null;
+}
+
+export function outboundTargetForInbound(
+  row: InboundOutboundSource,
+): ChannelOutboundTarget {
   return {
     conversationId: row.source_chat_id,
     ...(row.source_message_id !== null
--- a/packages/dreamux/src/dispatcher/runtime.ts
+++ b/packages/dreamux/src/dispatcher/runtime.ts
@@ -10,9 +10,8 @@
  *
  * Lifecycle: declared → starting → ready → (degraded) → stopping → stopped.
  *
- * Issue #2 §"崩溃与异常恢复":
- *   - On startup, mark stale `running` rows as `unknown` before opening
- *     inbound — at-most-once.
+ * Current MVP:
+ *   - accepted inbound work is process-local and is dropped on restart;
  *   - thread/resume failure does not degrade the whole dispatcher; we
  *     start a fresh thread, record the lost one in last_lost_thread_id,
  *     and post a visible warning to the next source chat.
@@ -22,10 +21,7 @@ import { mkdirSync } from 'node:fs';
 import { dirname } from 'node:path';
 
 import type { DispatcherRow, DispatcherStatus } from '../db/types.js';
-import type {
-  DispatcherRepo,
-  InboundRepo,
-} from '../db/repository.js';
+import type { DispatcherRepo } from '../db/repository.js';
 import { CodexProcess, type CodexProcessOptions } from '../codex/supervisor.js';
 import { CodexWsClient } from '../codex/rpc.js';
 import { performInitializeHandshake } from '../codex/handshake.js';
@@ -33,7 +29,7 @@ import type {
   ThreadResumeResponse,
   ThreadStartResponse,
 } from '../codex/types.js';
-import { TurnManager } from './turn-manager.js';
+import { TurnManager, type InboundTurnInput } from './turn-manager.js';
 import { createFailFastApprovalHandler } from './approval.js';
 import {
   dispatcherCodexCwd,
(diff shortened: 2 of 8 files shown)


YourWildDad (commented):
## Task 7 review — PASS (posted as comment; cannot self-approve, owner to approve/merge)

Reviewed against the Task 7 spec, top-level-design D3/D4, and the adversarial axes (boundedness, deadlock, coalesce edges, dedupe race). All five requirements are correctly implemented. One genuine boundedness gap remains (your recurring concern) — it matches the design's documented Known Limitation, so I'm flagging it for an elevate-to-must-fix decision rather than blocking. Details below.

### Requirements — met
1. **Serial execution, guaranteed.** A single `drainLoop` with a `running` guard, and `notify()` only schedules a new loop when `!running && !drainScheduled` (otherwise it resumes the existing waiter via `wakeup`). JS is single-threaded and the guard checks have no intervening `await`, so two turns can never run concurrently. `processBatch` is `await`ed in the loop.
2. **Coalesce merges *pending*, not in-flight, and is ordered.** The running batch is `shift()`-ed off `this.queue` before processing, so `enqueue`'s `queue.find(same chat)` only ever merges into a not-yet-running batch. Messages append in arrival order; `batchPrompt` joins `parsed_text` in order; `source_message_id`/`sender_id` advance to the newest so the reply targets the latest message. Verified by `turn-manager.test` (`['first\n\nsecond']`, reply→msg-2) and the smoke coalescing test (chat-b 2-block batch behind chat-a, reply→msg-b2). Cross-chat batches stay FIFO and serialized; a chatty chat folds into its one slot, so no starvation.
3. **Bounded dedupe, correct eviction.** `seenMessageIds` Set + `seenMessageIdOrder` FIFO, window `DEFAULT_MESSAGE_ID_DEDUPE_WINDOW = 1024`, oldest-evicted-first; a redelivery within the window is dropped, one evicted past the window is reprocessed (documented). Tested at window=2. FIFO-1024 comfortably covers Feishu's near-immediate at-least-once redelivery. Empty/null ids bypass dedupe but `normalizeInboundEvent` already guarantees a non-empty message_id by enqueue time.
4. **No legacy replay.** `recoverInboundOnStartup` (running→unknown sweep + user ping), `retryPendingOutbound`, and the startup `notify()` backlog kick are all removed; the `inbound` repo dep is dropped from the runtime and turn-manager; `enqueueInbound` no longer writes the SQLite table. Tests confirm a pre-seeded `running` row stays `running` (no sweep, no user message) and a pre-seeded `queued` row stays `queued` (no drain). Legacy table left intact for Task 11.
5. **Tests** cover serial+coalesce, dedupe+eviction, no-replay (both running and queued), all on fake codex/bot.

### Adversarial axes
- **Deadlock: none.** A turn that throws is caught, best-effort-notifies the chat, and the loop continues to the next batch. A turn that *hangs* blocks the dispatcher queue (no turn timeout) — this is exactly D5's accepted behavior; recovery is dispatcher stop/start (kills the codex child → `runTurn` WS rejects → catch → loop frees). `stop()` clears the queue and wakes the waiter.
- **Dedupe race: none.** `enqueue` is fully synchronous (rememberMessageId → queue mutate → `notify` which only schedules); check-then-add cannot interleave on a single thread.
- **Same-tick coalescing** works because `notify` defers the first drain via `setTimeout(0)`, letting same-tick bursts merge before draining — nicely done.

### Red line — clean
Placeholders only (`chat-group-a`/`msg-b2`/`sender-test`/`thread-1`); Chinese strings are user-facing templates. No internal IDs/domains/registry/hostnames/username paths.

### Findings
1. **[Highest value — your recurring boundedness concern] The work queue and per-chat batch are still genuinely unbounded.** `enqueue` has no max queue depth and no max batch size; `pending.messages.push` grows without limit, and `batchPrompt` joins them all into one prompt. The dedupe window is bounded, but the *work* isn't. Combined with D5 (no turn timeout), a stuck/slow turn + a chatty chat grows memory until OOM, and a giant coalesced batch can blow Codex's context independently of memory. This **matches the design's documented Known Limitation** ("per-chat batches and the cross-chat queue are in-memory and not globally size-bounded in the MVP… a stuck turn plus continued inbound can grow memory until restart"), so the code is not in violation of the current spec — but it is the genuine boundedness gap flagged across the last two reviews. Recommend a cheap bound now: cap messages-per-batch and total-queued, drop-oldest-in-batch (or drop-with-log) on overflow. Decision for the owner: add the bound (honor "boundedness = must-fix") or consciously keep the Known Limitation.
2. **(staging, not a Task 7 defect) Auto-outbound still active.** `processBatch` → `sendOutbound(batch, assistantText)` auto-forwards Codex assistant text to Feishu, which contradicts top-level-design's "model text is never automatically forwarded / MCP-reply-only." This is legacy correctly preserved for now — removing it before Task 10's MCP `reply` lands would mute the dispatcher (and the smoke tests depend on it). It is explicitly slated for removal in Task 11 ("移除自动 assistant-text 出站"). Surfacing so it stays tracked; the empty-turn placeholder (`本轮没有文本回复。`) goes away with it.
3. **(minor)** A batch dequeued with `threadId === null` is logged and silently dropped (no user notification). Invariant-violation path ("ready" implies a thread), acceptable, but a chat ping would be friendlier.
4. **(minor test gap)** No explicit test for stuck-turn-blocks-queue or cross-chat non-starvation; coalesce-behind-running-turn is covered via smoke.

PR #47, opened 2026-06-03
Bound Feishu channel ingress delivery

YourWildDad:
## Summary
- Add a channel-side per-dispatcher delivery buffer for Feishu inbound messages.
- Apply a 1 second jitter window before dispatcher delivery.
- Enforce a per-delivery-thread ingress limit of 10 accepted messages per 10 seconds; overflow is rejected with a log, does not enter Codex, and receives no channel-owned received reaction.
- Move process-local `message_id` dedupe to the channel layer so redelivery is rejected before received reactions.
- Update the top-level design record and tests for the new channel contract.

## Codex steering research
- Codex app-server's steering primitive is `turn/steer`: it appends user input to an already in-flight regular turn and returns the active `turnId`.
- dreamux currently drives Codex through `runTurn()`, which sends `turn/start` with `[{ type: "text", text, text_elements: [] }]` and waits for `turn/completed`.
- Directly using `turn/steer` as the only delivery path is not viable for channel flushes: it requires an active regular turn and a non-empty matching `expectedTurnId`; it fails when the thread is idle, when the active turn id has changed, when the active turn is non-steerable such as review/compact, or when input is too large. It also does not accept turn settings overrides.
- This PR therefore bounds channel ingress before the existing dispatcher turn path. Active-turn steering can be added later as a separate optimization, after outbound semantics are moved to the MCP reply path so steered cross-chat input does not inherit the current temporary automatic outbound target.

## Tests
- `.agents/scripts/check.sh`
- `git diff --check`

Related: #46.

### DIFF
--- a/.agents/decisions/top-level-design.md
+++ b/.agents/decisions/top-level-design.md
@@ -230,8 +230,17 @@ out of scope for the MVP.
 The MVP handles `im.message.receive_v1`. Other Feishu event kinds are ignored
 until a later decision adds them.
 
-Accepted messages enter one per-dispatcher in-memory queue. The queue is
-serialized: only one Codex turn runs per dispatcher at a time.
+After the access gate accepts a message, the channel layer applies a
+per-dispatcher delivery buffer before handing work to the dispatcher runtime:
+
+- a 1 second jitter window batches bursty inbound messages before delivery;
+- each dispatcher delivery thread accepts at most 10 messages per 10 seconds;
+- messages beyond that rate are explicitly rejected and logged by the channel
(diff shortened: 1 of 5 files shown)


YourWildDad:
MVP 版本先不做任何有界限制（用户 2026-06-04 决策）。channel 层 jitter / 限流 / dedupe 作为后续追加需求，已 defer 到 #46，等当前整批任务清完再做。本 PR 先关闭，分支 `codex/channel-jitter-bounds` 保留供后续参考；研究结论已记到 #46。

PR #48, opened 2026-06-03
Task 8: restart Codex child runtimes

YourWildDad:
## Summary

- Add Codex child-process exit observation and dispatcher-owned restart scheduling.
- Restart a dispatcher Codex runtime with backoff only after child exit or child WebSocket close.
- Resume the saved/current thread id after restart instead of starting a fresh thread.
- Keep slow in-flight turns from triggering restarts; manual dispatcher stop/start cancels pending restart work and can resume the saved thread.
- Harden fake Codex teardown so tests can simulate WebSocket death.

## Tests

- `.agents/scripts/check.sh`
- `git diff --check`

Tracking: #36 Task 8.

### DIFF
--- a/packages/dreamux/src/codex/supervisor.ts
+++ b/packages/dreamux/src/codex/supervisor.ts
@@ -47,6 +47,13 @@ export interface CodexProcessOptions {
   readyTimeoutMs?: number;
 }
 
+export interface CodexProcessExit {
+  code: number | null;
+  signal: NodeJS.Signals | null;
+}
+
+export type CodexProcessExitHandler = (exit: CodexProcessExit) => void;
+
 /** A handle to one running codex app-server child process. */
 export class CodexProcess {
   readonly socketPath: string;
@@ -54,6 +61,7 @@ export class CodexProcess {
   private child: ChildProcess | null = null;
   private _pid: number | null = null;
   private reaped = false;
+  private readonly exitHandlers: CodexProcessExitHandler[] = [];
 
   constructor(private readonly opts: CodexProcessOptions) {
     this.socketPath = opts.socketPath;
@@ -62,6 +70,10 @@ export class CodexProcess {
 
   get pid(): number | null {
     return this._pid;
+  }
+
+  onExit(handler: CodexProcessExitHandler): void {
+    this.exitHandlers.push(handler);
   }
 
   /** Spawn the daemon and resolve once its listen socket is bound. */
@@ -129,6 +141,16 @@ export class CodexProcess {
     // Future post-spawn `error` emissions must not crash the supervisor.
     child.on('error', () => {
       /* daemon-side error, can no longer affect this process */
+    });
+    child.once('exit', (code, signal) => {
+      if (this.reaped) return;
+      for (const handler of this.exitHandlers) {
+        try {
+          handler({ code, signal });
+        } catch {
+          /* exit observers must not poison process event dispatch */
+        }
+      }
     });
 
     try {
--- a/packages/dreamux/src/dispatcher/runtime.ts
+++ b/packages/dreamux/src/dispatcher/runtime.ts
@@ -22,7 +22,11 @@ import { dirname } from 'node:path';
 
 import type { DispatcherRow, DispatcherStatus } from '../db/types.js';
 import type { DispatcherRepo } from '../db/repository.js';
-import { CodexProcess, type CodexProcessOptions } from '../codex/supervisor.js';
+import {
+  CodexProcess,
+  type CodexProcessExit,
+  type CodexProcessOptions,
+} from '../codex/supervisor.js';
 import { CodexWsClient } from '../codex/rpc.js';
 import { performInitializeHandshake } from '../codex/handshake.js';
 import type {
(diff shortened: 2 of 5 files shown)


YourWildDad (commented):
## Task 8 review — PASS (posted as comment; cannot self-approve, owner to approve/merge)

Reviewed against the Task 8 spec, top-level-design's Dispatcher Lifecycle (D5), and the adversarial axes (restart/resume correctness, slow-turn-vs-crash, in-flight races, leaks, carry-forwards). The restart/resume core is carefully built and correct; no blocking issues. Findings are non-blocking.

### Completeness — all requirements met
1. **One child/thread per dispatcher** preserved; `startCodexRuntime` is the single bring-up path used by both `start()` and `restartCodexRuntime`.
2. **Save/restore thread_id, correct on both paths.** `resolveThread` now reads `this.threadId ?? this.row.thread_id`. Auto-restart keeps `this.threadId` across `teardownCodexRuntime` (which does not null it) → `thread/resume` same session; manual stop→start builds a fresh runtime and resumes via the DB-persisted `row.thread_id` (re-fetched in `doStartDispatcher`). Resume failure still falls back to a fresh thread + `recordLostThread` (visible degradation).
3. **PATH injection** unchanged (`dispatcherProcessEnv`).
4. **No turn timeout.** Restart is triggered *only* by `process.onExit` or `client.onClose` — there is no timeout heuristic that could misclassify a long turn. The slow-turn test confirms status stays `ready` and no respawn.
5. **child/WS death → backoff restart**, capped (`base*2^(n-1)`, default 1s→30s cap). 6. **Manual stop/start recovery** works.
Tests cover child-exit-restart-resume, manual-stop-cancels-pending-restart + start-resumes, WS-death-restart-resume, and slow-turn-no-restart.

### Correctness I verified (the subtle parts)
- **Intentional teardown never triggers a spurious restart:** `teardownCodexRuntime` sets `this.client=null`/`this.process=null` *before* `close()`/`reap()`, so the `if (this.client !== client)` / `if (this.process !== process)` identity guards drop the self-inflicted close/exit. Belt-and-suspenders: `reap()` sets `reaped=true` first, so `CodexProcess`'s own `exit` handler short-circuits.
- **Double-fire (child exit + WS close from the same death) → single restart:** `scheduleRestart` guards on `restartTimer !== null || restarting`.
- **No restart storm:** backoff is capped; `restartAttempts` resets to 0 only on success; one timer at a time.
- **No duplicate turn:** an in-flight turn whose child dies rejects in `runTurn` → `processBatch` catch sends the user a failure message and returns; the batch is not requeued.
- **`reap()` is a real kill** (SIGTERM→SIGKILL of the process *group*) and removes the socket file — no child/socket leak on restart.
- **stop-during-restart does not leak a child:** `restartCodexRuntime` re-checks `this.stopping` after `teardownCodexRuntime` and again after `startCodexRuntime` (tearing down a freshly-spawned process if stop raced in).

### Red line — clean
Only placeholders (`thread_fake_`, `app-smoke`, `chat-group-a`); user-facing Chinese strings carry no secrets. No internal IDs/domains/registry/hostnames/username paths.

1. **Inbound during the restart window is silently dropped with no feedback.** While `restartCodexRuntime` runs, `this.turnManager === null`, so `enqueueInbound` returns `false` — the server handler then skips the received reaction and sends nothing. The user gets no reaction and no error during a (degraded) restart. Worse, `false` is the same signal as a dedupe-drop, so it's indistinguishable. Consider buffering or at least a transient "busy/degraded" reaction.
2. **A child restart drops queued + in-flight inbound.** `teardownCodexRuntime` stops the old `TurnManager` (clears its queue) and builds a fresh empty one. The in-flight turn gets a failure message, but already-queued batches are dropped silently — and their received reactions were already added, so this *widens* the documented "dangling received reaction" limitation from a full server restart to every transient Codex child crash. It's consistent with the in-memory lossy model (D4), but since the server process is still alive, the queue could be preserved across a child restart. Worth an explicit decision.
3. **The restart `catch` sets `degraded` without re-checking `stopping`.** If a manual `stop()` races into the brief restart *execution* window, `startCodexRuntime` throws (its client was closed by stop), and the catch overwrites the status with `degraded` after stop set `stopping`/`stopped`. No leak (teardown reaps the process; the catch's reschedule is suppressed by the `stopping` guard), but the final status row can read `degraded` instead of `stopped`. One-line fix: `if (this.stopping) return;` at the top of the catch before `setStatus('degraded')`.
4. **No give-up / log throttle on permanent failure.** A persistently-broken Codex (e.g. bad binary) retries every 30s forever, logging a warn each time. Fine for a daemon, but consider a `degraded`-and-stop threshold or log throttling.
5. **The dedupe window is TurnManager-local, not process-local.** A child restart builds a fresh `TurnManager` with an empty `seenMessageIds`, so a Feishu redelivery straddling a restart can be processed twice — narrower than the "within this server process" wording in the design. Minor.

### Carry-forwards
Task 11 items (SQLite removal, auto-outbound removal, config-shape migration, residual-naming cleanup) are untouched and not regressed by this PR; thread_id persistence still rides the legacy repo as before. Still tracked under Task 11.

PR #50, opened 2026-06-03
Task 9: add Feishu MCP stdio shim

YourWildDad:
## Summary

- Add `dreamux feishu-mcp --dispatcher <id>` as a dispatcher-scoped stdio MCP shim.
- Expose `reply` and `react` MCP tools and forward tool calls over the local admin socket with dispatcher routing metadata only.
- Inject `mcp_servers.feishu` Codex config overrides after operator Codex args when starting a dispatcher app-server.
- Add explicit `mcp.reply` / `mcp.react` admin IPC placeholders; Task 10 will attach these to serve-owned Feishu outbound and reaction cleanup.
- Document the MVP MCP tool parameter shape in the top-level design.

## Codex MCP spike

Before coding, I validated the default stdio path against `codex-cli 0.136.0`: `codex app-server` accepted `mcp_servers.<name>.command` and `mcp_servers.<name>.args` config overrides, completed the app-server initialize handshake, then `mcpServerStatus/list` started the injected stdio MCP server and returned its `reply` tool. This keeps Task 9 on the STDIO design path; no HTTP fallback is needed.

## Tests

- `.agents/scripts/check.sh`
- `git diff --check`
- staged diff red-line scan
- local `dreamux feishu-mcp` initialize smoke

Tracking: #36 Task 9.

### DIFF
--- a/.agents/decisions/top-level-design.md
+++ b/.agents/decisions/top-level-design.md
@@ -344,10 +344,12 @@ or a world-readable file. An unauthenticated HTTP fallback is not allowed.
 
 The MVP MCP tool surface is:
 
-- `reply`: send a message to a Feishu chat, using `chat_id` and optionally
-  `message_id` so topic-mode replies stay in the original topic when Feishu
-  supports it.
-- `react`: add a model-owned reaction to a Feishu message.
+- `reply`: send a message to a Feishu chat. Parameters are `chat_id`, `text`,
+  optional `message_id`, and optional `mention_user_ids`. The `message_id`
+  should be the inbound Feishu message id when the model wants Feishu to keep
+  topic-mode replies under the original topic.
+- `react`: add a model-owned reaction to a Feishu message. Parameters are
+  `message_id` and `emoji`.
 
 `edit_message` and model-owned `remove_reaction` are out of scope for the MVP.
 
--- a/packages/dreamux/src/admin/client.ts
+++ b/packages/dreamux/src/admin/client.ts
@@ -1,0 +1,119 @@
+import { connect, type Socket } from 'node:net';
+
+import { adminSocketPath as defaultAdminSocketPath } from '../runtime/paths.js';
+import type { AdminRequest, AdminResponse } from './protocol.js';
+
+export interface SendAdminRequestOptions {
+  socketPath?: string;
+  timeoutMs?: number;
+  requestId?: string;
+}
+
+export class AdminClientError extends Error {
+  constructor(
+    public readonly code: string,
+    message: string,
+  ) {
+    super(message);
+    this.name = 'AdminClientError';
+  }
+}
+
+const DEFAULT_TIMEOUT_MS = 10_000;
+let nextRequestId = 1;
+
+export function sendAdminRequest(
+  method: string,
+  params: Record<string, unknown>,
+  options: SendAdminRequestOptions = {},
+): Promise<unknown> {
+  const socketPath = options.socketPath ?? defaultAdminSocketPath();
+  const request: AdminRequest = {
+    id: options.requestId ?? adminRequestId(),
+    method,
+    params,
+  };
+  return sendOne(socketPath, request, options.timeoutMs ?? DEFAULT_TIMEOUT_MS);
+}
+
+export function sendOneAdminRequest(
+  socketPath: string,
+  request: AdminRequest,
+  timeoutMs: number = DEFAULT_TIMEOUT_MS,
+): Promise<unknown> {
+  return sendOne(socketPath, request, timeoutMs);
+}
+
+function sendOne(
+  socketPath: string,
+  request: AdminRequest,
+  timeoutMs: number,
+): Promise<unknown> {
+  return new Promise((resolve, reject) => {
+    let buf = '';
+    let settled = false;
+    let sock: Socket | null = null;
+    const timer = setTimeout(() => {
+      settle(new Error(`admin socket request timed out after ${timeoutMs}ms`));
+      try {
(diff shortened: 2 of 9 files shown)


YourWildDad (commented):
## Task 9 review — PASS (posted as comment; cannot self-approve, owner to approve/merge)

Reviewed against the top-level-design Feishu MCP / shim contract, the Task 9/10 boundary, the spike claim, and the adversarial axes (admin IPC auth, serve-not-ready, shim lifecycle vs Task 8 restart, secret isolation, reply-failure surfacing). The shim is a correct, secure, honest partial implementation. No blocking defects. One strong should-fix (the spike is not codified) plus minor findings.

### Completeness — met, and honestly partial
- **Thin shim, verified.** `runFeishuMcp` imports only the admin client, `runtime/paths` (for the default socket path), and `dispatcher-id`. It does **not** import config/secrets/Feishu transport; it never builds a Feishu client and holds no channel/reaction state. The injected argv (`feishuMcpCodexArgs`) carries only `--dispatcher <id>` and `--admin-socket <path>` — no `app_id`, no `app_secret`. Confirmed the only process that reads Feishu credentials is `serve` (`resolveBotSecret`). 
- **stdio MCP → admin.sock forward.** Handles `initialize` / `(notifications/)initialized` / `tools/list` / `tools/call`; `reply`/`react` forward to `mcp.reply`/`mcp.react` over the admin socket with `{dispatcher_id, ...args}` routing metadata. Tested.
- **Task 9/10 boundary is clean and honest.** Serve-side `mcp.reply`/`mcp.react` validate params then `throw AdminError('NOT_IMPLEMENTED', '…reserved for Task 10…')`. The shim surfaces that as an MCP tool error (`isError:true`, `[NOT_IMPLEMENTED] …`). So end-to-end a `reply` call returns a clear not-implemented tool error rather than pretending to work. Honest.

### Adversarial axes — checked
- **admin IPC auth holds.** `createAdminSocketServer` hard-chmods the socket to `0600` and *refuses to start* if it can't (fail-fast), so the "no bearer token, file-permission scoped" argument is real — only the same uid can connect. The shim is just another owner-uid client of the existing admin trust boundary.
- **serve not ready / down.** `serve` binds the admin socket before spawning dispatchers, so the shim (spawned later by the codex child) always finds it. If serve is down/restarting (Task 8) or the socket is stale, `sendAdminRequest` maps `ENOENT`/`ECONNREFUSED`/timeout to an error → `forwardToolCall` returns an MCP tool error. No crash.
- **shim lifecycle vs Task 8 restart.** The shim is a child of the codex app-server; Task 8's `reap()` SIGKILLs the process *group*, so the old shim dies with the old codex child, and the restarted codex re-spawns a fresh shim from the same injected config. The admin client opens a short-lived connection per RPC (`connect`→write→read one line→`end`), so there is no long-lived/stale socket handle to leak across restarts.
- **reply failure is not silent.** All admin failures (incl. NOT_IMPLEMENTED, BAD_REQUEST, timeout, serve-down) become MCP tool errors. Tested.

### Red line — clean
Placeholders only (`dispatcher-a`/`chat-a`/`THUMBSUP`/`/tmp/not-used.sock`). No internal IDs/domains/registry/hostnames/username paths.

### Findings
1. **[Strong should-fix — recommend as a near-term gate] The Codex 0.136 MCP spike is asserted, not codified.** The PR proves the *shim* and the *arg format/ordering* in isolation (fake admin server + fake codex), but the load-bearing claim — that codex actually accepts `mcp_servers.feishu.command/args`, completes initialize at protocol `2024-11-05`, and `mcpServerStatus/list` starts the shim and lists `reply` — has no automated/live-codex test. The repo already has the opt-in-skippable `tests/codex-0135-live.test.ts` pattern (and a CLAUDE.md rule that live-codex tests fail loudly) to follow. This matters more than usual because `resolveExtraArgs` now injects the MCP config **unconditionally for every dispatcher**: if the override format is wrong or a future codex bump changes it, **all dispatcher startups fail**, not just outbound — undetected until a human notices. Recommend a `codex-0136-mcp-live` test (spawn real codex with `feishuMcpCodexArgs`, assert `mcpServerStatus/list` returns `reply`) before relying on the integration / before Task 10 ships real outbound. (Alternatively, gate the injection behind a verified-codex check until that test exists.)
2. **[minor] Tool-argument validation errors surface as JSON-RPC `-32603`, not MCP tool errors.** A missing `chat_id`/bad `arguments` makes `replyArgs`/`asToolCallParams` throw → caught as a protocol error (`-32603`), while forward/admin failures correctly become `isError` tool results. For a model to gracefully retry malformed args, argument-validation failures should also return `toolError(...)` (isError) rather than a protocol error.
3. **[minor] `command: 'dreamux'` is bare and relies on PATH injection.** Verified `dispatcherProcessEnv` prepends the package bin dir (and `dirname(DREAMUX_BIN)`), and the codex-spawned shim inherits that PATH, so it resolves today. An absolute bin path (e.g. via `DREAMUX_BIN`) would be more robust against PATH surprises in odd install layouts.
4. **[minor] `protocolVersion '2024-11-05'` is hardcoded** with no negotiation/fallback; if a codex version offers a different MCP protocol version, initialize could mismatch. Folds into the live-codex test above.

### Carry-forwards
Task 10 owns serve-side `mcp.reply`/`mcp.react` (honest NOT_IMPLEMENTED placeholders today). Task 11 still owns removing the legacy auto-outbound (Task 7's `sendOutbound`): until then, the model sees a `reply` tool that returns not-implemented **while assistant text is still auto-forwarded** — a harmless but slightly confusing transitional state worth keeping tracked.

Issue #51, opened 2026-06-03
Verify codex MCP-injection with a live test (GATE before Task 10) + feishu-mcp shim hardening


YourWildDad:
Follow-ups from the adversarial review of #50 (Task 9). The shim code itself passed (correct, secure, honest partial impl — serve side returns NOT_IMPLEMENTED rather than pretending to work). One near-term GATE plus three minors.

**1. [GATE — do before Task 10] The most load-bearing assumption has no automated live test.**
The shim is wired so that `resolveExtraArgs` injects the `mcp_servers.<name>` config into **every** dispatcher's codex child, unconditionally. The claim that a real codex (0.136) actually accepts this `mcp_servers` override, completes `initialize`, and that `mcpServerStatus/list` starts the shim and lists the `reply` tool was verified only by a one-off manual spike — there is **no automated live-codex test**. Two consequences:
- The whole outbound path (Task 10) is about to be built on top of this assumption; if it's wrong, it surfaces late (only at the cross-cutting e2e).
- Because injection is unconditional and global, a format error or a future codex change would **break startup for all dispatchers**, not just outbound — and nothing would catch it.
Recommendation: add a `codex-0136-mcp-live` test that launches a real codex and asserts `mcpServerStatus/list` returns the `reply` tool (follow the existing `codex-0135-live` opt-in-skip pattern). Do this before depending on the integration / before Task 10's outbound lands. Alternatively, gate the MCP injection behind a verified-codex check until that test exists.

**2. [minor] Tool-arg validation failures return a JSON-RPC protocol error (-32603) instead of an MCP tool error.**
The model can't gracefully retry on a protocol error; only forward/admin failures currently come back as tool errors. Route arg-validation failures through the same tool-error path.

**3. [minor] The shim spawns the binary as a bare command name, relying on PATH injection.**
PATH injection works today (the dispatcher process env adds the package bin), but resolving via an absolute binary path (e.g. an explicit env var) would be more robust.

**4. [minor] The MCP `protocolVersion` is hardcoded with no negotiation/fallback.**
Fold coverage of this into the live test from #1.

**Carry-forward:** Task 10 implements the serve-side `reply`/`react` (currently honest NOT_IMPLEMENTED placeholders); Task 11 removes the legacy auto-outbound from Task 7. Until Task 11, the model will see a `reply` tool that returns not-implemented while assistant text is still auto-sent — harmless transitional state, tracked.

YourWildDad:
[...]

PR #52, opened 2026-06-03
Add live gate for Feishu MCP injection

YourWildDad:
## Summary

This is the Task 9 gate follow-up for #36 before Task 10. It turns the manual Codex 0.136 MCP injection spike into an automated live compatibility test.

Changes:

- Add a live Codex app-server test that injects the Feishu stdio MCP shim and asserts `mcpServerStatus/list` exposes `reply` and `react`.
- Resolve the generated MCP command to an absolute `dreamux` launcher path, preferring `DREAMUX_BIN` when present.
- Return `tools/call` argument validation failures as MCP tool errors instead of JSON-RPC protocol errors.
- Negotiate supported MCP `protocolVersion` values and fall back safely.
- Update the top-level design record to document the absolute launcher path contract and the live compatibility gate.

## Codex MCP live result

Verified locally with `codex-cli 0.136.0`: a real `codex app-server` accepted the injected `mcp_servers.feishu` stdio config, completed initialize, started the shim, and returned the Feishu `reply` and `react` tools from `mcpServerStatus/list`.

## Validation

- `cd packages/dreamux && ./node_modules/.bin/vitest run tests/bin-launcher.test.ts tests/feishu-mcp.test.ts tests/codex-live.test.ts tests/smoke.test.ts`
- `.agents/scripts/check.sh`
- `git diff --check`
- staged diff red-line scan for public-repo leak patterns

Refs #36.

### DIFF
--- a/.agents/decisions/top-level-design.md
+++ b/.agents/decisions/top-level-design.md
@@ -307,11 +307,13 @@ Codex sends Feishu outbound actions only through a dispatcher-scoped MCP server.
 Model text is never automatically forwarded to Feishu.
 
 The default MCP transport between Codex and the Feishu MCP server is stdio.
-`dreamux` injects a per-dispatcher MCP server command into the Codex thread, for
-example:
+`dreamux` injects a per-dispatcher MCP server command into the Codex thread. The
+generated command path must be an absolute `dreamux` launcher path, resolved from
+the launcher-provided `DREAMUX_BIN` environment variable when available and from
+the package bin path otherwise. Schematically:
 
 ```text
-dreamux feishu-mcp --dispatcher dispatcher-a
+<dreamux-bin> feishu-mcp --dispatcher dispatcher-a
 ```
 
 The stdio process is a thin MCP shim. It is scoped to exactly one dispatcher,
@@ -467,6 +469,9 @@ resume path.
 - Multi-chat dispatcher access emits an operator-visible trust-domain warning.
 - Codex receives a dispatcher-scoped Feishu MCP stdio shim, not a loopback HTTP
   listener, on the default path.
+- The live Codex compatibility gate starts a real Codex app-server with the
+  injected stdio MCP config and asserts `mcpServerStatus/list` exposes the
+  Feishu `reply` and `react` tools.
 - The `feishu-mcp` shim forwards `reply` and `react` to `dreamux serve` through
   the local admin socket and never reads Feishu credentials.
 - If HTTP fallback is ever implemented, it refuses to start without a per-boot,
--- a/packages/dreamux/src/codex/mcp-config.ts
+++ b/packages/dreamux/src/codex/mcp-config.ts
@@ -1,15 +1,18 @@
+import { dreamuxBinPath } from '../runtime/package-bin.js';
+
 export const FEISHU_MCP_SERVER_NAME = 'feishu';
 
 export interface FeishuMcpCodexArgsOptions {
   dispatcherId: string;
   adminSocketPath: string;
   command?: string;
+  env?: NodeJS.ProcessEnv;
 }
 
 export function feishuMcpCodexArgs(
   opts: FeishuMcpCodexArgsOptions,
 ): string[] {
-  const command = opts.command ?? 'dreamux';
+  const command = opts.command ?? dreamuxBinPath(opts.env);
   return [
     '-c',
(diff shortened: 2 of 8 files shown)


YourWildDad (commented):
## Task 9 gate review (#52) — PASS (posted as comment; cannot self-approve, owner to approve/merge)

This genuinely discharges the PR #50 should-fix (#51). The manual Codex 0.136 spike is now a real, CI-enforced live test that cannot pass without verifying the load-bearing assumption. No blocking issues; a few minor robustness notes.

### Gate is real — verified end to end
- **Not a fake, not a silent skip.** The new `it` lives inside the existing `describe('codex live integration')` whose gating is: `DREAMUX_SKIP_LIVE_CODEX=1` → loud `console.warn` + `it.skip` (opt-in only); codex missing and not opted out → a test that **throws** (red CI); codex present → runs for real. Since CI installs `@openai/codex@latest` (install-model decision), the MCP gate runs for real in CI. This matches the repo's fail-loud-unless-opt-in pattern.
- **Strong assertion against real codex.** It spawns a real `CodexProcess` with `feishuMcpCodexArgs`, completes the real `initialize` handshake, calls real `mcpServerStatus/list`, and asserts the `feishu` server exposes **both** `reply` and `react` by name. Because an MCP stdio server has no static manifest, those tools appearing proves codex accepted the injected config, spawned the shim at the absolute launcher path, and the shim's stdio handshake interoperated with real codex — exactly the chain the spike claimed. Fails loud on codex `< 0.136` and on a missing/relative dreamux bin.
- **No false-confidence vector found.** A crashed/not-started shim yields no `tools` → the assertion fails (never a false pass). Worst case is a flaky-fail, not a false pass. `finally { proc.reap() }` SIGKILLs the process group, so neither codex nor the spawned shim leaks after the test.

### The 3 minors — all correctly fixed and tested
1. **Arg-validation → MCP tool error.** `callTool` is now wrapped so `replyArgs`/`reactArgs`/`asToolCallParams` failures return `toolError(...)` (`isError`) instead of JSON-RPC `-32603`. Tested (`{chat_id only}` → `isError` with "text must be a non-empty string", and asserts no `error` property). `tools/call` now never emits a protocol error.
2. **Absolute launcher path.** `dreamuxBinPath()` returns `resolve(DREAMUX_BIN)` or the absolute `<package>/bin/dreamux`; `feishuMcpCodexArgs` uses it. `packageBinDirs` was refactored to reuse it (the re-pushed dir dedupes via the existing `seen` set — no behavior change). The path is runtime-derived, so no username-bearing path is committed; the smoke test compares against the runtime value and only asserts it ends in `/dreamux`.
3. **protocolVersion negotiation.** `negotiateProtocolVersion` echoes a client-requested version when in `{2025-06-18, 2025-03-26, 2024-11-05}`, else falls back to `2024-11-05`. Unit-tested (echo + fallback), and the live test exercises real codex 0.136's actual requested version end to end.

### Red line — clean
Placeholders only (`dispatcher-a`/`chat-a`/`/tmp/not-used.sock`); the absolute bin path is generated at runtime on the operator/CI machine, never committed. No internal IDs/domains/registry/hostnames or committed username paths.

### Non-blocking findings (track if desired)
1. **No wait-for-started poll in the live MCP test.** It issues a single `mcpServerStatus/list` immediately after `initialize`. If a future codex starts MCP servers more lazily/asynchronously, this could become intermittently flaky (tools not yet populated). It can only flaky-*fail*, never false-pass, but a short poll-until-`feishu`-tools-present within the existing 30s budget would harden it.
2. **The test asserts `tools` but ignores any server `status` field.** Fine today (a failed shim has no tools, so it fails correctly), but asserting a started/running status would give a clearer failure message when codex lists the server but the shim crashed.
3. **`SUPPORTED_MCP_PROTOCOL_VERSIONS` is a hardcoded set** that will need maintenance as MCP versions advance; the live gate is the backstop (a newer codex requesting an unlisted version falls back to `2024-11-05`, and the gate would catch a hard break). Just flagging the maintenance coupling.
4. **(trivial)** `versionAtLeast` truncates pre-release suffixes (`0.136.0-rc1` parses as `0.136.0`), so a pre-release of the minimum version passes the gate — acceptable. Separately, CLAUDE.md still references `tests/codex-0135-live.test.ts` while the file is `tests/codex-live.test.ts` (pre-existing doc drift, not introduced here) — worth a one-line fix in a later docs pass.

PR #54, opened 2026-06-03
Implement serve-owned Feishu MCP outbound

YourWildDad:
## Summary

Implements #36 Task 10: serve-owned outbound reply/react handling behind the Feishu MCP shim.

Changes:

- Replace the Task 9 `mcp.reply` / `mcp.react` admin placeholders with live serve-side handlers.
- Keep Feishu credentials and SDK calls in `dreamux serve`: the MCP shim still only forwards to the admin socket.
- Send MCP `reply` through the dispatcher-owned Feishu bot using `chat_id`, optional `message_id`, and optional mentions.
- Add MCP `react` as a model-owned reaction path that does not touch channel-owned received reactions.
- Expose `removeReaction` through the dreamux Feishu bot adapter and fake bot.
- Clear only current-process channel-owned received reactions after successful MCP `reply`.
- Cover the interleaving where `reply` wins the race before the async received `addReaction` has stored its `reaction_id`; the server records a pending clear and removes the reaction after the id arrives.

## Validation

- `cd packages/dreamux && ./node_modules/.bin/vitest run tests/smoke.test.ts tests/feishu-bot.test.ts tests/feishu-mcp.test.ts`
- `.agents/scripts/check.sh`
- `git diff --check`
- staged diff red-line scan for public-repo leak patterns

Refs #36.

### DIFF
--- a/packages/dreamux/src/admin/methods.ts
+++ b/packages/dreamux/src/admin/methods.ts
@@ -112,28 +112,42 @@ export const adminMethods: Record<string, AdminHandler> = {
     return { dispatcher_id: id, status: 'stopped' };
   },
 
-  'mcp.reply': (server, params) => {
+  'mcp.reply': async (server, params) => {
     const id = mustDispatcherId(params);
     mustExistingDispatcher(server, id);
-    mustString(params, 'chat_id');
-    mustString(params, 'text');
-    optionalString(params, 'message_id');
-    optionalStringArray(params, 'mention_user_ids');
-    throw new AdminError(
-      'NOT_IMPLEMENTED',
-      'mcp.reply IPC is reserved for Task 10 serve-owned outbound handling',
-    );
+    mustRunningDispatcher(server, id);
+    const chatId = mustString(params, 'chat_id');
+    const text = mustString(params, 'text');
+    const messageId = optionalString(params, 'message_id');
+    const mentionUserIds = optionalStringArray(params, 'mention_user_ids');
+    try {
+      return await server.replyFromMcp({
+        dispatcherId: id,
+        chatId,
+        text,
+        ...(messageId !== null ? { messageId } : {}),
+        ...(mentionUserIds !== null ? { mentionUserIds } : {}),
+      });
+    } catch (err) {
+      throw new AdminError('OUTBOUND_FAILED', parseMessage(err));
+    }
   },
 
-  'mcp.react': (server, params) => {
+  'mcp.react': async (server, params) => {
     const id = mustDispatcherId(params);
     mustExistingDispatcher(server, id);
-    mustString(params, 'message_id');
-    mustString(params, 'emoji');
-    throw new AdminError(
-      'NOT_IMPLEMENTED',
-      'mcp.react IPC is reserved for Task 10 serve-owned outbound handling',
-    );
+    mustRunningDispatcher(server, id);
+    const messageId = mustString(params, 'message_id');
+    const emoji = mustString(params, 'emoji');
+    try {
+      return await server.reactFromMcp({
+        dispatcherId: id,
+        messageId,
+        emoji,
+      });
+    } catch (err) {
+      throw new AdminError('REACTION_FAILED', parseMessage(err));
+    }
   },
 };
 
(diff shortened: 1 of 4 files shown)


YourWildDad (commented):
## Task 10 review — PASS for scope (posted as comment; cannot self-approve, owner to approve/merge)

Task 10's requirements are fully implemented and correct, with a well-built race fix. No scope was cut. But merging this leaves the **integration branch double-posting** until Task 11, which I'm flagging as the top, high-priority finding for the integrator to gate on.

### Completeness — all Task 10 requirements implemented
- `mcp.reply`/`mcp.react` placeholders replaced with live `server.replyFromMcp`/`reactFromMcp`. reply uses `chat_id` + optional `message_id` + optional `mention_user_ids`; react adds a model-owned reaction. Failures wrap as `AdminError('OUTBOUND_FAILED'/'REACTION_FAILED')`.
- **Feishu SDK stays in serve; shim is pure forward.** `replyFromMcp`/`reactFromMcp` call the dispatcher-owned `slot.bot` (serve process). The shim is untouched by this PR — no Feishu credentials reach it, no direct-connect bypass. `bot.removeReaction` delegates to the transport core (the sole SDK importer).
- **Reaction ownership separation is correct.** `reply` success clears only the channel-owned received reaction stored in `receivedReactions` for the replied `message_id`; `react` stores nothing in that map, so a later clear never touches model-owned reactions. Verified the `react`-doesn't-clear test and that clearing a message that also has a model reaction removes only the channel-owned one.
- **Race fix is sound.** When `reply` wins before async `addReaction` stored its id, `clearReceivedReaction` records `pendingReceivedReactionClears`; when `addReaction` later resolves it sees the pending flag and clears. Traced both orderings: exactly one `removeReaction`, no double-clear, no dangling in the covered cases; `addReaction` failure/empty-id paths delete the pending entry so nothing is stranded. `pendingReceivedReactionClears` is capped at 1024 (oldest-evicted). No cross-dispatcher/cross-process contamination — all state is per-slot `channelState` in serve; the shim is stateless.

### Red line — clean
Placeholders only (`chat-group-a`/`msg-mcp-reply`/`THUMBSUP`/`sender-test`). No internal IDs/domains/registry/hostnames/username paths.

### Findings
1. **[HIGH — integration regression; recommend gating the branch] Double-outbound: the now-functional `reply` tool coexists with Task 7's still-active auto-outbound.** Confirmed `turn-manager.sendOutbound(batch, assistantText)` still auto-forwards every turn's assistant text (untouched here), and Task 10 makes `reply` actually send. So a turn where the model calls `reply` now produces **two Feishu messages** — the reply + the auto-forwarded assistant text. This is correct for Task 10's isolated scope (auto-outbound removal is Task 11), and I previously flagged this coexistence as "harmless transitional" in the #50 review — but it is no longer harmless now that `reply` sends for real. Recommend landing Task 11 (remove auto-outbound) immediately after this, or disabling auto-outbound as part of this merge, so the integration branch is never left in a double-posting state. Not a Task-10 code defect; a sequencing hazard the integrator should decide on before relying on the branch end to end.
2. **[should-fix — boundedness] `receivedReactions` map is unbounded.** It is `set` on every successful received-reaction add (line 421) and `delete`d only when a `reply`'s `message_id` matches (line 455). Two reasons entries accumulate forever: (a) while auto-outbound is live (finding #1), a model may never call `reply`; (b) by design a coalesced-batch reply targets only the newest message, so the non-newest messages' received reactions are never cleared (they dangle in Feishu and stay in the map). So the map grows ~1 entry per inbound message that never gets a matching reply — a slow unbounded leak on a long-lived dispatcher. The `pendingReceivedReactionClears` set is capped (1024) but `receivedReactions` is not. Recommend a cap/TTL on the map (and decide whether a batch reply should clear all batched `message_id`s, not just the newest).
3. **[minor test gap] No serve-side failure-path test.** The spec calls for a failure-return test. The shim's admin-error→tool-error conversion is covered (Task 9) and the handlers wrap failures in `AdminError`, but nothing exercises `setSendError`/`setReactionError` → `mcp.reply`/`mcp.react` → `OUTBOUND_FAILED`/`REACTION_FAILED` end to end. Add one to lock the serve-failure→tool-error chain.
4. **[minor] A `reply` whose `message_id` has no stored received reaction adds a never-consumed `pendingReceivedReactionClears` entry** (bounded at 1024, so in an extreme burst a real pending clear could be evicted, leaving a dangling received reaction). Narrow; the cap protects memory.

### Verdict
PASS on Task 10's scope — complete, correct, SDK-in-serve boundary held, race fix solid. The double-outbound (#1) is the thing to decide before this branch is exercised end to end: I'd treat "Task 11 lands immediately after / auto-outbound gated off" as the merge condition, and track #2 (map boundedness) as a follow-up consistent with the queue-boundedness item from Task 7.

PR #56, opened 2026-06-03
Remove legacy dispatcher runtime behavior

YourWildDad:
## Summary

Implements Task 11 from #36.

- Replace the SQLite-backed dispatcher repository with a config-backed dispatcher store plus per-dispatcher `status.json` runtime snapshots.
- Move dispatcher declarations and Feishu credentials to the top-level `dispatchers[]` config shape; admin add/remove now return an explicit unsupported error because config is the declaration source.
- Keep access-gate allowlists out of `config.json`: `dispatchers[].access` is rejected, and runtime access policy is read directly from per-dispatcher `state/<id>/access.json`.
- Accept per-dispatcher `codex.extra_env` and merge it into the Codex child environment while still removing `CODEX_HOME`.
- Remove automatic assistant-text outbound delivery and the legacy channel outbound sink. Feishu output is now MCP reply/react only.
- Remove the SQLite dependency, migration, and db source files.
- Update onboard, uninstall, doctor, CLI help, README, and decision docs to match the current config-backed MVP contract and dispatcher naming.

## Tests

- `pnpm --dir packages/dreamux build`
- `pnpm --dir packages/dreamux test -- global-config feishu-gate onboard smoke`
- `.agents/scripts/check.sh`
- `git diff --check`
- `git diff --cached --check`
- `gitleaks protect --staged --redact --config .gitleaks.toml`
- staged public-repo red-line scan for internal IDs, registry URLs, hostnames, and local absolute paths


YourWildDad (commented):
## Task 11 review — CHANGES (1 blocking finding; posted as comment, cannot self-approve)

The core Task 11 work is excellent and complete — SQLite removal, auto-outbound removal, `dispatchers[]` migration, and naming cleanup are all done and well-tested. But the implemented config schema contradicts the authoritative design's Operator Config in a way that makes the documented config un-loadable and silently drops `access`/`extra_env`, which I'm flagging as blocking-to-resolve before merge.

### Verified clean
- **Double-outbound eliminated (headline acceptance).** `turn-manager.sendOutbound` is deleted, `channel/outbound.ts` is deleted, the runtime `outbound` dep and approval-rejection Feishu send are gone. The only Feishu outbound path is now `server.replyFromMcp`/`reactFromMcp` (MCP). The `happy path: … assistant text is not auto-sent` smoke test is a real assertion — the fake Codex genuinely echoes assistant text, yet `bot.sentMessages` is `[]` (only the gate's received reaction is added). A model that never calls `reply` is now fully silent, matching no-auto-outbound. Turn failures are now silent-to-user too (logged only) — a deliberate consequence; worth noting it leaves a dangling "received" reaction (the documented limitation).
- **SQLite fully removed.** `db/{repository,schema,types}.ts` + `db/migrations/0001_init.sql` deleted, `better-sqlite3` dropped from `package.json`, and a repo grep finds no residual `openDatabase`/`DispatcherRepo`/`InboundRepo`/`inbound_buffer`/`databasePath`/`db/` imports. `admin dispatcher.add/remove` now return `UNSUPPORTED` pointing to config.json. `dispatcher.status` drops the `inbound_buffer` counts.
- **thread_id persistence migrated correctly.** It moved from the DB to per-dispatcher `state/<id>/status.json` (0600): `setThreadId`/`recordLostThread` persist; `readStatusFile` reloads on startup into `row.thread_id`. Auto-restart resume is intact (Task 8 uses in-memory `this.threadId`; server/manual restart resumes via the status.json-loaded `row.thread_id`) — no regression to fresh-thread.
- **app_id dedup + dispatcher-id validation migrated to array form.** `readDispatchers` rejects both a duplicate `id` and a duplicate `feishu.app_id`, and runs `validateDispatcherId` (charset) per element. `DispatcherStore.create` also guards duplicate app_id. `redactFeishuSecrets` is now recursive (redacts `app_secret` anywhere) and `resolveBotSecret` reads `config.dispatchers[].feishu.app_secret`.
- **Naming clean.** No committed `codexmux` residuals remain.
- **Red line clean.** Placeholders only; status.json is 0600; no internal IDs/domains/registry/username paths.

### BLOCKING — resolve before merge
1. **The `dispatchers[]` schema contradicts top-level-design's Operator Config (and silently cuts scope).** The implemented `DispatcherConfig` allows only `{id, cwd, enabled, feishu, codex:{approval_policy, sandbox_mode, extra_args}}`, and `rejectUnknownKeys` actively throws on anything else. But top-level-design's Operator Config example (unchanged by this PR) still documents **`access: { allow_group_chats, allow_users }`** and **`codex.extra_env`** (plus a Rules line: "`codex.extra_env` is merged over the server process environment…"). Consequences:
   - An operator who copies the design's own example into `config.json` gets `dreamux serve` failing with "`dispatchers[0].access is not supported`" / "`…codex.extra_env is not supported`". The authoritative doc's config is un-loadable.
   - Dropping `access` **hardens the Task 5 gap**: the access gate reads `state/<id>/access.json`, whose default denies all DMs and mention-gates groups. With `access` rejected from config and no admin command, DM allowlists / chat allowlists / follow-users are now configurable **only by hand-editing server-owned `state/<id>/access.json`** — there is no operator-facing config surface for the gate at all. Group@mention works out of the box; DMs and restrictive allowlists do not.
   Resolve one way or the other before merge: **(a)** implement `access` (bridge `config.dispatchers[].access` → `access.json` so the gate is operator-configurable, closing the Task 5 gap) and `codex.extra_env`; or **(b)** consciously drop them and update top-level-design's Operator Config example + Rules to match, with a follow-up issue owning where access management lives (an admin command or onboard step). Don't ship the doc and code contradicting.

### Non-blocking notes
- **Legacy config now hard-fails.** Top-level `runtime_dir`/`admin_socket`/`outbound`/`feishu.bots` are rejected by `rejectUnknownKeys`, so an existing pre-Task-11 config.json fails to load. Intentional (no-migration stance) and consistent, but a breaking change for any current install — make sure the upgrade note says "rewrite config.json to the dispatchers[] shape."
- **`created_at` derives from `status.json.updated_at`**, so `DispatcherStore.list()`'s `created_at` sort reorders dispatchers by last-update time rather than declaration order. Cosmetic (affects iteration/startup order, not correctness).
- **Carry-forward (from Task 10 review):** `receivedReactions` map is still unbounded (pruned only on a matching MCP reply; coalesced non-newest messages and unreplied turns accumulate). Not in this PR's scope; still worth a follow-up.

YourWildDad (commented):
## Task 11 re-review (head ddc6de0) — PASS (comment; cannot self-approve)

The prior blocking finding is fully resolved via the chosen direction (access lives in `access.json` only; `extra_env` implemented). I re-pulled the latest diff and verified all four required points plus the adversarial axes. No blocking issues; two minor test-strength suggestions.

### Requirement coverage — all met
1. **`access` is config-only-rejected; `access.json` is standalone; no bridge.** Dispatcher-level `rejectUnknownKeys` allows exactly `{id, cwd, enabled, feishu, codex}` — `dispatchers[].access` throws "…access is not supported by the MVP config schema" (tested). There is no config→access.json seeding: config rejects `access`, `onboard/config-files.ts` writes no `access` (only `codex.extra_env`), and the runtime reads `access.json` directly.
2. **Access capability is intact and really gates (not a no-op).** New smoke test `reads access gate configuration from access.json and allows configured DMs`: it writes `access.json` with `dm.allow_users:['sender-dm']`, injects a p2p DM from that sender, and asserts `fake.turnsHandled === 1` (the DM was *delivered* — which the empty default would have *dropped*), plus the received reaction. The Task 5 deny paths (bot-loop / bot-sender / unmentioned-group → `turnsHandled===0`, no send, no reaction) are retained. So the gate reads and applies `access.json` both ways.
3. **The blocking is resolved by removing `dispatchers[].access` from the docs, and everything is self-consistent.** top-level-design's Operator Config example dropped `allow_group_chats`/`allow_users` (kept `codex.extra_env`) and points to "the per-dispatcher `access.json` file described below"; the State-and-Logs section documents the `access.json` shape (`dm.allow_users`, `group.{allow_chats,follow_users,require_mention}`, …) with field names matching the code and the test. Root README, package README, onboard-generated config, tests, and the PR body all agree (config = declarations + `codex.*`; allowlists = `access.json`). An operator copying the design example now loads cleanly.
4. **`codex.extra_env` implemented and retained.** Schema accepts it (codex allowed set `{approval_policy, sandbox_mode, extra_args, extra_env}`) and validates it as a string→string record (`extra_env: {EXAMPLE_FLAG: 1}` throws "…must be a string", tested). It is passed through `runtime.deps.extraEnv` → `dispatcherProcessEnv(process.env, extraEnv)` → `CodexProcess.env`, and is documented as "merged over the server process environment."

### Adversarial — the operator-controlled `extra_env` cannot break load-bearing invariants
`dispatcherProcessEnv` merges in the right order: `{...baseEnv, ...extraEnv}`, then **PATH is recomputed** as `prependPath(packageBinDirs(merged), merged.PATH)` (so even an `extra_env.PATH` cannot shadow the package bin → `dreamux`/`tm`/the MCP shim still resolve), then **`CODEX_HOME` is deleted last** (so `extra_env` cannot re-introduce a dispatcher-private Codex home). The merge test confirms `DREAMUX_EXAMPLE_FLAG` is present, `PATH` contains the injected `/custom/bin`, and `CODEX_HOME` is undefined.

### Other verifications
- SQLite removal still clean (no `better-sqlite3`/`DispatcherRepo`/`databasePath`/`db/` residual); thread_id persists via `status.json` and auto-restart resume is intact.
- Double-outbound stays eliminated (the `assistant text is not auto-sent` test still asserts no send despite the fake echoing text); MCP reply/react is the only outbound.
- **Red line clean** — a scan of the new diff finds no internal hostnames, registry URLs, `ou_`/`oc_`/`cli_` IDs, or username-bearing absolute paths.

### Non-blocking (minor test-strength; code is already correct)
1. The `extra_env` merge test asserts `PATH` *contains* `/custom/bin` but does not assert the package bin dir is still prepended, so a hypothetical regression that let `extra_env.PATH` clobber the package-bin prepend would still pass it. Recommend also asserting `PATH` contains (or starts with) the package bin dir.
2. No test sets `CODEX_HOME` inside `extra_env` to prove the "delete after merge" defense. Recommend adding `CODEX_HOME: '/evil'` to the test's `extra_env` and asserting it stays undefined.
3. (carry from Task 10) `receivedReactions` map is still unbounded — out of scope here; track as a follow-up.

### Verdict
PASS. The earlier CHANGES blocking is resolved; Task 11's four required points are implemented, tested with real assertions, documented self-consistently, and the operator-controlled `extra_env` is merged safely. Mergeable from my side.

PR #57, opened 2026-06-04
Add Task 12 e2e and live Codex coverage

YourWildDad:
## Summary

Implements Task 12 from #36.

- Add cross-module e2e coverage for fake Feishu inbound reaching Codex, then replying through the Feishu MCP stdio shim, admin socket, and serve-owned outbound path.
- Cover the accepted restart limitation: received-reaction cleanup is process-local, so a server restart does not remove reactions added by a previous process.
- Keep live Codex coverage on the established compatibility gate: real Codex app-server startup plus injected Feishu MCP stdio config exposing `reply` and `react` via `mcpServerStatus/list`.
- Add an e2e status redaction assertion so the cross-module path does not leak Feishu secrets.

## Tests

- `pnpm --dir packages/dreamux exec vitest run tests/e2e.test.ts tests/codex-live.test.ts`
- `pnpm --dir packages/dreamux build`
- `pnpm --dir packages/dreamux test`
- `.agents/scripts/check.sh`
- `git diff --check`
- `git diff --cached --check`
- `gitleaks protect --staged --redact --config .gitleaks.toml`
- staged public-repo red-line scan for internal IDs, registry URLs, hostnames, and local absolute paths

Notes:

- The live Codex test ran against `codex-cli 0.136.0` without `DREAMUX_SKIP_LIVE_CODEX=1`.
- Rush still warns that it cannot hash the untracked `.claude/` worktree for repo-state caching, and Rush reports test stderr as warnings; both commands completed successfully.

### DIFF
--- a/packages/dreamux/tests/e2e.test.ts
+++ b/packages/dreamux/tests/e2e.test.ts
@@ -1,0 +1,368 @@
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+import {
+  mkdirSync,
+  mkdtempSync,
+  rmSync,
+  writeFileSync,
+} from 'node:fs';
+import { tmpdir } from 'node:os';
+import { dirname, join } from 'node:path';
+import { PassThrough } from 'node:stream';
+
+import { RECEIVED_REACTION_EMOJI, Server } from '../src/server.js';
+import { sendAdminRequest } from '../src/admin/client.js';
+import { loadDispatcherAccess } from '../src/channel/feishu-gate.js';
+import { CodexWsClient } from '../src/codex/rpc.js';
+import {
+  CodexProcess,
+  type CodexProcessOptions,
+} from '../src/codex/supervisor.js';
+import {
+  createFakeFeishuBot,
+  type FakeFeishuBot,
+  type FeishuInboundEvent,
+} from '../src/feishu/bot.js';
+import { runFeishuMcp } from '../src/mcp/feishu-mcp.js';
+import { BUILT_IN_DEFAULTS, type DreamuxConfig } from '../src/runtime/config.js';
+import {
+  dispatcherCodexCwd,
+  dispatcherCodexHome,
+  dispatcherWorkspaceSkillPath,
+} from '../src/runtime/paths.js';
+import { startFakeCodex, type FakeCodex } from './fake-codex.js';
+
+class NoopCodexProcess extends CodexProcess {
+  constructor(opts: CodexProcessOptions) {
+    super(opts);
+  }
+
+  override async start(): Promise<void> {
+    // No child process; e2e tests connect the runtime to fake Codex.
+  }
+
+  override async reap(): Promise<void> {
+    // Nothing to reap.
+  }
+}
+
+class JsonLineReader {
+  private buffer = '';
+  private readonly waiters: Array<(value: unknown) => void> = [];
+
+  constructor(stream: PassThrough) {
+    stream.setEncoding('utf8');
+    stream.on('data', (chunk: string) => {
+      this.buffer += chunk;
+      this.drain();
+    });
+  }
+
+  next(): Promise<unknown> {
+    const line = this.shiftLine();
+    if (line !== null) return Promise.resolve(JSON.parse(line));
+    return new Promise((resolve) => {
+      this.waiters.push(resolve);
+    });
+  }
+
+  private drain(): void {
+    while (this.waiters.length > 0) {
+      const line = this.shiftLine();
+      if (line === null) return;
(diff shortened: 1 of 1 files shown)


YourWildDad (commented):
## Task 12 review — PASS (comment; cannot self-approve, owner to approve/merge)

The final epic task adds one file, `tests/e2e.test.ts` (368 lines, no source changes), and relies on the established live-Codex gate. I verified all four required points and the adversarial axes. No blocking issues; minor notes only.

### Requirement coverage — all met
1. **The e2e drives the real production middle — not architected away.** Only the two I/O endpoints are faked (fake Codex WS via `startFakeCodex`, fake Feishu bot, plus `NoopCodexProcess` to skip the real child). Everything between is real: the gate, the turn manager, `formatFeishuMessageForCodex`, **the real `runFeishuMcp` shim driven over real PassThrough stdio**, the **real admin Unix socket** round-trip, the real `server.replyFromMcp`, and the reaction cleanup. Test 1 asserts behavior, not fake-swallowing: the Codex prompt contains the real `<feishu_message`/`sender_name="Ada"` block; the reply response carries `structuredContent.message_ids`; `bot.sentMessages` is **exactly one** message with the correct `chat_id`/`replyToMessageId`/`mentionUserIds`/`text` (so if auto-outbound had survived, the echoed assistant text would make this fail — a real no-double-send check); `bot.removedReactions` confirms the received-reaction clear. The fake `replyFor` returns "assistant text must not be sent automatically" precisely so that any auto-send would be caught.
2. **Restart limitation + status redaction covered.** Test 2 injects, gets a received reaction, restarts the Server (new `channelState`, empty `receivedReactions`), then replies for the same message: the reply *succeeds* (`message_ids`, one `sentMessage`) but `removedReactions === []` and the original reaction remains — i.e. process-local cleanup does not survive restart (the documented limitation), asserted for the right reason (the reply ran, it just couldn't clear). Test 3 calls `server.status` and asserts the payload contains `flow` but not `secret-server-only`.
3. **Live-Codex gate stays fail-loud; no silent-skip introduced.** This PR does not touch `codex-live.test.ts`; I confirmed on this branch it is still the #52 gate — `DREAMUX_SKIP_LIVE_CODEX=1` → loud `it.skip`; codex missing and not opted out → a test that `throw`s (red CI); otherwise it spawns a real Codex app-server and asserts `mcpServerStatus/list` exposes both `reply` and `react`. CI installs `@openai/codex@latest`, so it runs for real. `e2e.test.ts` itself is all fake-Codex and deterministic — it adds no live test that could silently skip.
4. **No unit-test duplication.** `e2e.test.ts` only composes cross-module behavior; it does not re-test the units already pushed down in Tasks 1-11. It is strictly additive over `smoke.test.ts`'s mcp.reply test — that one called `sendAdminRequest` directly; this one routes through the real stdio shim, adding the codex-facing link smoke didn't exercise.

### Adversarial — coverage boundary, acknowledged not faked
The one link no single test drives deterministically is "a real Codex model decides to call `reply` and Codex transports that `tools/call` to the shim's stdin" — inherently nondeterministic with a real model. It is covered by composition: the live gate proves Codex↔shim stdio for `initialize` + `tools/list` (Codex must call the shim over stdio for `mcpServerStatus/list` to return its tools), and this e2e proves shim `tools/call` → admin.sock → serve → outbound. Since `tools/call` uses the identical stdio framing the live gate already exercises, the residual risk is negligible. Honest tradeoff, not a gap.

### Red line — clean
Placeholders only (`app-e2e`/`secret-server-only`/`chat-group-a`/`Ada`/`fake-open-id-app-e2e`) and `tmpdir()` paths. No internal hostnames/registry/IDs/username-bearing absolute paths.

### Non-blocking notes
1. **Test scaffolding duplication.** `JsonLineReader`, `NoopCodexProcess`, `fakeInbound`, `writeJson`, `waitFor`, `buildServer` are copied from `smoke.test.ts`/`feishu-mcp.test.ts`. Not case duplication (req #4 holds), but worth extracting into a shared test helper to avoid drift.
2. `callFeishuMcpTool` spawns a fresh shim per call (initialize → one tools/call → exit), whereas real Codex keeps one shim alive across calls. Fine for the forward-path assertion (the multi-call loop is covered by the shim's own unit tests), just noting the modeling difference.
3. (carry from Task 10) the restart test exercises the `pendingReceivedReactionClears` path that adds an entry which never resolves; `receivedReactions` remaining unbounded is still an open follow-up, out of scope here.

### Verdict
PASS — the e2e exercises the real production chain with genuine assertions, the restart/known-limitation and status-redaction design points are covered, and the live-Codex gate remains fail-loud and CI-enforced. This closes the epic's test surface from my side.

PR #61, opened 2026-06-04
Fix systemd service Node environment for nvm installs

YourWildDad:
## Summary

- Capture a service-safe Node/Codex environment during `dreamux onboard`, including `DREAMUX_NODE_BIN`, absolute `CODEX_HOST_CODEX_BIN`, and a minimal PATH for managed services.
- Make the `dreamux` launcher honor `DREAMUX_NODE_BIN` before falling back to bare `node`.
- Validate the managed service launch environment before systemd registration, and teach `dreamux doctor` to inspect the installed systemd unit environment instead of reporting a false green from the interactive shell.
- Update the service-launch decision record and focused tests.

## Verification

- `./node_modules/.bin/tsc -p tsconfig.json --noEmit` from `packages/dreamux`
- `./node_modules/.bin/vitest run tests/onboard.test.ts tests/doctor.test.ts tests/bin-launcher.test.ts` from `packages/dreamux`
- `.agents/scripts/check.sh`
- `env -i DREAMUX_NODE_BIN="$(command -v node)" PATH=/usr/bin:/bin packages/dreamux/bin/dreamux --help`
- `shellcheck packages/dreamux/bin/dreamux bin/dreamux`
- `node common/scripts/install-run-rush.js change --verify --no-fetch`
- pre-commit `gitleaks protect --staged`

Notes:
- Rush prints its standard Node-version warning for Node 22.16.0.
- `rush test` exits successfully; the dreamux test target reports warnings because test logs intentionally write to stderr.

### DIFF
--- a/.agents/decisions/global-bin-onboard-serve.md
+++ b/.agents/decisions/global-bin-onboard-serve.md
@@ -46,6 +46,14 @@ channel configuration, and registers a native service manager entry.
 `dreamux serve` runs the existing server in the foreground and lets launchd or
 systemd keep it alive.
 
+The generated service environment must be self-sufficient. It must not rely on
+interactive shell startup files such as `.zshrc` to make Node or Codex
+available. During onboarding, dreamux captures the Node executable that is
+running onboarding, validates that it satisfies the package's supported Node
+range, resolves the Codex executable to a runnable path for managed service use,
+seeds `HOME` for clean user-service probes, and renders those values into the
+user service environment.
+
 There is no public `dreamux daemon ...` command tree. The daemon form is
 foreground `dreamux serve` supervised by a native user-level service manager.
 After `dreamux onboard` registers that service, operators use native
@@ -119,6 +127,10 @@ status.
 - The `dreamux` launcher passes its resolved absolute path through
   `DREAMUX_BIN`; service generation uses that path so launchd and
   systemd execute the same global bin the operator invoked.
+- The `dreamux` launcher honors `DREAMUX_NODE_BIN` when present, falling back to
+  `node` for ordinary interactive and npm-bin use. Service generation sets
+  `DREAMUX_NODE_BIN` and a minimal `PATH` so nvm-installed Node works without
+  sourcing shell rc files.
 - Codex's global default home is intentionally reused, so login state, memory,
   and local Codex configuration follow the operator's machine. dreamux must not
   delete that home during uninstall.
--- a/common/changes/@excitedjs/dreamux/codex-systemd-nvm-service-env_2026-06-04-07-19.json
+++ b/common/changes/@excitedjs/dreamux/codex-systemd-nvm-service-env_2026-06-04-07-19.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Fix managed service startup when Node is provided by nvm.",
+      "type": "patch",
+      "packageName": "@excitedjs/dreamux"
+    }
+  ],
+  "packageName": "@excitedjs/dreamux",
+  "email": "15624809+YourWildDad@users.noreply.github.com"
+}
--- a/packages/dreamux/bin/dreamux
+++ b/packages/dreamux/bin/dreamux
@@ -30,4 +30,5 @@ if [ ! -f "$TARGET" ]; then
   echo "Build it first from the repo root: node common/scripts/install-run-rush.js build (rush build)." >&2
   exit 1
 fi
-exec env DREAMUX_BIN="$DREAMUX_BIN" node "$TARGET" "$@"
+NODE_BIN="${DREAMUX_NODE_BIN:-node}"
+exec env DREAMUX_BIN="$DREAMUX_BIN" "$NODE_BIN" "$TARGET" "$@"
--- a/packages/dreamux/src/cli/doctor.ts
+++ b/packages/dreamux/src/cli/doctor.ts
@@ -1,5 +1,7 @@
-import { existsSync } from 'node:fs';
+import { existsSync, readFileSync } from 'node:fs';
 import { homedir } from 'node:os';
+
+import { parse as parsePlist, type PlistValue } from 'plist';
 
 import { codexArgsToCli, parseCodexArgs } from '../runtime/codex-args.js';
 import {
@@ -23,7 +25,8 @@ import {
 import { ExecaCommandRunner } from '../onboard/commands.js';
 import {
   LAUNCHD_LABEL,
-  managedServiceEnvironment,
+  MIN_SERVICE_NODE_VERSION,
+  nodeVersionSatisfies,
   serviceUnitPath,
   SYSTEMD_UNIT,
 } from '../onboard/service.js';
@@ -46,6 +49,8 @@ export interface ServiceStatus {
   enabled: boolean;
   pid: number | null;
   detail: string | null;
+  environment: Record<string, string> | null;
+  execStart: string[] | null;
 }
 
 export interface DoctorCheck {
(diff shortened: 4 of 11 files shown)


YourWildDad:
## Claude review — PR #61 (systemd + nvm service Node env)

**Reviewed head:** `c414637e80885e8ea1430ee5a924e50e141769a7` (matches the author's expected SHA).
**State:** OPEN, not draft, `MERGEABLE` / `mergeStateStatus: CLEAN`. All 8 required checks green on run `26937107889` (bootstrap/build/typecheck/test on ubuntu + macos, shellcheck ×2, gitleaks full-history, KB link check, rush change, commit-metadata). Title has no CI-skip token. No prior reviews or comments on the PR.

I verified the claims against the code rather than trusting the description.

### What holds up

- **Both Node layers of the failing chain are addressed.** The launcher now execs `${DREAMUX_NODE_BIN:-node}` (`packages/dreamux/bin/dreamux`), the generated env sets `DREAMUX_NODE_BIN` + a minimal `PATH` whose first entry is `dirname(process.execPath)`, and `CODEX_HOST_CODEX_BIN` is resolved to an absolute path at onboard time (`resolveServiceExecutable`). That covers (a) the dreamux launcher's `node` and (b) the Codex CLI's own `#!/usr/bin/env node` shim, which still resolves via the service `PATH`. Good — an absolute `ExecStart` alone would not have fixed (b).
- **`managedServiceEnvironment` stays the single builder** feeding systemd render, launchd render, and the onboard preflight, so the env can't drift between them.
- **Doctor reads the *installed* unit** (`parseSystemdUnit`) and probes `DREAMUX_NODE_BIN`/launcher/Codex from those values instead of recomputing from doctor's own `process.execPath` — the false-green/stale-unit case is genuinely closed on systemd, and the new test asserts it (`runner.calls` contains `/service/node --version`, excludes `process.execPath`).
- **`CODEX_HOME` is untouched**; no relocation or new home dir. Uninstall scoping is unchanged.
- **Node version is gated** (`nodeVersionSatisfies`, `>=22.7`) in both onboard validation and doctor.

### Findings

**P1 — macOS doctor always reports the managed service as failed once a launchd service is installed.**
`packages/dreamux/src/cli/doctor.ts:309` (`addManagedServiceLaunchChecks`) runs for every platform and only short-circuits on `!service.installed`. But `launchdStatus` hard-codes `environment: null` / `execStart: null` (`packages/dreamux/src/cli/doctor.ts:256`) — it never parses the plist's `EnvironmentVariables` / `ProgramArguments`. So on macOS, with the plist installed, `env === null` → it pushes a failing `managed service environment` check (`doctor.ts:313`,`:321`) and `runDreamuxDoctor` returns `ok: false` unconditionally, with the remediation "rerun dreamux onboard" — which cannot fix it, since onboard never makes `launchdStatus` parse the plist. launchd is a first-class supported platform (`serviceUnitPath` supports darwin + linux), so this is a regression in the exact diagnostic this PR is hardening. It is uncaught because every doctor test uses `platform: 'linux'`.
*Suggested fix:* either parse the launchd plist (`EnvironmentVariables` → `environment`, `ProgramArguments` → `execStart`) in `launchdStatus` so the launch checks have parity, or gate `addManagedServiceLaunchChecks` to the systemd platform until launchd parsing exists. Add a darwin doctor test with an installed plist.

**P2 — the launch validation is not a true clean room.** `ExecaCommandRunner.check/capture` (`packages/dreamux/src/onboard/commands.ts`) calls `execa(..., { env })`, and execa's default `extendEnv: true` merges the passed env *over* the ambient process env. The keystone (`PATH`) is overridden so the "Node resolvable without nvm shell init" guarantee does hold, but any *other* interactive-only var the operator has set still leaks into the probe and would be absent under systemd — so onboard/doctor can pass where the real unit would fail. If a strict clean-room check is intended, pass `extendEnv: false` (and seed `HOME`, which systemd user units do get but a clean env would not).

**P2 — systemd env unescape round-trip is asymmetric for a literal `\x20` substring.** `systemdEscapeEnv` (`service.ts:357`) escapes `\`→`\\` before encoding spaces, while `systemdUnescapeEnv` (`doctor.ts:425`) decodes `\x20`→space first; a value containing the literal text `\x20` (or a literal backslash adjacent to `x20`) does not round-trip. Not reachable with normal filesystem paths, so low priority, but it is a parser the doctor now relies on — worth a unit test or a tokenizing decode.

**P2 — coverage gaps that let the P1 through.** No test exercises the darwin/launchd doctor path with an installed service, and no test asserts the launchd plist carries `DREAMUX_NODE_BIN`/`PATH` (onboard test is linux-only).

### Axes covered
Framing (both launcher + Codex Node-wrapper chains), generated systemd unit + launchd plist env, onboard preflight validation, doctor false-green/stale-unit, `CODEX_HOME`/uninstall boundaries, test mocked-vs-real and clean-env fidelity, Linux/macOS parity, shell quoting / env escaping, Node version compatibility.

Recommend addressing the P1 before merge (macOS doctor regression); the P2s are non-blocking.

YourWildDad:
## Claude recheck — PR #61, round 1

**Reviewed head:** `08b89d64b0b8cb1b53d5a11c8dd03e6bfa95f498` (matches the author's reported SHA).
**State:** OPEN, not draft, `MERGEABLE` / `CLEAN`. All 8 checks green on run `26937870471` (build/typecheck/test on ubuntu + macos, shellcheck ×2, gitleaks full-history, KB link check, rush change, commit metadata). Title unchanged, no CI-skip token.

Verified the changes against the code (`git diff c414637..08b89d6`), not just the description.

### P1 — RESOLVED

`launchdStatus` now reads the installed plist via `parseLaunchdPlist` (`packages/dreamux/src/cli/doctor.ts`), mapping `EnvironmentVariables` → `environment` and `ProgramArguments` → `execStart`. With `managedServiceEnvironment` populating the plist's env (`PATH`, `DREAMUX_NODE_BIN`, `CODEX_HOST_CODEX_BIN`, plus the new `HOME`/`DREAMUX_CONFIG_DIR`), `addManagedServiceLaunchChecks` now sees a non-null env with the required keys present on macOS and runs the real Node/launcher/Codex probes instead of failing unconditionally. The stale-plist case (no `EnvironmentVariables`) correctly falls through to "rerun dreamux onboard", which now *is* the right remediation since re-onboard rewrites the plist. New test `checks the installed launchd plist environment instead of failing unconditionally` (`platform: 'darwin'`) covers it, and macOS CI is green.

### P2 follow-ups — all addressed, no regressions

- **Clean-room probe:** `execaEnvironment` sets `extendEnv: options.env === undefined` (`onboard/commands.ts`), so explicit-env probes no longer inherit ambient vars while normal `systemctl`/`launchctl` calls still do. Covered by `commands.test.ts` (`does not inherit ambient environment when explicit env is passed`). The probe is now stricter than the live systemd env (which also gets systemd-injected vars), i.e. fail-closed — acceptable for a preflight gate, and the `PATH`/`HOME` keystones are seeded.
- **`HOME` seed:** added to `managedServiceEnvironment` and documented in `.agents/decisions/global-bin-onboard-serve.md`. (`HOME` is a local home path written only into the local unit/plist — no public-repo leakage.)
- **`\x20` round-trip:** `systemdUnescapeEnv` rewritten as a single-pass tokenizer; I traced the previously-broken literal-`\x20` case and it now round-trips. Covered by a `commands`/`doctor` test asserting `/service/codex\x20literal`.

### Axes covered this round
launchd plist parse (env + ProgramArguments), macOS doctor no-longer-unconditional-fail, stale-plist remediation path, clean-room env inheritance direction, `HOME` seed and its hygiene, systemd env unescape round-trip, CI/mergeability, title skip-token.

**Verdict: no P0/P1 remaining and no new P0/P1 introduced. LGTM — mergeable.**

Issue #63, opened 2026-06-04
Dispatcher long-running turn head-of-line-blocks Feishu inbound — operator sees ✓ but Codex never receives


YourWildDad:
## Symptom

Operator sends Feishu messages to a dispatcher; the channel layer **acks each one with the “received” reaction (✓)**, but Codex never responds and the messages never appear in the Codex thread. Observed in a live session: a single dispatcher turn ran for 30+ minutes while every subsequent inbound got reacted-but-never-delivered.

This reads to the operator as “the channel broke / Codex stopped receiving”, but the transport, gate, MCP shim and Codex child are all healthy. It is **head-of-line blocking on the serialized turn worker**.

## Root cause

The inbound path is:

```
Feishu → gate → TurnManager.enqueue → (serialized) runTurn → Codex thread
```

Two facts combine into the trap:

1. **The “received” reaction is applied at enqueue time, not at delivery time.** In `packages/dreamux/src/server.ts`, the bot handler calls `runtime.enqueueInbound(...)` and, *iff it returns `true`*, immediately calls `addReceivedReaction(...)`. So the ✓ only means “accepted into the in-memory queue”, **not** “injected into Codex”.

2. **`TurnManager` is a single serialized worker.** In `packages/dreamux/src/dispatcher/turn-manager.ts`, `drainLoop` processes one batch at a time with `await processBatch` → `await runTurn(...)`. `runTurn` (`packages/dreamux/src/codex/events.ts`) only resolves on `turn/completed`. While the current turn is in flight, the queue is **not** drained — later messages just coalesce into a pending batch and wait.

So any inbound that legitimately kicks off a **long-running dispatcher turn** silently freezes the channel for that whole turn, while the operator keeps getting ✓ on every new message.

### How a turn stays open for tens of minutes

The dispatcher’s own coordination workflows hold the turn open by design. In the observed case the dispatcher ran a full author→review→fix cycle inside **one** turn, blocking synchronously on teammate replies (repeated `write_stdin` / `exec_command` polls of 60–300s each). The turn never reached `turn/completed`, so the serialized worker never advanced to the queued messages.

## Evidence (redacted)

- Codex thread: last `task_started` with **no following `task_complete`**; the trailing record is a multi-hundred-second blocking `write_stdin` poll. The turn had been open ~30+ min.
[...]

PR #64, opened 2026-06-04
Make queued inbound reactions honest

YourWildDad:
## Summary

- Move the channel-owned received reaction from Feishu enqueue time to after Codex accepts `turn/start` for a batch.
- Leave messages queued behind an in-flight dispatcher turn without the same received marker until their batch actually starts moving into Codex.
- Add stuck-turn smoke coverage and update the documented invariant.

Refs #63.

## Testing

- `./node_modules/.bin/vitest run tests/smoke.test.ts`
- `./node_modules/.bin/vitest run tests/turn-manager.test.ts`
- `./node_modules/.bin/tsc -p tsconfig.json --noEmit`
- `./node_modules/.bin/vitest run`
- `.agents/scripts/check.sh`

### DIFF
--- a/.agents/decisions/top-level-design.md
+++ b/.agents/decisions/top-level-design.md
@@ -293,8 +293,10 @@ When no group chat allowlist or follow-user allowlist is configured, group
 messages are still mention-gated. This keeps bootstrap usable while preventing
 ambient group chatter from entering the dispatcher.
 
-The gate adds a channel-owned received reaction only after a message is accepted
-and enqueued. If the message is rejected, no reaction is added.
+The channel adds a channel-owned received reaction only when that message's
+batch starts moving into Codex. If a message is accepted but still queued behind
+an in-flight dispatcher turn, it does not get the same received marker yet. If
+the message is rejected, no reaction is added.
 
 When one dispatcher is configured for, or observes, multiple allowed chats, the
 gate records an operator warning that those chats share one Codex context.
@@ -447,6 +449,9 @@ resume path.
   cross-chat queue are in-memory and not globally size-bounded in the MVP. A
   stuck turn plus continued inbound traffic can grow memory until operator
   intervention or process restart.
+- Messages queued behind an in-flight dispatcher turn are accepted in memory but
+  are not marked with the channel-owned received reaction until their batch
+  starts moving into Codex.
 - Codex version drift can still break protocol behavior. `dreamux` detects and
   reports this through `doctor`, live tests, and version diagnostics instead of
   pinning Codex.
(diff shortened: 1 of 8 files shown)


YourWildDad (commented):
**APPROVE** — heterogeneous review (Claude reviewing Codex-authored change).

Head SHA reviewed: `e18157191f91fbff8a430097b27d30581bce2bb0`

### What the change does
Moves the channel-owned "received" reaction from Feishu **enqueue time** to the point where Codex accepts `turn/start` for a batch. The hook chain is clean and minimal: `Server.onTurnBatchStart` → `DispatcherRuntime.onBatchStart` → `TurnManager.notifyBatchStart` → `runTurn`'s new `onStarted` callback. `onStarted` fires exactly once per batch, right after the `turn/start` RPC resolves.

### Correctness verification
- **Single `runTurn` caller, single `onStarted` source** — no other code paths affected; the new `opts` arg defaults to `{}`, so the signature change is backward-compatible.
- **No double-reaction risk** — a failed batch is logged and dropped, never retried, so `onStarted`/`notifyBatchStart` cannot fire twice for the same batch.
- **Add/clear race preserved** — the reaction add is now fire-and-forget (`void addReceivedReaction`), but the pre-existing `pendingReceivedReactionClears` mechanism still covers the case where a reply clears before the async add registers, and `addReceivedReaction` swallows its own errors so the `void` cannot leak an unhandled rejection.
- **Coalesced batches** — each message in a batch is reacted by its own `source_message_id`; null/empty ids are guarded.
- **Honest semantics confirmed by the new smoke test** — with `holdTurns` (acks `turn/start`, never emits `turn/completed`), the in-flight message gets ✓ while a message queued behind it gets **no** ✓ until its batch moves into Codex. This is exactly the misleading-✓ symptom from the issue.

### Scope note (important for any merging maintainer)
This PR fixes the **reaction-honesty** facet of #63 (direction 1) and adds the **stuck-turn test** (direction 4). It does **not** address head-of-line blocking (direction 2) or backpressure notification (direction 3): a long dispatcher turn still serializes the worker, so queued inbound is still not delivered to Codex until the turn ends — the operator simply no longer gets a false ✓. The commit message "Covers issue #63" slightly overstates this; the residual blocking is correctly documented as a known limitation in `top-level-design.md`. Worth keeping in mind so the operability half of #63 isn't assumed solved.

### Residual notes (non-blocking)
- On the unhappy path (`getThreadId()` null, or `turn/start` rejects), the batch is dropped (pre-existing) and now yields no reaction — a dropped message is silent rather than falsely-acked. This is the correct direction for honest semantics, just a conscious behavior change worth naming.
- The negative test relies on `sleep(120)` + `holdTurns`; flake risk is low because `holdTurns` makes the second batch undequeuable regardless of timing, so the assertion can't race.

### CI
All 9 checks green on the head SHA (rush build/typecheck/test on ubuntu + macos, shellcheck ubuntu + macos, gitleaks full-history, KB link/orphan, rush-change declaration, commit author-email).

YourWildDad:
Closing this version for redesign.

This approach improves the meaning of the received reaction, but it does not solve the core #63 problem: inbound messages can still remain stuck behind a long-running dispatcher turn, invisible to Codex and without an actionable backpressure/status path.

The next design should address the head-of-line blocking behavior itself, or provide an explicit non-blocking/backpressure mechanism where long teammate coordination does not keep the dispatcher turn open and queued inbound remains observable.

PR #65, opened 2026-06-04
Fix non-blocking Feishu inbound delivery

### LINKED ISSUE #63: Dispatcher long-running turn head-of-line-blocks Feishu inbound — operator sees ✓ but Codex never receives

YourWildDad:
## Summary
- Submit every accepted, deduped Feishu inbound via aggregate `turn/start` without waiting for `turn/completed`.
- Replace the old completion-gated queue/coalescing path with per-message delivery and process-local dedupe.
- Implement the channel-owned reaction tri-state: received -> in progress -> cleared on MCP reply, including reply-before-add reconciliation.
- Add fake Codex active-turn folding coverage and a real Codex app-server live gate for issue #63.
- Document the issue #63 runtime contract in `.agents/domains/non-blocking-dispatcher-inbound.md`.

## Verification
- `cd packages/dreamux && pnpm exec vitest run tests/turn-manager.test.ts tests/smoke.test.ts tests/e2e.test.ts`
- `cd packages/dreamux && pnpm exec vitest run tests/codex-live.test.ts` (real Codex 0.137.0; issue #63 mid-turn model gate passed)
- `cd packages/dreamux && CI=true pnpm exec vitest run tests/codex-live.test.ts` (CI-mode loud-skip for model-auth-only gate)
- `node common/scripts/install-run-rush.js test` (full suite; local run includes the real issue #63 model gate; Rush exits successfully with expected stderr logs)
- `.agents/scripts/check.sh`
- `gitleaks protect --staged --redact --verbose`

Fixes #63

### DIFF
--- a/packages/dreamux/src/dispatcher/turn-manager.ts
+++ b/packages/dreamux/src/dispatcher/turn-manager.ts
@@ -1,16 +1,14 @@
 /**
- * Per-dispatcher in-memory turn worker.
+ * Per-dispatcher in-memory inbound submitter.
  *
  * Contract:
- *   - one serialized worker per dispatcher;
  *   - accepted inbound messages are not persisted;
- *   - consecutive pending messages from the same chat are coalesced into one
- *     Codex turn;
  *   - Feishu message_id redelivery is deduped within this server process;
+ *   - one accepted inbound message becomes one Codex `turn/start` submission;
  *   - Codex text output alone does not send anything to Feishu.
  */
 
-import { runTurn } from '../codex/events.js';
+import { submitTurnStart } from '../codex/events.js';
 import type { CodexWsClient } from '../codex/rpc.js';
 
 export const DEFAULT_MESSAGE_ID_DEDUPE_WINDOW = 1024;
@@ -25,9 +23,18 @@ export interface InboundTurnInput extends InboundTurnSource {
   parsed_text: string;
 }
 
-interface TurnBatch extends InboundTurnSource {
-  id: number;
-  messages: InboundTurnInput[];
+export type InboundDeliveryResult =
+  | { status: 'duplicate' }
+  | { status: 'stopped' }
+  | { status: 'submitted'; turnId: string }
+  | { status: 'failed'; error: Error };
+
+export interface InboundDeliveryHooks {
+  /**
+   * Called after process-local dedupe accepts the message and before
+   * `turn/start` is submitted.
+   */
+  onAccepted?: (input: InboundTurnInput) => void | Promise<void>;
 }
 
 export interface TurnManagerOptions {
@@ -47,14 +54,9 @@ export interface TurnManagerOptions {
 }
 
 export class TurnManager {
-  private readonly queue: TurnBatch[] = [];
   private readonly seenMessageIds = new Set<string>();
   private readonly seenMessageIdOrder: string[] = [];
-  private running = false;
   private stopped = false;
-  private drainScheduled = false;
-  private wakeup: (() => void) | null = null;
-  private nextBatchId = 1;
   private readonly log: NonNullable<TurnManagerOptions['log']>;
   private readonly messageIdDedupeWindow: number;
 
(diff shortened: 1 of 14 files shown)


packages/dreamux/src/dispatcher/turn-manager.ts
```diff
@@ -71,103 +73,63 @@ export class TurnManager {
...
+  ): Promise<InboundDeliveryResult> {
+    if (this.stopped) return { status: 'stopped' };
+    if (!this.rememberMessageId(input.source_message_id)) {
+      return { status: 'duplicate' };
     }
 
-    this.notify();
-    return true;
-  }
+    await this.notifyAccepted(input, hooks);
```
YourWildDad, line 88: Non-blocking: `notifyAccepted` (which awaits the Feishu `[了解]` add) is awaited *before* `submitTurnStart`. This guarantees `[了解]` precedes `[在做了]`, but it also couples codex delivery to the Feishu reaction round-trip — a slow/hung `addReaction` now stalls the `turn/start` submission for that message. It does not reintroduce HOL (each inbound is independent), so it is acceptable for the MVP. If reaction-API flakiness ever shows up, consider a bounded timeout on the hook, or submit-first-then-reconcile-ordering. Just flagging the coupling.

packages/dreamux/src/server.ts
```diff
@@ -386,63 +419,81 @@ export class Server {
...
+  if (previous !== undefined) {
+    try {
+      await bot.removeReaction(messageId, previous.reactionId);
+    } catch (err) {
+      console.error(
+        `[server] failed to replace the ${previous.state} reaction for dispatcher '${dispatcherId}' message '${messageId}':`,
+        err,
+      );
+    }
+    channelState.inboundReactions.delete(messageId);
```
YourWildDad, line 444: Non-blocking edge: if `removeReaction(previous)` throws during the `[了解] → [在做了]` replace, the catch only logs but the ledger entry is still deleted on the next line, then a new `[在做了]` is added and stored. Net effect on a Feishu remove-API failure: the old `[了解]` is orphaned on the message (no longer tracked, so the reply-time clear can't remove it) while `[在做了]` is tracked. Minor visual artifact under API failure only. Optional: keep the old entry tracked when removal fails so reply-time clear can retry it.

packages/dreamux/tests/codex-live.test.ts
```diff
@@ -88,9 +108,158 @@ interface McpServerStatusListResponse {
...
+
+function turnStartRequests(client: RecordingCodexWsClient): RecordedRequest[] {
+  return client.requests.filter((request) => request.method === 'turn/start');
+}
+
 describe('codex live integration', () => {
   const skipRequested = process.env[SKIP_ENV] === '1';
   const detection = detectCodex();
+  const runModelGate =
+    process.env[MODEL_GATE_ENV] === '1' || process.env['CI'] !== 'true';
```
YourWildDad, line 262: Caveat to surface (not a defect): with `CI === 'true'` and `DREAMUX_RUN_LIVE_MODEL_GATE` unset, the load-bearing mid-turn fold gate is `it.skip` (loud-warned). So a green CI run does **not** execute the fold premise — it is verified only by the operator's local/auth'd run (reported as codex 0.137.0). That is reasonable (public CI has no operator model auth) and the skip is loud, but please record in the PR/commit which codex version the fold gate actually passed against, so the premise's provenance is auditable. If an auth'd runner ever exists, a scheduled job gating on this would close the loop.

YourWildDad (commented):
Adversarial implementation review (heterogeneous review side). I read the full diff, the new source, and every test file; I did not take the self-report at face value. Verdict: **mergeable — no blocking items** (submitting as COMMENT because the review bot is the PR author and GitHub blocks self-approval; treat this as an approve). All five locked-constraint checks pass against the actual code.

**1. always-`turn/start` truly landed.** `TurnManager.enqueue` submits one `turn/start` per accepted+deduped message via `submitTurnStart` and returns a delivery result. `activeTurnId`, the idle/mid-turn branch, the `turn/steer` helper, the `NoActiveTurn` fallback, and the old `drainLoop`/queue/coalescing are all gone — no half-deleted state. `runTurn`/`subscribeTurnCollection` survive only as a test/diagnostic helper and are not referenced anywhere in `src/` outside `events.ts`, so the per-turn notification-listener accumulation is eliminated from the production path.

**2. The live gate is genuinely live, not fake-simulated.** `tests/codex-live.test.ts` spawns a real codex app-server (`CodexProcess`, version-gated ≥0.136.0) and drives it through a real model. The fold proof is load-bearing and real: inject A (forces a `sleep 6` `exec_command`), confirm the turn is active (`turn/completed` == 0), inject B mid-turn, then after settle assert `turnStartRequests` == 2 **and** `turn/completed` == 1 — i.e. two `turn/start`s collapse into a single turn (not a parallel turn, not a rejected/queued submission), and the model actually replies to B's marker through the Feishu MCP `reply` (real processing, not just RPC acceptance), with attribution to B's `message_id` and the full `[了解] → [在做了] → removed` reaction sequence. The only fakes are the Feishu bot (correct — the channel is external) and a recording subclass of the real WS client. Skip conditions are reasonable and loud: missing codex fails loudly; `DREAMUX_SKIP_LIVE_CODEX=1` is a loud opt-out; the model-fold gate runs by default locally and loud-warns in CI (public CI has no operator codex model auth). See the one caveat inline.

**3. Reaction tri-state + ledger.** Per-message `InboundReactionLedgerEntry{reactionId,state}`; `[了解]` added in the `onAccepted` hook before submission (after access-pass + dedupe-miss), `[在做了]` on `turn/start` acceptance, removed on MCP reply. clear-before-add/replace reconciliation via `pendingReceivedReactionClears` is preserved and hardened (re-checked after the remove and after the add). Drop and duplicate `message_id` correctly get no reaction (verified by smoke tests).

**4. Failure policy A honored.** Submission failure keeps `[了解]` + logs; post-acceptance turn failure may leave `[在做了]` with no cleanup. No observer was sneaked back in.

**5. No new HOL / loss / leak.** Each inbound is independent and async; no dreamux-side queue or completion gate. The cross-chat fold smoke test (3 inbound → 3 `turn/start` → 1 completion, all three `<feishu_message>` blocks present) shows no loss and no dreamux coalescing.

Three non-blocking nits inline. None violate a locked constraint or reintroduce HOL; fine to merge as-is or fold into a follow-up.

PR #157, opened 2026-06-08
Fix beta CLI cold-start TDZ regression

YourWildDad:
## Root cause

`@excitedjs/dreamux@0.13.0-beta.56` could crash before printing `dreamux --version` after the top-level `agents[]` config refactor. The built CLI cold-start path exposed an ESM initialization cycle: config loading imports the builtin runtime catalog, builtin runtime modules import path helpers, and those path helpers loop back through `platform/paths`. A couple of module-top-level reads touched imported bindings while they were still in the temporal dead zone.

## Minimal fix

- Delay `BUILT_IN_DEFAULTS` fallback reads in `platform/paths` until `getRuntimeConfig()` is called.
- Re-export the Codex socket path budget binding without copying it at module initialization time.

This keeps the existing public helper names and avoids a larger import graph refactor in the beta fix.

## Added harness

Adds a `smoke-built-cli` Rush command. The Dreamux package script starts the built `bin/dreamux --version` entry in a fresh Node process, which catches cold-start ESM initialization failures against compiled output.

The new gate runs:

- in CI immediately after `rush build`;
- in stable release before publish;
- in beta release before prerelease apply / pack / publish.

## Validation

Commands run locally:

```sh
npm pack @excitedjs/dreamux@0.13.0-beta.56
# install the packed beta into a temporary local prefix and run .bin/dreamux --version
node common/scripts/install-run-rush.js update
node common/scripts/install-run-rush.js build
node common/scripts/install-run-rush.js smoke-built-cli
packages/dreamux/bin/dreamux --version
node packages/dreamux/dist/cli/dreamux.js --version
node common/scripts/install-run-rush.js lint
node common/scripts/install-run-rush.js typecheck
npx vitest run tests/runtime-paths.test.ts tests/dispatcher-codex-home.test.ts tests/global-config.test.ts tests/bin-launcher.test.ts
DREAMUX_SKIP_LIVE_CODEX=1 node common/scripts/install-run-rush.js test
node common/scripts/install-run-rush.js change --verify --target-branch origin/next --no-fetch
```

Static coverage checked:

- existing ESLint would not have caught this because it currently enforces the synchronous-blocking-IO gate, not import cycles;
- `madge` can see the relevant cycles, but it also reports existing broader cycles, so turning it into an immediate hard gate would be noisy;
- `dpdm` and default `dependency-cruiser --no-config` did not catch the actual source graph with the current TypeScript + `.js` specifier setup.

## Residual risk

The import cycle still exists. This PR removes the module-top-level TDZ reads that broke the beta CLI and adds a cold-start guard for the built CLI path; it is not full import-cycle governance. A broader cycle cleanup/static report can follow separately.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,6 +173,8 @@ jobs:
       # built first or dependents fail to find its types on a clean runner.
       - name: rush build
         run: node common/scripts/install-run-rush.js build
+      - name: rush built CLI smoke
+        run: node common/scripts/install-run-rush.js smoke-built-cli
       - name: rush typecheck
         run: node common/scripts/install-run-rush.js typecheck
       # Enforce the issue #85 synchronous-blocking-IO ban (n/no-sync + import /
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -173,6 +173,7 @@ jobs:
         run: |
           node common/scripts/install-run-rush.js install
           node common/scripts/install-run-rush.js build
+          node common/scripts/install-run-rush.js smoke-built-cli
 
       - name: Publish changed packages (rush native)
         env:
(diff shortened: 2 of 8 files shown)


YourWildDad:
Closing in favor of #158, which supersedes this minimal TDZ hotfix with an architecture-level config import-cycle fix and keeps the useful built CLI smoke guard.

PR #153, opened 2026-06-08
Normalize config to top-level agents[] (#148)

YourWildDad:
Normalizes runtime configuration into a top-level named `agents[]` array;
dispatchers and TeamMates reference a runtime by id via `agentRuntime` and carry
no runtime config block. Absorbs #146 (provider self-reports config + diagnostic)
and folds in the #147 codex 0.137 version gate.

## Schema (BREAKING — #98 fail-loud, no migration)

```jsonc
{
  "agents": [
    { "id": "codex",  "provider": "builtin:codex",       "config": { ... } },
    { "id": "claude", "provider": "builtin:claude-code", "config": { ... } }
  ],
  "dispatchers": [
    { "id": "d1", "agentRuntime": "codex", "channels": [ { "provider": "builtin:feishu", "config": {...} } ] }
  ]
}
```

- config lives ONLY in `agents[]`; dispatcher/teammate just reference an agent id.
- channels stay inline (unchanged).
- An inline `dispatchers[].runtime`, a missing/dangling `agentRuntime`, a
  duplicate `agents[].id`, or a non-array `agents` all fail loud with rebuild
  guidance. `onboard` writes the new shape. A pre-#148 config.json must be
  migrated (re-onboard or hand-edit) — see the bundled rush change.

## Design (M2 contract)

The FILE schema breaks, but in-memory `DispatcherConfig.runtime` stays the
resolved `{ provider, config }`: at load, each `agentRuntime` id is resolved
against the parsed `agents[]` map and populated into `.runtime`. Downstream
readers (createRuntime via `dispatcherCodexConfig`, doctor, npm-detection,
daemon) are unchanged. `stringifyConfig` is the in-memory→file translator.

- builtin→config import cycle severed (refs→`registry/`, primitives→
  `config/validate.ts`, codex/claude config+defaults+readers→builtins,
  `provider.readConfig` implemented); config registers builtins idempotently and
  drops `WIRED_RUNTIME_CONFIG_READERS`.
- TeamMates resolve their OWN agent id → own `{provider, config}`, which
  structurally fixes the cross-provider spawn bug (claude TM under codex
  dispatcher). Identity store persists `agent_runtime` (was `provider_ref`),
  fail-loud on a legacy record.
- doctor is provider-self-reported via a new `diagnostic` capability
  (bin checks + codex-home + codex 0.137 floor); core doctor no longer branches
  on runtime identity.

## Verify / residuals (triaged, non-blocking)

[...]

PR #158, opened 2026-06-08
Break config import cycle via caller-composed builtins

YourWildDad:
## Problem

`#153` (the `#148` `agents[]` normalization) merged to `next` and broke the runtime: the built CLI crashes on cold start, before printing anything.

```
$ node packages/dreamux/dist/cli/dreamux.js --version
ReferenceError: Cannot access 'BUILT_IN_DEFAULTS' before initialization
    at platform/paths.js:35
```

## Root cause — a static ESM import cycle

`#148` made `config/config.ts` import `registerBuiltinAgentRuntimeProviders` from `agent-runtime/catalog.ts` so config could self-register the builtin providers for parsing. That closed a static import cycle:

```
config/config.ts → agent-runtime/catalog.ts → agent-runtime/builtin/* → platform/paths.ts → config/config.ts
```

`platform/paths.ts` reads `BUILT_IN_DEFAULTS` at module top level (`let currentConfig = BUILT_IN_DEFAULTS`), so under that cycle the binding is read in its temporal dead zone. `tsc` (hoisted, transpiled) and `vitest` never surfaced it — only the compiled artifact run in a fresh Node process did.

## Why not the minimal patch (#157)

#157 defers the TDZ read (lazy `BUILT_IN_DEFAULTS`) and re-exports without copying. By its own admission *"the import cycle still exists"*. That treats the symptom; the inverted dependency (a schema/parse leaf reaching up into the runtime catalog) remains a latent trap for the next top-level read. This PR removes the cycle at its root and **supersedes #157**, while keeping #157's genuinely useful `smoke-built-cli` gate.

## Fix — caller-composed builtins

`config/config.ts` becomes a true schema/parse leaf that never imports the catalog. Builtin registration moves to the composition layer, matching the provider-self-reported design and the seam `cli/server.ts` already used (it passes its own factory-bearing registry to `loadConfig`):

- New `agent-runtime/load-config.ts` exports `loadConfigWithBuiltins` / `loadOrInitConfigWithBuiltins`: build a registry, register the builtins idempotently, then delegate to `config`.
- `doctor` / `daemon` / `onboard` switch to the wrapper; `cli/server.ts` is unchanged (already composes).
- A bare `loadConfig` against a registered-but-unimplemented builtin now fails loud in `readAgents` ("registered but not runnable"), with a hint pointing at `loadConfigWithBuiltins`.
- External `npm:` providers still load inside `readConfigFile` via the dynamic-import loader (unchanged).

**Invariant established:** `config/config.ts` and `platform/paths.ts` must never statically import `agent-runtime/catalog.ts` or any `agent-runtime/builtin/*`; `agent-runtime/load-config.ts` must never be imported by `platform/paths.ts` or a builtin. Documented in `.agents/decisions/agents-config-normalization.md`.

## Regression guard

Adopts #157's `smoke-built-cli` Rush command — a fresh-Node `bin/dreamux --version` run wired into CI (after `rush build`) and before release publish. This exercises the compiled cold-start path that unit tests structurally cannot. ESLint enforces the sync-IO ban, not cycles; `madge` is noisy on pre-existing type-only cycles, so the smoke gate is the practical guard.

## Verification

- `node packages/dreamux/dist/cli/dreamux.js --version` → `0.0.0` (was a crash); `serve` and `doctor` load all modules cleanly and fail-loud on missing config rather than TDZ.
- `madge --circular dist/` (runtime truth, types erased): config↔paths cycle **gone**; only the 2 pre-existing intra-`builtin/codex/` cycles (`provider↔diagnostic`, `provider↔runtime`) remain — neither touches `config`/`paths`, neither TDZs. The `provider↔diagnostic` one is `#148`-introduced and tracked for separate follow-up.
- `rush smoke-built-cli` passes.
- Full suite: **518 passed, 1 skipped** (live-codex opt-in). Test migration was suite-classified: only call sites that parse a builtin agent moved to the wrapper; pre-`readAgents` error-path tests and the deliberate injected-registry tests stay on bare `loadConfig`.
- `rush lint`, `rush typecheck` clean.

Refs #148. Supersedes #157.

### DIFF
--- a/.agents/decisions/agents-config-normalization.md
+++ b/.agents/decisions/agents-config-normalization.md
@@ -74,9 +74,39 @@ deferred and carry per-dispatcher credentials).
 - `DispatcherConfig` also keeps the referenced `agentRuntime` id in memory so the
   config round-trips back to the file shape (`stringifyConfig` is the in-memory →
   file translator; `DEFAULT_CONFIG_JSON` is routed through it).
-- The builtin runtime providers are registered into the registry idempotently
-  (guarded by `getImplementation(id) === undefined`) so config can register them
-  for parsing while the server's factory-bearing registration still wins.
+- The builtin runtime providers must already be registered in the registry
+  `readAgents` validates against (each agent's `config` is parsed through its
+  provider's `readConfig`). Registration is **caller-composed, not done by
+  `config/config.ts`**: the config module is a schema/parse leaf and never
+  imports the runtime catalog. `cli/server.ts` hands `loadConfig` a
+  factory-bearing registry it built; the leaf entry points
+  (doctor / daemon / onboard) call `loadConfigWithBuiltins`
+  (`agent-runtime/load-config.ts`), which builds a registry, registers the
+  builtins idempotently (guarded by `getImplementation(id) === undefined`), then
+  delegates to `loadConfig`. A bare `loadConfig` against a registry with a
+  registered-but-unimplemented builtin fails loud in `readAgents` ("registered
+  but not runnable"). External `npm:` providers still load inside
+  `readConfigFile` via the dynamic-import loader.
+
+#### De-cycle (post-#148 hotfix)
+
+The first #148 cut had `config/config.ts` import
+`registerBuiltinAgentRuntimeProviders` from `agent-runtime/catalog.ts` directly.
+That formed a static ESM import cycle —
+`config/config.ts → catalog → builtin/* → platform/paths.ts → config/config.ts`
+(`platform/paths.ts` reads `BUILT_IN_DEFAULTS` at module top level) — which
+crashed the built CLI on cold start with
+`Cannot access 'BUILT_IN_DEFAULTS' before initialization` (a temporal-dead-zone
+read). `tsc` and `vitest` (transpiled, hoisted) did not surface it; only the
+built artifact did. The fix moves registration out of the leaf to the
+caller-composed `loadConfigWithBuiltins`, severing the upward edge at its root
+(rather than deferring the TDZ read). **Invariant: `config/config.ts` and
+`platform/paths.ts` must never statically import `agent-runtime/catalog.ts` or
+any `agent-runtime/builtin/*` module; `agent-runtime/load-config.ts` must never
+be imported by `platform/paths.ts` or a builtin.** Guarded by the
+`smoke-built-cli` gate (a fresh-Node `bin/dreamux --version` run in CI and
+before release publish), which exercises the compiled cold-start path that the
+unit tests cannot.
 
 ### Provider-self-reported doctor diagnostics (issue #146 doctor half)
 
--- a/.agents/decisions/agents-config-normalization.md
+++ b/.agents/decisions/agents-config-normalization.md
@@ -100,13 +100,31 @@ crashed the built CLI on cold start with
 read). `tsc` and `vitest` (transpiled, hoisted) did not surface it; only the
 built artifact did. The fix moves registration out of the leaf to the
 caller-composed `loadConfigWithBuiltins`, severing the upward edge at its root
-(rather than deferring the TDZ read). **Invariant: `config/config.ts` and
-`platform/paths.ts` must never statically import `agent-runtime/catalog.ts` or
-any `agent-runtime/builtin/*` module; `agent-runtime/load-config.ts` must never
-be imported by `platform/paths.ts` or a builtin.** Guarded by the
-`smoke-built-cli` gate (a fresh-Node `bin/dreamux --version` run in CI and
-before release publish), which exercises the compiled cold-start path that the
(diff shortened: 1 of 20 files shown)


YourWildDad:
Review at head `d58ec50d52b4a62baac0187aa2597a81a0dece84` against base `next` (`b6da0eb80269a816a73b7ff609d5bc4a03646221`).

**Verdict: would request changes / BLOCK**

The PR does remove the old `config -> catalog -> builtin -> platform/paths -> config` cold-start cycle, and the built CLI smoke path passes. However, the documented architecture invariant is still not satisfied: `packages/dreamux/src/config/config.ts` still statically value-re-exports builtin config modules from `agent-runtime/builtin/*` (`codex/config.js` and `claude-code/config.js`). The dist graph still includes `config/config.js -> agent-runtime/builtin/{codex,claude-code}/config.js`.

That conflicts with the decision record's invariant that `config/config.ts` must never statically import `agent-runtime/builtin/*`. In other words, the current head appears to fix this specific TDZ path, but the claimed schema/parse-leaf boundary is not yet true.

Please either move those compatibility exports behind a neutral/caller-owned surface, or update the invariant if this builtin-config coupling is intentional.

Non-blocking P2: `packages/dreamux/src/cli/server.ts` still has a comment implying `loadConfig` performs idempotent builtin registration itself. Registration is now caller-composed / server-registry-owned, so that comment should be updated to avoid misleading future changes.

Validation run by the reviewer included: `rush update`, `rush build`, `rush smoke-built-cli`, direct built `dreamux --version`, focused Vitest for `global-config` and `onboard` (66 passed), `.agents/scripts/check.sh`, `git diff --check`, and a public-safety grep. Full `rush test` and full leak scan were not run.

Formal REQUEST_CHANGES was attempted but GitHub rejected it because the reviewer account owns the pull request, so I am posting the substantive review as a top-level comment instead.

YourWildDad:
Thanks — pushed `8580e09`. I verified the finding and acted on it, but landed on the second option you offered (update the invariant) rather than moving the re-exports. Reasoning:

**The factual observation is correct, the BLOCK conclusion over-reads my invariant.** `config/config.ts` does keep a static value re-export of the two builtin **config** modules (`codex/config.ts`, `claude-code/config.ts`). But that edge does **not** close a cycle:

- Both builtin config modules import only `registry/index.ts` and `config/validate.ts` — never `platform/paths.ts`, never `config/config.ts`. They are genuine leaves.
- So `config → builtin/*/config` is a one-way, leaf-ward edge. `madge --circular dist/` (types erased = runtime truth) confirms it: zero `config`/`paths` cycle; only the two pre-existing intra-`builtin/codex/` cycles remain. The cold-start TDZ is genuinely dead, not just deferred.
- These re-exports are the intentional **M2 back-compat surface** (the decision record keeps them so ~a dozen non-builtin callers keep their `config/config.js` import paths). Reversing that is real churn across all those callers for **zero** cycle-safety gain.

The actual defect was my own invariant wording in the same PR — I wrote "never import any `builtin/*`", which is broader than the real hazard (reaching `platform/paths.ts`) and contradicted the M2 re-export. Fixed in `8580e09`:

- Tightened the invariant to target the hazard precisely: `config/config.ts` / `platform/paths.ts` must not import `catalog.ts` or any builtin **runtime/provider/transport/paths** module; the builtin **config** leaves are an explicitly-allowed re-export, *conditional on* them staying `platform/paths`-free.
- Added a load-bearing guard comment to each builtin config file pinning that leaf property to the #148 cycle.

**P2 (server.ts comment):** valid, fixed — the comment claimed `loadConfig` self-registers builtins; registration is now caller/server-owned.

Net: schema/parse-leaf boundary is true in the sense that matters (no path to `platform/paths`); the remaining edge is a deliberate, guarded, cycle-free leaf re-export.

YourWildDad:
Re-review at head `8580e0944a141b34641f55e0fca3667e2d8aad2e`.

**Verdict: would approve / CLEAN**

The follow-up delta is documentation/comment-only and resolves the previous findings:
- The remaining `config/config.ts` builtin config re-exports are now explicitly documented as the intentional M2/back-compat surface, and the invariant is narrowed to the actual cycle hazard: imports that can reach `platform/paths.ts` through builtin runtime/provider/transport/paths modules.
- The stale `cli/server.ts` comment was corrected to describe the server-owned populated registry passed into `loadConfig`, rather than implying that `loadConfig` self-registers builtins.

Validation for the follow-up delta included head/base verification, diff inspection for `d58ec50d..8580e094`, targeted invariant greps, `git diff --check`, and the KB check script. Build/smoke/tests were not rerun for this comment/docs-only delta; the prior behavior validation remains applicable.

No remaining P0/P1/P2 findings on the current head. Formal approval was attempted but GitHub rejected it because the reviewer account owns the pull request, so this substantive verdict is posted as a top-level comment.

Issue #164, opened 2026-06-09
Align MCP push-back task notifications + Feishu message format with native Claude Code


YourWildDad:
## Background

Dreamux's dispatcher receives TeamMate completions through the MCP push-back mechanism, which currently reuses Claude Code's task-notification channel. Two behaviors diverge from native Claude Code and should be aligned. This issue is the shared input for a heterogeneous design negotiation (see "Negotiation process" below) — it states the requirement, not the solution.

> Implementers: take an adversarial stance toward the claims below. Verify each premise against the actual Claude Code source (vendored in this workspace) and the current Dreamux implementation **before** designing or coding. Do not hardcode numbers or assumptions taken verbatim from this description.

## 1. Long task-notification truncation

**Current (claimed):** the MCP push-back reuses task notifications but does **not** apply the native length-truncation. A large TeamMate result is inlined into the notification in full.

**Native Claude Code (reference, verify):** when a task notification exceeds a length threshold (reportedly ~2000 chars), the full content is spilled to a file and the notification inlines only the **file path**, not the full body.

**Ask:** align Dreamux's push-back with native behavior — when a notification exceeds the threshold, write the full body to a file and inline the file path instead of the entire text.

To investigate:
- What is the native threshold and the exact spill/truncation logic? (read the Claude Code source, don't assume 2000)
- Where does Dreamux currently build the push-back notification, and why is truncation bypassed?
- Where should the spill file live, and what should the inlined reference look like to match native?

## 2. Feishu message format

**Ask:** the Feishu channel's outbound message format should align with the native channel message format.

**Maintainer design direction (locked):**
- Add a `Channel input` function to the existing **Agent Runtime** interface; delegate assembly of the channel data format to each Agent Runtime (each runtime owns formatting its own channel payload).
- The **Codex** runtime already formats correctly: the `feishu_message` block currently delivered is the format Codex designed for itself and is the reference ("native") shape. **Codex's channel format does NOT need changing.**
- Only the **Claude** runtime's channel format needs adjusting to match. Scope the change to the Claude runtime's channel payload assembly.

To investigate:
- Where is the Agent Runtime interface defined, and how does a channel-input function fit it?
- How does Codex currently assemble the `feishu_message` shape (the target format)?
- Where does the Claude runtime assemble its channel payload, and what is the minimal change to make it match Codex's shape?

## Negotiation process

[...]

PR #165, opened 2026-06-09
Bound completion push-back + assemble channel input per runtime (#164)

YourWildDad:
Implements the finalized design for #164. Two separate commits.

## ① Spill long teammate completions to a file

Completion push-back inlined the full teammate result into the dispatcher
context unbounded. Added a shared spill valve:

- `platform/paths.ts`: neutral `teamMateCompletionOutputPath(source, id)` →
  `/tmp/teammate-{source}-{id}.output` (sanitized).
- `agent-runtime/completion-body.ts`: neutral `resolveCompletionBody` used by
  both builtins (no cross-builtin import). Inline budget default 32000 chars,
  `TASK_MAX_OUTPUT_LENGTH` override, clamped to 160000 — mirroring claude-code's
  native task-output budget. Over budget → full result written to a `0600` file
  (async `fs/promises` only, per #85), only the path inlined.
- **claude-code**: dropped the fake `<task-notification>` XML entirely. The old
  shape mimicked claude-code's *real* task-notification system, so the model
  could mistake the fabricated `task-id`/`output-file` for a live background
  task. Now delivers a plain status-varied English user turn with
  `isSynthetic:false`. Capability kind `claudeCodeTaskNotification` →
  `claudeCodePlainTurn`.
- **codex**: kept the `<teammate_session_completion>` developer-role wrapper;
  only the body changes (inline result or spill pointer).
- `CompletionEnvelope.status` widened to `completed|failed|stopped`
  (`TurnSettledSignal` already carried `stopped`); the teammate service stops
  folding `stopped`→`failed`. In-process contract only.

## ② Assemble channel input in each runtime as `<channel>`

The Feishu channel rendered the final `<feishu_message>` XML, so the channel —
not the runtime — owned the inbound format. Moved assembly into each runtime:

- `turn.ts`: `InboundTurnInput` widens to
  `{ text; sourceId; source?; attrs?; body?; attachments? }` — channel-neutral
  field names; the literal `"feishu"` is data the channel supplies as `source`.
  Adds shared neutral `renderChannelInput`/`renderChannelBlock` (mirroring
  claude-code's `wrapChannelMessage`: safe-key validation + XML-escaping); both
  builtins reuse them with no cross-builtin import.
- `feishu-message.ts` returns structured `{ attrs, body, attachments }` instead
  of a pre-rendered string; `body` keeps the full prior inner content (message +
  mentions + fallback + attachment refs + group-bots), so no model-visible
  content is dropped.
- Both runtimes render the native `<channel source="feishu" …>` envelope;
  codex's old `<feishu_message>` wrapper is retired. Identical output today, the
  per-runtime seam kept for future divergence (e.g. claude inlining image blocks
  vs codex text refs).
- **no-leak invariant clarified, not retired**: routing/identity *decisions*
  (`chat_id`/`sender_id`/`message_id`) still never cross into the runtime;
  opaque *display* passthrough (`attrs`) may. Documented in both `CLAUDE.md`
  files + a new `.agents/decisions/channel-input-runtime-assembly.md`. Dispatcher
  base prompt + Feishu MCP tool descriptions updated to reference the `<channel>`
  block.

## Verification

- `rush build` / `rush test` / `rush lint` green.
- KB `check.sh` link-resolution + orphan-reachability verified.
- New tests: `completion-body.test.ts` (inline/spill/perms/env), channel-format
  tests (both runtimes render the same `<channel>`; attachments + group_bots
  survive; channel no longer emits `<feishu_message>`); existing completion +
  smoke/e2e tests updated.

### DIFF
--- a/.agents/decisions/README.md
+++ b/.agents/decisions/README.md
@@ -9,7 +9,7 @@ topic slugs remain reviewable and merge-friendly.
 | Theme | Records |
 |---|---|
 | Repository shape | [rush-pnpm-monorepo](rush-pnpm-monorepo.md), [install-model](install-model.md) |
-| Runtime architecture | [top-level-design](top-level-design.md), [issue-110-epic-closure](issue-110-epic-closure.md), [provider-references-and-capability-registry](provider-references-and-capability-registry.md), [agent-runtime-provider](agent-runtime-provider.md), [channel-provider](channel-provider.md), [server-hosted-teammate](server-hosted-teammate.md), [providerized-config-state-compatibility](providerized-config-state-compatibility.md), [agents-config-normalization](agents-config-normalization.md), [global-config-dir](global-config-dir.md), [logging](logging.md), [feishu-inbound-attachments](feishu-inbound-attachments.md) |
+| Runtime architecture | [top-level-design](top-level-design.md), [issue-110-epic-closure](issue-110-epic-closure.md), [provider-references-and-capability-registry](provider-references-and-capability-registry.md), [agent-runtime-provider](agent-runtime-provider.md), [channel-provider](channel-provider.md), [server-hosted-teammate](server-hosted-teammate.md), [providerized-config-state-compatibility](providerized-config-state-compatibility.md), [agents-config-normalization](agents-config-normalization.md), [global-config-dir](global-config-dir.md), [logging](logging.md), [feishu-inbound-attachments](feishu-inbound-attachments.md), [channel-input-runtime-assembly](channel-input-runtime-assembly.md) |
 | Public surface | [cli-and-package-naming](cli-and-package-naming.md), [dispatcher-tm-boundary](dispatcher-tm-boundary.md), [dispatcher-tm-packaging](dispatcher-tm-packaging.md), [global-bin-onboard-serve](global-bin-onboard-serve.md), [global-config-dir](global-config-dir.md) |
 | Release and safeguards | [npm-release-oidc](npm-release-oidc.md), [anti-leak-guardrail](anti-leak-guardrail.md), [no-sync-io-lint-gate](no-sync-io-lint-gate.md) |
 
@@ -19,6 +19,7 @@ topic slugs remain reviewable and merge-friendly.
 - [anti-leak-guardrail](anti-leak-guardrail.md)
 - [agent-runtime-provider](agent-runtime-provider.md)
 - [channel-provider](channel-provider.md)
+- [channel-input-runtime-assembly](channel-input-runtime-assembly.md)
 - [cli-and-package-naming](cli-and-package-naming.md)
 - [dispatcher-tm-boundary](dispatcher-tm-boundary.md)
 - [dispatcher-tm-packaging](dispatcher-tm-packaging.md)
--- a/.agents/decisions/channel-input-runtime-assembly.md
+++ b/.agents/decisions/channel-input-runtime-assembly.md
@@ -1,0 +1,62 @@
+# Channel input is assembled by each runtime; no-leak clarified to routing-only
+
+- **Status:** Accepted
+- **Date:** 2026-06-09
+- **Affects:** the runtime contract
+  [`/packages/dreamux/src/agent-runtime/turn.ts`](/packages/dreamux/src/agent-runtime/turn.ts),
+  both builtins' `channelInput`, the Feishu channel message layer
+  [`/packages/dreamux/src/channel/feishu/feishu-message.ts`](/packages/dreamux/src/channel/feishu/feishu-message.ts),
+  the no-leak boundary in
+  [`/packages/dreamux/CLAUDE.md`](/packages/dreamux/CLAUDE.md) and
+  [`/packages/dreamux/src/agent-runtime/CLAUDE.md`](/packages/dreamux/src/agent-runtime/CLAUDE.md)
+- **PR / Issue:** [#164](https://github.com/excitedjs/dreamux/issues/164)
+
+## Context
+
+Inbound Feishu messages were rendered to their final `<feishu_message>` XML by
+the channel layer (`formatFeishuMessageForRuntime`), which handed the runtime a
+single pre-rendered `InboundTurnInput.text`. Both builtins consumed that string
+verbatim, so the channel — not the runtime — owned the model-visible inbound
+format. We wanted each runtime to own assembling its own channel block (so the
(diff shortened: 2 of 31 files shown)


YourWildDad:
Gate verdict: REQUEST CHANGES.

I reviewed PR #165 against the locked #164 design, using the checked-out branch plus `gh pr diff 165`.

## Must-fix

- `rush.json:50` + `packages/channel/feishu-channel/src/inbound.ts:88` / `packages/channel/feishu-channel/src/inbound.ts:115` — the workspace still contains an active `@excitedjs/feishu-channel` package that exports `formatFeishuMessageForCodex` and emits `<feishu_message>`. The locked design says the Codex `<feishu_message>` format is retired and grep-clean; leaving a Rush project and its tests asserting the old wrapper keeps a second agent-facing Feishu formatter alive.

## Should-fix

- `packages/dreamux/README.md:316` — the verification path still says the server injects a `<feishu_message>` block; that public doc now contradicts the runtime-owned `<channel source="feishu" ...>` design.
- `packages/dreamux/src/dispatcher-service/teammate/service.ts:385` — the comment says a `stopped` turn maps to `failed`, but the code now correctly passes `settled.status` through at `packages/dreamux/src/dispatcher-service/teammate/service.ts:423`; update the comment so the widened `CompletionEnvelope.status` contract is not documented backwards.

## Nice-to-have

- `packages/dreamux/src/agent-runtime/completion-body.ts:45` — `Number.parseInt()` makes partially numeric env values effective (for example, a malformed value starting with digits), while the comment says non-integers fall back to the default. Either parse strictly or update the comment/tests to state the intended behavior.

## Clean

- `packages/dreamux/src/agent-runtime/completion-body.ts:57` + `packages/dreamux/src/platform/paths.ts:230` — completion spill is neutral, async-only, defaults to 32000 chars, honors `TASK_MAX_OUTPUT_LENGTH`, clamps to 160000, writes the full result with 0600, and sanitizes source/id segments before building `/tmp/teammate-{source}-{id}.output`.
- `packages/dreamux/src/agent-runtime/builtin/claude-code/runtime.ts:152` + `packages/dreamux/src/agent-runtime/builtin/claude-code/runtime.ts:363` — Claude completion delivery drops `<task-notification>`, uses the required status-varied natural-language lines, submits as `isSynthetic:false`, and still awaits the serial queue so accepted/failed retry semantics are preserved.
- `packages/dreamux/src/agent-runtime/builtin/codex/runtime-support.ts:73` + `packages/dreamux/src/agent-runtime/builtin/codex/runtime.ts:512` — Codex preserves the developer-role `<teammate_session_completion>` injection path and swaps only the body to inline text vs spill pointer before the existing trigger turn.
- `packages/dreamux/src/agent-runtime/turn.ts:29` + `packages/dreamux/src/channel/feishu/feishu-channel.ts:351` + `packages/dreamux/src/agent-runtime/builtin/codex/runtime.ts:434` + `packages/dreamux/src/agent-runtime/builtin/claude-code/runtime.ts:312` — the new `InboundTurnInput` fields are neutral (`source`, `attrs`, `body`, `attachments`), Feishu still owns routing/gating, and both runtimes render channel input through the shared helper instead of interpreting `chat_id` or `sender_id`.
- `packages/dreamux/src/agent-runtime/turn.ts:80` — `renderChannelBlock` uses the required safe-key regex behavior and XML-escapes attribute values; `packages/dreamux/tests/channel-input-format.test.ts:11` covers byte shape, escaping, and unsafe-key dropping.
- `packages/dreamux/CLAUDE.md:58` + `packages/dreamux/src/agent-runtime/CLAUDE.md:69` + `.agents/decisions/channel-input-runtime-assembly.md:31` — the no-leak clarification is logically consistent: routing decisions stay in the channel layer; opaque display passthrough is allowed.
- Test coverage is meaningful for the locked behavior: spill inline/spilled/perms/env/clamp in `packages/dreamux/tests/completion-body.test.ts:27`, shared channel format in `packages/dreamux/tests/channel-input-format.test.ts:11`, Claude plain completion and channel wrapping in `packages/dreamux/tests/claude-code-runtime.test.ts:595`, and Codex wrapper/spill behavior in `packages/dreamux/tests/codex-completion.test.ts:122`.

## Verification

- `node common/scripts/install-run-rush.js build`: PASS. Rush reported a local worktree hash warning and continued without diff-based state; all 4 project operations completed successfully.
- `node common/scripts/install-run-rush.js test`: PASS WITH WARNINGS. All project operations completed; the warning bucket was test stderr/log output, not a failed assertion.
- `node common/scripts/install-run-rush.js lint`: PASS. All lint operations completed successfully.
- Leak scan: PASS. I ran a gitleaks-style pattern scan over `gh pr diff 165 --patch`; no restricted ID, token, private mirror/domain, email, or machine-local absolute-path patterns were found. A secondary broad scan only matched the expected `/tmp/teammate-...` spill templates/tests.

YourWildDad:
Addressed in `ec2faf1`. Build / test / lint all green locally (`node common/scripts/install-run-rush.js build|test`, `rush lint`), and `grep -r '<feishu_message>' packages/` is now clean tree-wide.

**MUST-FIX — `@excitedjs/feishu-channel` placeholder**
Confirmed your read: it's an unwired #4-refactor scaffold (`version 0.0.0`, `shouldPublish:false`, no importers), not a live second formatter. Aligned (not removed — keeps the scaffold coherent for #4): `formatFeishuMessageForCodex` now emits `<channel source="feishu" …>` instead of the retired `<feishu_message>` wrapper, and its tests were updated to match. Also swept the remaining textual references so the grep is clean: both channel-package `CLAUDE.md`s, the dreamux README verification path, and a dreamux comment/test that mentioned the old tag.

**SHOULD-FIX**
- `packages/dreamux/README.md` step 4 now describes the runtime assembling the inbound into a `<channel source="feishu" …>` block (channel hands neutral structured pieces).
- `teammate/service.ts` comment fixed: the settle status (completed|failed|stopped) now passes through to the envelope verbatim — no longer documented as folding `stopped`→`failed`.

**NICE-TO-HAVE**
- `completion-body.ts` now parses `TASK_MAX_OUTPUT_LENGTH` strictly: a non-decimal value (`123abc`, `32k`, `12.5`) falls back to the default instead of partially parsing, matching the doc comment. Added test coverage for those cases.

Note: this PR targets `next`, and `ci.yml` only triggers on `main` PRs, so the local run above is the gate. Re-ran the public-repo leak grep on the diff — clean.

PR #168, opened 2026-06-09
fix(claude-code): return completion accepted at enqueue, not after the turn

YourWildDad:
## Problem

TeamMate completion reverse-delivery **over-delivered**: a finished teammate's result was delivered to the dispatcher twice (verbatim, ~183s apart).

Root cause: claude-code `completionInput` `await`ed the **full** delivery turn and derived `accepted`/`failed` from the turn's **model outcome** — unlike its sibling input methods `channelInput` / `systemInput`, which return at **enqueue** and let the turn outcome flow to `markTurnSucceeded` / `markTurnFailed` → runtime `degraded` + `last_error`. Two consequences:

1. Delivery-accept latency was coupled to the full turn duration (the observed ~183s) — acceptance should be decided at enqueue, decoupled from model thinking time.
2. The result parser classified a result as an error whenever `is_error: true`, even with `subtype: "success"`. So a benign turn was reported `failed`; the neutral retry loop treats `failed` as "never delivered, safe to retry" and records the idempotency id only on `accepted` — so it re-delivered.

codex `completionInput` already returns `accepted` at submit and did not have this bug; this brings claude to parity.

## Fix (minimal)

- `agent-runtime/builtin/claude-code/runtime.ts` `completionInput`: submit-then-serialize — enqueue the completion turn, return `{status:'accepted'}` at enqueue; the turn's eventual outcome flows via `markTurnSucceeded`/`markTurnFailed` → `degraded`/`last_error`. `failed` now means a pre-submit failure only.
- `agent-runtime/builtin/claude-code/stream.ts` `parseResult`: terminal `subtype` is authoritative when present — `subtype:"success"` is not an error even if `is_error:true`; a non-success subtype is an error even if `is_error:false`; absent subtype falls back to `is_error`/`errors`.
- Regression tests in `claude-code-runtime.test.ts` and `claude-code-stream.test.ts`; rush change file.

## Out of scope (deliberately)

Durable completion outbox, delivery state machine, `completionDurability`/`replay` capability fields, and cross-crash/restart exactly-once. Within a server lifetime the existing in-memory dedup + in-flight coalescing already give exactly-once; cross-restart durability is a separate, larger question tracked on its own.

## Review

Undergoing independent adversarial review before merge.


YourWildDad:
Independent adversarial review (heterogeneous engine, separate worktree) verdict: **CLEAN**. Re-ran the full suite locally — `rush update/build/lint/test` all green; vitest 566 passed | 2 skipped (the 2 skips are the opt-in live-codex cases; codex parity tests green). Verified: `accepted` resolves at enqueue independent of model outcome; pre-submit failure still returns `failed`; a failed delivery turn still degrades the runtime with a persisted `last_error` (error reporting intact); `subtype` authoritative in classification; scope held (no outbox / state machine / capability / persisted-format creep); leak self-check clean.

Issue #228, opened 2026-06-14
test: guard package-split production provider parity


YourWildDad:
## Background

Issue #209 split Dreamux provider contracts and built-in providers into npm packages. The package split did create publishable runtime/channel packages, but the review uncovered an important harness gap: the built-in Agent Runtime packages can be loaded through the generic provider loader, while the production path still uses per-runtime core adapters under `packages/dreamux/src/agent-runtime/builtin/*`.

This is not a simple implementation bug. It is a missing architectural invariant in the harness. A package can satisfy all of these checks and still fail the plugin-readiness goal:

- the provider package does not import `@excitedjs/dreamux`;
- `@excitedjs/dreamux` depends on the built-in provider package by default;
- the generic loader can import the real package and construct a provider;
- existing runtime behavior is preserved through a core adapter.

The real invariant is stronger: an external Agent Runtime provider must be usable by the production launcher without adding a provider-specific glue layer in `@excitedjs/dreamux` core.

Related context:

- #209: original package split epic.
- #213: Codex runtime extraction. The first architecture review found that `builtin:codex` was advertised as loadable through the generic provider path, but the real package could not satisfy that path.
- #220: package-boundary guards. These guarded package independence and manifest direction, but did not guard production provider parity or residual core implementation.
- #223: final feature PR where this gap became visible during release review.

## What Went Wrong

PR #213 fixed the review blocker by making the package constructible through the loader: the package added a default factory and standalone fallbacks, so `loadExternalAgentRuntimeProviders({ refs: ['builtin:codex'] })` could load and construct a provider.

That closed the literal review finding, but it did not prove the production path used the same provider seam. Production still constructed runtimes through a core-owned adapter that translated Dreamux's host-shaped context into the neutral provider context and injected host hooks. The decision record then documented that split as intentional later-slice work, which made later reviews validate the wrong invariant.

The missing distinction was:

- **Package constructibility**: the generic loader can import the provider package and call `createRuntime` with a neutral context.
- **Production parity**: the actual dispatcher/team launcher can drive that loaded provider without a provider-specific adapter in core.

The harness only covered the first property. It must cover the second.

## Current Failure Mode

A future runtime provider could compile against `@excitedjs/dreamux-types` and pass import-boundary checks, but still require adding a new file under `@excitedjs/dreamux` to translate host runtime state into the provider's context. If that is required, the plugin contract is incomplete.

The core package must provide one generic host-to-neutral boundary, not one hand-written adapter per runtime.

[...]

PR #229, opened 2026-06-15
refactor(i209): core provider-neutrality cleanup + SDK secret redaction

YourWildDad:
Stacked on `feature/i209-npm-package-split` (the npm package-split branch).

## Summary

This PR finishes the #209 provider-neutrality cleanup for `@excitedjs/dreamux` core. Core now coordinates dispatchers, teams, bindings, provider loading, and routing through neutral contracts from `@excitedjs/dreamux-types`; provider-specific runtime/channel behavior lives in the provider packages.

Key boundaries settled here:

- Core has no built-in runtime/channel implementation trees and does not import or instantiate built-in provider implementation packages directly. Built-ins are loaded through the same provider loader path as package-backed refs.
- Provider packages must echo the seed descriptor from core. Agent runtime providers are validated for the required `systemPrompt` capability shape during load.
- A dispatcher may attach channels from different providers, but a single dispatcher still cannot configure two channels with the same provider ref.
- Bidirectional chat channels and future subscription channels remain separate model shapes: chat channels have channel targets and reply/react-style tools; subscription channels have no chat id / message id / reply / Team binding concept.
- Team MCP remains the owner of Team binding and transfer-back (`bind_channel` / `transfer_back`). The generic `channel-mcp` CLI is only a provider-tool bridge: provider descriptors expose tool metadata, and calls flow through the neutral `channel.invoke_tool` admin method to the live provider session.
- Onboard and doctor/diagnostic behavior is provider-owned where provider-specific checks are needed; core coordinates provider refs and invokes the provider capabilities.

The PR also tightens regression coverage:

- `@excitedjs/dreamux-types` test fixtures are now included in Rush `typecheck`, so provider-authoring examples cannot drift silently.
- Core package-boundary guards reject direct imports/calls into built-in provider implementation packages and reject reintroduced provider-specific adapter source trees.
- README, `.agents` decisions, repo-structure docs, and rush change files have been realigned to the final #209 channel/provider model.

## Verification

Latest pushed head: `dd32dedd5789f7340d8aa257a4ac0c186fabfa73`.

Local checks run before push:

- `node common/scripts/install-run-rush.js change --verify`
- `git diff --check`

GitHub Actions for the latest head are green, including both macOS and Ubuntu `rush — bootstrap / build / typecheck / lint / test` jobs, gitleaks, shellcheck, commit metadata, rush change verification, and the knowledge-base check.

Refs #209.


YourWildDad:
## Final aggregate review for PR #229

Reviewed head: `da3b3106197e655e87e22d236b6be410c7977664` against `feature/i209-npm-package-split`.

Verdict: **not merge-ready as-is**. The main architecture cleanup is moving in the right direction, but the remaining P1 items below need to be fixed, or explicitly split into tracked follow-up work before this PR can claim the #209/#228 cleanup is complete.

The important positive result: the effective reviews agree that this PR addresses the central architecture failure. Core-owned runtime/channel glue is no longer the production integration path: runtime provider implementations have moved out of `@excitedjs/dreamux` core, production paths load providers through the neutral provider loader/catalog shape, external runtime packages should not depend on the core package, and Feishu SDK ownership remains in `feishu-transport`.

I am treating the following as intended design, so these are **not blockers by themselves**:

- A dispatcher may register at most one channel per provider ref.
- Channel binding and transfer-back remain Team MCP/core capabilities; they take `channel_id` plus provider-owned `meta`.
- There is no generic `list_peers` capability.
- Channels are split conceptually into simplex/subscription channels and duplex/chat channels. `chat_id`, `message_id`, `reply`, and `react` are duplex-channel concepts; simplex channels do not need to expose reply/react tools.

Please make those decisions explicit in the decision docs / PR text / changelog so future reviewers do not rediscover the same design from code behavior.

### Blocking items

1. **External runtime `systemPrompt` contract is still under-validated.**
   `packages/dreamux/src/agent-runtime/external-provider.ts` validates several required capabilities, but does not validate the required `systemPrompt` capability. Core later dereferences `runtimeProvider.getCapabilities().systemPrompt.mode` in dispatcher startup paths, so a malformed external provider can load successfully and fail later. This should fail loud at provider load time, with a fixture/test covering the missing-field case.

2. **The #228 anti-regression harness is not yet strong enough.**
   `packages/dreamux-types/tests/fixtures/external-provider.ts` has drifted from the current `ChannelRoutes.deliver` signature, and the package test typecheck is not wired into the Rush bulk path that CI exercises. The external-provider fixture needs to be part of the normal gate. The harness should also guard the architectural invariants that caused this cleanup: no production core-owned runtime/channel glue, no reintroduction of provider-specific core adapters, and no stale single-create-context integration path.

3. **The final channel design record is missing/stale.**
   The repo needs a committed decision record for the final model: Team MCP owns binding/transfer-back; channel `meta` is provider-owned; provider refs are unique per dispatcher; generic `list_peers` is removed; simplex and duplex channel capabilities differ. Without that, the code behavior and review expectations remain ambiguous.

4. **Changelog / KB / README / PR-text drift must be cleaned up.**
   Some existing text still describes superseded intermediate designs, including Channel MCP owning binding, production core-owned adapters, and neutral `channel.list_tools` wording that does not match the current implementation. The public upgrade notes and knowledge base should describe the final design only.

### Non-blocking follow-ups / hardening

- Assert provider descriptor identity consistently, including mismatch between the seed descriptor and loaded provider descriptor.
- Harden Feishu transport diagnostics so raw SDK error objects cannot leak unsanitized data through fallback stderr paths.
- Revisit skipped legacy config/onboard tests and socket-budget/config-validation coverage after the main architecture gate is stable.
- Keep Feishu-specific helper tools separate from the removed generic `list_peers` concept, and document that boundary explicitly if they remain.

Issue #233, opened 2026-06-16
refactor: Service 架构重构（Collection + Service 分层 + Agent 统一模型）


YourWildDad:
# Service Architecture Refactor (Collection + Service Model)

- **Status:** Proposed
- **Date:** 2026-06-16
- **Affects:** `dispatcher-service/` entire module, `platform/paths.ts`, `admin/methods.ts`, state directory layout
- **PR / Issue:** #233
- **Related:** issue #209 follow-up; refines `dispatcher-local-aggregate.md`

## Context

The `dispatcher-service/` object model evolved across multiple iterations and carries several structural problems:

1. **Blurred responsibility boundaries.** `TeamMateAgentService` and `TeamManager` are process-wide god objects mixing collection operations, entity operations, storage access, and runtime management.
2. **Inconsistent naming.** Some are called Service, some Manager, some Store — names don't reflect actual responsibilities.
3. **No entity objects.** Teammates and teams have no corresponding entity objects; all operations live on manager/service classes, separating data from behavior.
4. **Asymmetric layering.** The Team side has a collection + single-entity split (`TeamManager` / `TeamService`), while the Teammate side has one monolithic Service.
5. **Redundant `LiveTeammateRegistry`.** A global registry of live runtimes is unnecessary when each teammate can manage its own runtime.
6. **No unified agent model.** Dispatcher runtime and team leader are conceptually the same thing — an entity with an agent runtime — but implemented as two completely separate code paths.
7. **Delivery logic is duplicated.** Completion delivery to dispatcher runtime and to team leader follow the same pattern but are coded separately.

## Decision

Restructure the service layer around a **Collection + Service pattern**, and unify agent runtime lifecycle management through a shared `TeammateService` entity. The dispatcher *has* an agent (not *is* an agent); the team *has* a leader agent. Delivery flows through domain events, not direct references.

### Symmetric Collection + Service Pattern

Both Team and Teammate follow the same two-level pattern:

| Level | Team side | Teammate side | Responsibility |
|---|---|---|---|
| Collection | `TeamCollection` | `TeammateCollection` | Holds store, create/list/get, factory methods, event emission |
| Entity | `TeamService` | `TeammateService` | Holds its own record/identity, domain operations, runtime lifecycle |

No inheritance between collections or entities. Differences (dispatcher-owned vs team-owned) are expressed through constructor injection of configuration, not subclassing.

### Shared Agent Runtime Entity (has-a, not is-a)

[...]

PR #240, opened 2026-06-25
refactor(team): own per-team construction in TeamService (#233)

YourWildDad:
## Why

The `team_leader` launch policy belongs to the layer that owns the
`team_leader` role. Previously `TeamCollection` built each team's members
`TeammateCollection`, leader, and scheduler, and held `teamMateLaunchPolicy`;
`TeamService` merely received them via `TeamServiceOptions`. That left a
team-layer policy living a level too high and forced an awkward indirection
(`launchPolicyForTeamMate` threaded down from the collection).

## What

Move per-team construction **into** `TeamService`:

- `TeamService` gains a private **sync constructor** that builds its own
  team-scoped `TeammateCollection` + `SchedulerService`, plus two static
  **async factories** `createNew()` / `rebuild()`. This breaks the
  construction cycle (the leader is a `TeammateService` whose launch policy
  is a dep, so the policy owner must exist before the leader starts): the
  closure `(input) => this.teamMateLaunchPolicy(input.identity)` is captured
  at construction and only invoked at leader/teammate launch (by which time
  `deps` is set); the scheduler is created unstarted and started only after
  `leader_` is assigned.
- `teamMateLaunchPolicy` is now a true **private instance method** of
  `TeamService`.
- `TeamCollection` collapses to a thin factory/registry: `depsBase()`
  forwards the shared deps; each call site injects the per-team `evict`
  closure.
- `record` / `leader_` become nullable with `mustRecord()` / `mustLeader()`
  guards + a `view()` helper. Behavior is unchanged (store create/update
  sequence, create-with-prompt vs idle leader, scheduler start gating,
  dissolve cleanup).
- The public `teammates` surface stays **narrow**: the collection is held as
  a private concrete field, exposed via `get teammates(): TeammateOps`, so
  internal lifecycle/factory verbs (`createTeamLeader`, `allocateLeaderName`,
  `stopAll`, `applyWorktreeCleanup`, the workspace-injecting raw `spawn`) are
  not re-exposed to admin callers — `spawnTeamMate` remains the only spawn
  path.

## Review

Reviewed by two independent heterogeneous reviewers (codex + claude). Both
converged on one finding — the `teammates` getter had been widened to the
concrete `TeammateCollection` (a deviation from the design intent of keeping
the narrow `TeammateOps` seam); that is fixed in this PR (private field +
narrowed getter).

## Verification

- `tsc --noEmit` (src + tests) — clean
- `eslint` on both changed files — clean
- `vitest run` (`DREAMUX_SKIP_LIVE_CODEX=1`) — **525 passed | 4 skipped**

### DIFF
--- a/packages/dreamux/src/service/team-service/index.ts
+++ b/packages/dreamux/src/service/team-service/index.ts
@@ -1,4 +1,5 @@
 import type {
+  CompletionEnvelope,
   AgentRuntimeMcpServer,
   ChannelTarget,
   DreamuxLogger,
@@ -15,6 +16,7 @@ import type {
   CompletionInitiator,
   CompletionRouter,
 } from '../completion-router/index.js';
+import { completionKey } from '../completion-router/index.js';
 import type { ChannelBindingStore } from '../channel-binding/store.js';
 import type { ChannelBinding } from '../channel-binding/store.js';
 import { cronMcpServerDescriptor } from '../scheduler/mcp-config.js';
@@ -36,6 +38,7 @@ import {
   type TeamMateRuntimeStatus,
   type TeamMateTurnResult,
 } from '../teammate-collection/types.js';
+import { allocateConcreteName } from '../teammate-collection/name-allocator.js';
 import type { TeammateService } from '../teammate-service/index.js';
 import type { TeamStore } from '../team-collection/store.js';
 import type {
@@ -46,6 +49,7 @@ import type {
   TeamView,
 } from '../team-collection/types.js';
 import type { WorktreeManager } from '../worktree/manager.js';
+import { createTeamLeaderAgent } from './leader-agent.js';
 
 /**
  * The narrow dispatcher seam a {@link TeamService} needs for channel-bound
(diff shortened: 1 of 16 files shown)


packages/dreamux/src/service/team-service/index.ts
```diff
@@ -39,53 +59,189 @@ export interface TeamChannelContext {
...
+      }));
+    team = await deps.store.update(team, {
+      status: 'starting',
+      closedAt: null,
+      closeNote: null,
+      worktree: input.workspace.worktree,
+      intent: input.intent,
+      leaderName,
+    });
+    const { leader, result } = await service.teammateCollection.createTeamLeader(
```
YourWildDad: leader的构造器怎么在 teammateCollection 里？dispatcher 那边也是从 teammateCollection 里构造的吗？

YourWildDad: Confirmed asymmetric, and worth naming why.

- **Dispatcher agent**: built standalone in `dispatcher-service/agent.ts`
  (`createDispatcherAgent` → `new TeammateService(...)` directly), NOT via any
  collection — it persists to `status.json` (via `buildLaunch` + `DispatcherStore`),
  has a fixed name, no worktree, and is held directly as `DispatcherService.agent`.
- **Team leader**: built via `TeammateCollection.createTeamLeader`, because the
  leader is by construction *a teammate-shaped agent* — same entity / identity-store
  persistence / runtime / turn recording as a member, only `role: 'team_leader'` so
  it lands at the team root (see the method's doc comment). It reuses the
  collection's machinery (identity create, dispatcher-global name allocation, the
  entity cache so `leader(name)` returns the held instance on rebuild, completion
  registration); the dispatcher agent shares none of that.

So it's grounded, not arbitrary — but the honest cost is real: the members-scoped
collection also owns the leader factory. This is pre-existing #233 design, not
introduced by this PR (which only added the `options` arg to `createTeamLeader`).

Per the follow-up discussion: the right fix is a **unified `TeammateService`
factory** used by `DispatcherService` / `TeamService` / `TeammateCollection`, plus
extracting the shared *(conversational agent + members collection + scheduler +
member lifecycle)* host so the two services are symmetric. That's the next #233
phase, layered on top of this PR — not folded into it.

PR #245, opened 2026-06-26
refactor(service): extract a shared AgentHost (#233, step 2/3)

YourWildDad:
**Step 2 of 3** toward Dispatcher/Team responsibility symmetry. Targets the
integration branch `teamservice-launch-policy` (#240); the series merges to `next`
once, via #240. Builds on step 1 (#244, independent factory).

## Why

`DispatcherService` and `TeamService` already share a shape — a conversational
agent (`TeammateService`) + a members `TeammateCollection` + a scheduler (on the
agent, #243) + member/scheduler lifecycle — but each held and drove it separately.
That asymmetry is what makes the two services look unalike when they shouldn't.

## What (this step)

Extract `AgentHost` (`service/agent-host/index.ts`): holds the conversational
agent + the members collection, and owns the previously-duplicated surface
(`agent`/`mustAgent`, `scheduler`, narrow `teammates: TeammateOps`, concrete
`members`, `startScheduler`/`stopScheduler`/`deleteSchedulerStore`, `stopAll`).
Both services now CONTAIN an `AgentHost`:

- `DispatcherService` = host + `TeamCollection` + `ChannelSessions` + routing
  (agent set at construction).
- `TeamService` = host + record/bindings (`leader_`/`mustLeader()` move into the
  host; the leader is set via `host.setAgent(leader)` in `createNew`/`rebuild`,
  absorbing the nullable-agent pattern).

## Byte-identical behavior

- Dispatcher `shutdown()` sweeps MEMBERS only (`host.members.stopAll()`), agent
  stopped exactly once in `stop()` AFTER channels close — not `host.stopAll()`
  (which would stop the agent early + twice). `TeamService.stopAll()` =
  `host.stopAll()` keeps the old members-then-leader order.
- Scheduler start/stop ordering, `'miss'`/`'submit'`, boot eager-arm without
  starting the leader runtime, and the `dissolve` cron-store delete are unchanged.
- `teammates` stays narrowed to `TeammateOps`; completion-routing instances
  unchanged.

## Scope

Leader construction/caching is UNCHANGED — still built via the collection and
cached there; the host holds the same instance. Removing the leader from the
collection cache is **step 3**.

## Review

Two independent heterogeneous reviewers (codex + claude) — both **PASS**, no
blocking findings (teardown order, scheduler lifecycle, and `mustAgent()` timing
traced clean against pre-change source). One reviewer-independent fix was applied
during self-review: `shutdown()` → `host.members.stopAll()` (not `host.stopAll()`).

## Verification

- `tsc --noEmit` (src + tests), `eslint` — clean
- `vitest run` (`DREAMUX_SKIP_LIVE_CODEX=1`) — **525 passed | 4 skipped**
- files ≤ 700 lines; `none`-type Rush change file (pure refactor)

### DIFF
--- a/common/changes/@excitedjs/dreamux/agent-host-step2_2026-06-25-00-00.json
+++ b/common/changes/@excitedjs/dreamux/agent-host-step2_2026-06-25-00-00.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "",
+      "type": "none",
+      "packageName": "@excitedjs/dreamux"
+    }
+  ],
+  "packageName": "@excitedjs/dreamux",
+  "email": "053700@gmail.com"
+}
--- a/packages/dreamux/src/service/agent-host/index.ts
+++ b/packages/dreamux/src/service/agent-host/index.ts
@@ -1,0 +1,77 @@
+import type { SchedulerService } from '../scheduler/service.js';
+import {
+  TeammateCollection,
+  type TeammateOps,
+} from '../teammate-collection/index.js';
+import type { TeammateService } from '../teammate-service/index.js';
+
+export interface AgentHostOptions {
+  members: TeammateCollection;
+  agent?: TeammateService;
+  agentDescription: string;
+}
+
+/**
+ * Shared host for a conversational agent plus its owned member collection.
+ * DispatcherService and TeamService keep their domain-specific policy and
+ * records, while this owns the duplicated member/scheduler lifecycle surface.
+ */
+export class AgentHost {
+  private readonly members_: TeammateCollection;
+  private agent_: TeammateService | null;
+  private readonly agentDescription: string;
+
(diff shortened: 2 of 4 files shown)


YourWildDad:
Closing without merging. Per the follow-up design review: (1) composition via a shared AgentHost forces delegation boilerplate and was the wrong factoring — the Dispatcher/Team symmetry already comes from the shared primitives (TeammateService + TeammateCollection), not an aggregate wrapper; (2) more importantly, the scheduler does NOT belong on TeammateService at all (a generic runtime entity that members also are) — it is a conversational-context concern owned by the container. Replanned: move the scheduler back to DispatcherService/TeamService (so 'only dispatcher + team leader have cron' is structural, no per-instance capability policy), then build+hold the leader directly via the factory out of the collection cache. AgentHost is dropped.

PR #271, opened 2026-06-30
Refine AgentRuntime lifecycle contract

YourWildDad:
## Summary

Refines the AgentRuntime provider contract around the lifecycle boundaries needed before adding more external runtimes.

- Carry turn settlement results on the settled turn instead of deriving routed completion text from runtime-level `getLast()` state.
- Move provider-facing thread/session concepts to opaque checkpoints while keeping public `thread_id` / `session_id` fields as compatibility projections.
- Make runtime state sinks instance-scoped, split model turn/control notice/completion delivery paths, and fail loudly for legacy teammate-completion delivery through `systemInput`.
- Reverse dispatcher prompt injection ownership: Dreamux now supplies both `replace` and `append` prompt forms and providers choose their native mapping.
- Validate external runtime handles against the new provider-facing contract while treating legacy projections as optional.
- Add the AgentRuntime lifecycle proposal and Rush change files for affected packages.

## Validation

- `CI=true node common/scripts/install-run-rush.js test --to @excitedjs/dreamux-types --to @excitedjs/dreamux --to @excitedjs/agent-runtime-codex --to @excitedjs/agent-runtime-claude-code --ignore-hooks`
- `node common/scripts/install-run-rush.js build --to @excitedjs/dreamux --to @excitedjs/dreamux-types --to @excitedjs/agent-runtime-codex --to @excitedjs/agent-runtime-claude-code`
- `node common/scripts/install-run-rush.js build --to @excitedjs/agent-runtime-claude-code --ignore-hooks`
- `node common/scripts/install-run-rush.js test --to @excitedjs/agent-runtime-claude-code --ignore-hooks` (passes with expected stderr logging warning)
- `pnpm --filter @excitedjs/dreamux run typecheck:tests`
- Package typechecks for `@excitedjs/dreamux`, `@excitedjs/dreamux-types`, `@excitedjs/agent-runtime-codex`, and `@excitedjs/agent-runtime-claude-code`
- Focused runtime/provider/router/scheduler tests around the changed contract
- `.agents/scripts/check.sh`
- `node common/scripts/install-run-rush.js change --verify --target-branch origin/next --no-fetch`
- `git diff --check`

## Notes

A non-CI Codex live gate for issue #63 currently times out in this environment because the live model does not begin the expected command execution. The same narrow live gate also fails on a clean `origin/next` baseline under the same Codex version, so current evidence does not point to this PR's diff. CI skips that live model gate as designed.

PR #277, opened 2026-07-02
docs: add OpenCode agent runtime proposal

YourWildDad:
## Summary
- Add an active `.agents` proposal for OpenCode as a Dreamux `AgentRuntimeProvider`.
- Index the proposal from the KB root.
- Align the proposal with the simplified runtime capability contract from #271.

## Validation
- `.agents/scripts/check.sh`
- `git diff --cached --check`
- gitleaks pre-commit hook

### DIFF
--- a/.agents/proposals/opencode-agent-runtime.md
+++ b/.agents/proposals/opencode-agent-runtime.md
@@ -1,0 +1,424 @@
+# Proposal: OpenCode Agent Runtime provider
+
+- **Status:** Active proposal (draft for review)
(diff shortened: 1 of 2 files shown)


YourWildDad:
Closing as obsolete. This proposal targets the pre-#350 Agent Runtime interface and the previous KB layout; the current contract is centered on `start` / `submit` / `stop` plus the current activity capability. If OpenCode remains a product requirement, it should be proposed again against the current provider seam.

PR #297, opened 2026-07-15
fix(feishu): route topic groups to collaboration spaces

YourWildDad:
## Root cause

The original Feishu path lost `message.thread_id` during inbound normalization and reconstructed every channel route as a chat-level group/P2P target. Dispatcher therefore never received the provider-owned `topic_group` container or stable topic target required by collaboration-space provisioning.

After topic projection was added, a compatibility gap remained: a verified topic message used `thread_id` as its exact target key, while pre-existing Team bindings for that chat still used the enclosing `group/chat_id` target. Core only looked up the exact target, so a topic-mode group with no collaboration-space binding bypassed its active group Team and fell back to Dispatcher.

## Feishu topic projection and routing

- Preserve `thread_id`, `root_id`, and `parent_id` from the raw Feishu event.
- Resolve the enclosing chat through `im.v1.chat.get`.
- Project a topic route only for `chat_type=group`, a non-empty `thread_id`, and verified `chat_mode=topic`.
- Emit `container_type=topic_group` with `container_key=chat_id`.
- Emit a bindable `target_type=topic` with `target_key=thread_id`.
- Declare the enclosing `group/chat_id` target through the provider-neutral `ChannelTarget.binding_fallbacks` capability.
- Route in this order: exact target binding, accepted collaboration-space provision/reclaim of the exact target, existing exact claim, existing provider-declared fallback binding, Dispatcher.
- Never provision collaboration targets from a fallback and never derive a fallback from provider metadata in core.
- Stop fallback traversal when a more-specific active binding exists but its Team is unavailable.

This preserves an existing group-bound Team for a non-collaboration topic group, while a bound collaboration space still provisions/reuses one Team per Feishu topic and an explicit topic binding remains highest priority.

## Model-visible inbound identity

Every non-empty Feishu `thread_id` is now included in the provider-owned display attrs and therefore rendered in the model-visible `<channel ... thread_id="...">` envelope. This is independent of routing classification: an ordinary group thread may expose the raw identifier while remaining a group target with no collaboration container.

## Ordinary-group fail-safe

A normal group thread remains a group target even when it carries `thread_id`, `root_id`, or `parent_id`. Missing chat lookup capability, permission failures, network errors, and missing or unknown `chat_mode` log a safe structured warning and do not emit a collaboration container. Failed lookups are not cached, so a later accepted inbound retries. Access-gated messages do not call the chat API or populate the target ledger.

A topic-mode group with neither an exact/group binding nor an accepted collaboration-space binding retains the established Dispatcher fallback.

## Topic egress authorization

The Feishu session records the exact normalized target for every accepted inbound `message_id`. TeamLeader authorization resolves through that ledger, rejects conflicting chat/thread selectors, and confirms exact message-to-topic ownership before any group fallback binding can authorize egress. A group-bound TeamLeader can therefore reply safely to an observed topic message through Feishu's source-message reply API, while standalone `thread_id`, empty/unknown message ids, wrong chats, and cross-topic selectors remain denied.

## Production-only Feishu package surface

The legacy root-exported `createFakeFeishuBot` factory and `FakeFeishuBot` type were pure test fixtures, so they have been removed from production `src/` and the published package root. Feishu Channel tests and Dreamux host/provider integration tests now own package-local `FeishuBot` doubles behind the existing `botFactory` seam.

This deliberate public API removal is recorded as a `major` Rush change for `@excitedjs/feishu-channel`. A `prepack` clean build prevents orphaned `dist/fake-bot.*` files from entering a tarball, and runtime/type/source guards prevent the test API from returning.

## Validation

- `cd packages/channel/feishu-channel && node ../../../common/scripts/install-run-rushx.js test tests/public-api.test.ts tests/import-boundary.test.ts tests/feishu-topic-routing.test.ts tests/feishu-inbound-submit-cleanup.test.ts tests/feishu-pairing-card.test.ts` (22/22)
- `cd packages/dreamux && node ../../common/scripts/install-run-rushx.js test tests/channel-provider.test.ts tests/channel-service-feishu-topic-auth.test.ts tests/dispatcher-collaboration-space.test.ts tests/e2e.test.ts` (34/34)
- `node common/scripts/install-run-rush.js test` (passed, including the installed Codex live gate)
- `node common/scripts/install-run-rush.js typecheck:tests`
- `node common/scripts/install-run-rush.js lint`
- `node common/scripts/install-run-rush.js smoke-built-cli`
- `cd packages/channel/feishu-channel && npm pack --dry-run --json` (71 files; an injected stale fake artifact was removed by `prepack`, with no fake bot file or symbol in the pack list)
- `.agents/scripts/check.sh`
- `node common/scripts/install-run-rush.js change --verify --target-branch origin/next --no-fetch`
- `git diff --check`

## Residual limitations

- Topic classification requires the bot to be a chat member and have a group-information read permission accepted by the chat-get API, such as `im:chat:readonly`.
- If verification fails, that inbound safely follows ordinary group routing; a later inbound is required to retry classification and provisioning.
- Topic identity and topic reply authorization depend on Feishu providing `thread_id`; topic-group main-area messages or events without it intentionally do not provision.
- The message-target ledger is session-local, so pre-restart messages are not authorized until a new inbound is observed.
- Feishu does not expose topic-created/topic-closed lifecycle through this change; Teams provision on first inbound and are not automatically dissolved when an external topic closes.
- The in-memory message ledger remains unbounded, matching the previous chat ledger behavior.


packages/channel/feishu-channel/tests/helpers/fake-feishu-bot.ts
```diff
@@ -0,0 +1,195 @@
...
+  }>;
+  readonly reactions: Array<{
+    messageId: string;
+    emoji: string;
+    reactionId: string;
+  }>;
+  readonly removedReactions: Array<{
+    messageId: string;
+    reactionId: string;
+  }>;
```
YourWildDad, line 41: 为什么要在 src 里保留fake bot？这个不应该只在test 代码里吗？

YourWildDad:
thread_id 作为入站信息，也序列化到 channel 消息体里。现在有吗？没有的话也加上

packages/channel/feishu-channel/tests/helpers/fake-feishu-bot.ts
YourWildDad: Confirmed that the fake implementation is only used by tests in this repository. However, this PR did not introduce that public API: `origin/next` already root-exports both `createFakeFeishuBot` and the `FakeFeishuBot` type, and git history shows they were introduced with the published provider in `feebea6`. This PR only extracted the pre-existing implementation from `bot.ts` because the topic work would otherwise exceed the package's source-line lint limit. Removing the root exports here would be an undeclared breaking change for `@excitedjs/feishu-channel`, so I kept them source-compatible and documented their removal as a separately versioned breaking migration.

YourWildDad:
Addressed the `thread_id` request from https://github.com/excitedjs/dreamux/pull/297#issuecomment-4977227921 in `1323ec4`: every non-empty Feishu inbound `thread_id` is now included in the provider-owned attrs rendered into the model-visible `<channel>` envelope. Missing/empty values are omitted, and an ordinary-group thread may expose the raw value without being promoted to a collaboration target.

The same commit also fixes the live group-binding compatibility regression through a neutral, provider-declared binding fallback: exact topic bindings remain first, accepted collaboration spaces still provision/reuse one Team per topic, non-collaboration topic groups may reuse an existing `group/chat_id` Team binding, and groups with neither route still fall back to Dispatcher. TeamLeader fallback egress requires a non-empty observed `message_id` whose exact topic ownership the provider confirms. Targeted coverage is 40/40 in Dreamux plus 5/5 in Feishu Channel; the full Rush test gate also passes.

packages/channel/feishu-channel/tests/helpers/fake-feishu-bot.ts
YourWildDad: 不要顾虑 breaking change，这种垃圾代码不管什么时候引入的，发现了就要给我干掉

YourWildDad: Addressed in `a7c2e43`. I rechecked the usage rather than preserving the API by inertia: the fake bot had no production caller, only test consumers. The implementation now lives under Feishu Channel tests, Dreamux owns a narrower package-local test double, and `@excitedjs/feishu-channel` no longer compiles or root-exports `createFakeFeishuBot` / `FakeFeishuBot`. The deliberate API removal has a `major` Rush change file. Runtime/type/source guards pin the production-only surface, and `prepack` now clean-builds so stale `dist/fake-bot.*` cannot leak into a tarball. Targeted tests pass (22/22 Feishu Channel, 34/34 Dreamux), as do full Rush build, typecheck, test-aware typecheck, lint, test (including the live Codex gate), smoke CLI, KB, and change validation.

PR #304, opened 2026-07-22
feat(feishu): preserve rich inbound messages

YourWildDad:
## What changed

- add a transport-owned `im.message.get` seam with a narrow provider-local projection
- preserve rich post Markdown/code plus inline image, file, audio, and video resources
- resolve interactive cards through structured/default reads with bounded visible-content parsing
- represent merged-forward messages lazily with fact-only `message_id` and message-type context; the Channel does not prescribe a tool or command
- surface reply/quote ancestry with `parent_message_id` and a validated `parent_message_type`; parent bodies and children are not copied into the turn
- render downloaded attachments as `<attachment path="..." />` and unavailable attachments as `<attachment status="not_downloaded" key="..." />`, without duplicate metadata or retrieval instructions
- resolve accepted unnamed human senders through the transport-owned contact seam with a two-second bound, positive caching, same-user single-flight, and retry after every failed or nonzero response
- project stickers, shared users/chats, and unknown message types honestly instead of silently losing them
- fence enrichment to the active Feishu session and enforce total time, byte, resource-count, parser, and rendered-size budgets
- close truncated Markdown code fences before truncation markers and the Channel reminder
- make the channel reply acknowledgement conditional: direct answers do not need a separate ACK
- document the final Feishu inbound contract and include Rush change files for affected packages

## Why

Dreamux previously reduced several Feishu message types to generic placeholders or silently dropped visible text and attachments. The model could not tell that rich-post images, card fields, code blocks, merged-forward history, or quote ancestry existed. Attachment metadata also repeated the same fact, while an aggressive contact-permission circuit could suppress all later sender-name lookups after one failure.

Merged-forward and quoted history are intentionally not copied into every Agent turn. The Channel supplies only stable identity and type facts. Tool choice and retrieval policy remain outside the Channel prompt. Sender-name lookup is best-effort: a failure affects only the current message, and the next positive-cache miss retries.

Feishu SDK access remains in `@excitedjs/feishu-transport` and model-facing policy remains in `@excitedjs/feishu-channel`. This adds no Dreamux core/runtime capability, persisted identity store, message cache, or merged-forward aggregate.

## Validation

- `rush build`
- `rush typecheck`
- `rush typecheck:tests`
- `rush lint`
- `rush test` including the issue #63 live Codex gate
- `rush rebuild --to @excitedjs/dreamux`
- `rush smoke-built-cli`
- `.agents/scripts/check.sh`
- `git diff --check`
- commit-hook and focused diff gitleaks scans
- independent final reviews with Codex High, Trae Claude, and Claude Seed


packages/channel/feishu-channel/src/feishu-inbound-work.ts
```diff
@@ -0,0 +1,195 @@
...
+  const remaining = Math.max(0, effectiveDeadlineAt - Date.now());
+  const endsAtMessageDeadline = effectiveDeadlineAt === work.deadlineAt;
+  const timeoutError = ():
+    | FeishuSessionRevokedError
+    | FeishuEnrichmentDeadlineError
+    | FeishuResourceTimeoutError =>
+    endsAtMessageDeadline
+      ? deadlineError(work)
+      : new FeishuResourceTimeoutError();
+  if (remaining === 0) throw timeoutError();
```
YourWildDad: **P2 — Run immediate-timeout cleanup before returning.** When an operation-local deadline is already expired, this branch throws before invoking `onOperationTimeout`. In the production attachment path, `fetchMessageResource` has already returned `response.stream`, and the following `runFeishuInboundWork` call passes `onOperationTimeout` specifically to destroy that stream (`feishu-message.ts:452-458`). If the deadline expires between those two steps, the acquired stream remains open. I reproduced this on the current head with an expired local deadline: `FeishuResourceTimeoutError`, `cleanupCalls=0`, `stream.destroyed=false`. Please invoke the same best-effort timeout callback in this immediate-expiration path and add a regression test.

YourWildDad (commented):
One remaining resource-cleanup edge case on the current head.

packages/channel/feishu-channel/src/feishu-inbound-work.ts
YourWildDad: When an operation-local deadline is already expired, this early throw bypasses `onOperationTimeout`. In `resolveAttachment`, the resource response (and its stream) has already been obtained before the stream reader is wrapped. If the fetch resolves at the deadline boundary and this call observes `remaining === 0`, it returns `FeishuResourceTimeoutError` without starting the reader or destroying `response.stream`, leaving the response stream/socket open. A probe against this head returns `{ errorName: "FeishuResourceTimeoutError", streamDestroyed: false }`. Please run the same best-effort timeout cleanup before this throw (or destroy explicitly in the caller) and add a regression case where the fetch returns as the resource deadline expires.

packages/channel/feishu-channel/src/feishu-message.ts
```diff
@@ -105,11 +137,33 @@ export async function formatFeishuMessageForRuntime(
...
+  const ancestry = renderReplyAncestry(event);
+
+  const renderedPrefix = `${body}${ancestry}${fallback}${attachmentBlock}${attachmentOmission}`;
+  let groupBotsRendered = groupBots !== '';
+  let renderedBody: string;
+  if (!isRichMessage(event.messageType)) {
+    renderedBody = `${renderedPrefix}${groupBots}`;
+  } else if (
+    groupBots !== '' &&
+    groupBots.length + RICH_BODY_TRUNCATION_MARKER.length <= MAX_RICH_BODY_CHARS
```
YourWildDad: The marker is reserved even when no truncation is needed, so a baseline that would fit with the complete prefix can still be dropped. I reproduced this with a 159,960-character `groupBots` block and a one-character rich prefix: the combined 159,961 characters are below `MAX_RICH_BODY_CHARS`, but this condition is false, `groupBotsRendered` becomes false, and the baseline remains pending. Please first accept `renderedPrefix.length + groupBots.length <= MAX_RICH_BODY_CHARS` without reserving a marker; only reserve `RICH_BODY_TRUNCATION_MARKER` when the prefix actually needs truncation. A boundary test for this fit-without-truncation case would lock it.

YourWildDad (commented):
One confirmed edge case remains in the rich-body baseline reservation.

packages/channel/feishu-channel/src/feishu-inbound-work.ts
YourWildDad: This pre-start timeout bypasses `onOperationTimeout`. In the resource path the fetch can return a response stream just before `resourceDeadline`, then the subsequent stream wrapper reaches this branch; the read operation is correctly not started, but the already-owned response stream is never destroyed. A direct probe with `deadlineAt = Date.now()` produces `FeishuResourceTimeoutError`, `operationCalls = 0`, and `cleanupCalls = 0`. Please run the best-effort cleanup before this throw (and account for the preceding `assertEnrichmentActive()` preflight as well), or have the caller destroy the acquired stream on every pre-start failure. The test for the already-expired local deadline should assert the cleanup callback too.

YourWildDad (commented):
A pre-start timeout path still skips the new stream cleanup hook.

packages/channel/feishu-channel/src/feishu-inbound-work.ts
YourWildDad: Addressed in 6b61f08. Stream cleanup now stays with the attachment owner: any rejection from the bounded read destroys the already-acquired response stream, including pre-start deadline/session failures. The production-path regression test advances across the operation deadline after fetch and asserts both timeout classification and stream destruction.

YourWildDad: Addressed in 6b61f08. Stream cleanup now stays with the attachment owner: any rejection from the bounded read destroys the already-acquired response stream, including pre-start deadline/session failures. The production-path regression test advances across the operation deadline after fetch and asserts both timeout classification and stream destruction.

YourWildDad: Addressed in 6b61f08. Stream cleanup now stays with the attachment owner: any rejection from the bounded read destroys the already-acquired response stream, including pre-start deadline/session failures. The production-path regression test advances across the operation deadline after fetch and asserts both timeout classification and stream destruction.

packages/channel/feishu-channel/src/feishu-message.ts
YourWildDad: Addressed in 6b61f08. Rich rendering now accepts the complete prefix plus group-bot block whenever it fits within 160,000 characters; truncation-marker space is reserved only when prefix truncation is actually required. The new boundary test fills the body to exactly 160,000 characters and asserts the baseline is rendered without a marker.

packages/channel/feishu-channel/src/feishu-message.ts
```diff
@@ -187,27 +259,145 @@ function renderTextWithMentions(text: string, mentions: Mention[]): string {
...
+    if (required > maxChars) {
+      return `${output}${RICH_BODY_TRUNCATION_MARKER}${closingTags(openTags)}`;
+    }
+    output += tag;
+    openTags.splice(0, openTags.length, ...nextOpenTags);
+    cursor = match.index + tag.length;
+  }
+
+  appendText(value.slice(cursor));
+  return `${output}${RICH_BODY_TRUNCATION_MARKER}${closingTags(openTags)}`;
```
YourWildDad: **P2 — Keep Markdown code fences balanced when truncating rich bodies.** `truncateEscapedRichBody()` only tracks channel-owned XML tags. With a post body shaped as `````ts\n${'x'.repeat(170_000)}\n``` ``, the current formatter returns a 160,000-character body containing exactly one ``` fence; `deliverAcceptedMessage()` then appends the truncation note and `<channel-reminder>` inside that open code block. This breaks the promised code fidelity and can make the standing reminder look like quoted code. Please make truncation fence-aware (reserving room for the active fence closer) and add a regression test for an over-limit fenced code block.

YourWildDad (commented):
One actionable rich-content truncation edge case remains.

packages/channel/feishu-channel/src/feishu-markdown-fence.ts
```diff
@@ -0,0 +1,37 @@
...
+  const leadingNewline = value.endsWith('\n') ? '' : '\n';
+  return `${leadingNewline}${fence.marker.repeat(fence.length)}\n`;
+}
+
+function activeMarkdownFence(value: string): MarkdownFence | null {
+  let active: MarkdownFence | null = null;
+  for (const rawLine of value.split('\n')) {
+    const line = rawLine.endsWith('\r') ? rawLine.slice(0, -1) : rawLine;
+    if (active === null) {
+      const opening = /^ {0,3}(`{3,}|~{3,})/.exec(line);
```
YourWildDad: **P2 — Validate backtick fence info strings before synthesizing a closer.** CommonMark does not treat a backtick-fence line as an opener when its info string contains another backtick, but this regex does. With an over-limit post beginning with such a line, the formatter appends a bare backtick fence before the truncation marker; `marked` treats the original line as plain text and that synthesized fence as a new opener, so both the marker and `<channel-reminder>` remain inside a code block. Please reject these invalid backtick openings (tilde fences are unaffected) and add a regression fixture for this case.

YourWildDad: **P2 — Reject invalid backtick-fence info strings before synthesizing a closer.** CommonMark does not treat a backtick fence as open when its info string contains another backtick, but this regex accepts every suffix. I reproduced this on the current head with a rich post whose text is three backticks plus `foo` plus one backtick plus `bar`, a newline, and 170,000 `x` characters: the formatter appends a three-backtick line immediately before the truncation marker. Because the original line is not a valid opener, that synthesized line becomes the actual opening fence, so the marker and the later channel reminder can be interpreted as code. Please validate the full opener suffix for backtick fences and add this adversarial production-path regression.

packages/channel/feishu-channel/src/feishu-message.ts
```diff
@@ -284,27 +553,7 @@ function renderAttachment(
...
-    `--type ${attachment.type}`,
-    `--output ./${outputName}`,
-  ].join(' ');
-  return [
-    `\n\n<attachment ${attrText}>`,
-    'Use lark-cli to fetch it if needed:',
-    escapeXmlText(command),
-    '</attachment>',
-  ].join('\n');
+  return `\n\n<attachment ${attrText} />`;
```
YourWildDad: **P2 — Update or supersede the accepted attachment-body contract.** This fact-only self-closing projection is a clean boundary, but `.agents/decisions/feishu-inbound-attachments.md` still says every failed resource must include a lark-cli fallback direction and shows a non-self-closing attachment body. That file is an Accepted decision and current architecture source, while this PR only updates the package guide and archived proposals. Please update the accepted decision in this PR so future work is not instructed to restore the command that this change deliberately removes.

PR #308, opened 2026-07-30
feat(provider): manage npm providers in a local plugin store

YourWildDad:
## Summary

- Materialize external `npm:` provider references into Dreamux's private `~/.dreamux/plugins` store instead of resolving ambient or globally installed packages.
- Validate the host-owned config envelope before npm access or provider code execution, then pin one exact immutable generation across runtime and channel refs for the entire load.
- Treat `selected_version` as last-known-good: first use and background checks stage `candidate_version`; only a complete provider factory, contract, and `readConfig` validation commits the candidate.
- Keep background updates non-disruptive. The updater checks at most every four hours, stages candidates without importing or selecting them, records settled errors, and closes its timer/npm work cleanly.
- Give doctor and dry-run flows explicit installed-only inspection results without fabricating runnable providers or partially resolved `DreamuxConfig` values.
- Keep onboard on one non-serializable plugin session from provider collection through final strict validation, and retain canonical uninstall safety with separate physical-removal and logical-alias ledger entries.

## Architecture

The registry layer owns immutable generations, versioned metadata, generation-local ESM bridges, candidate/selected transitions, and one abortable updater. Config owns host validation, provider declaration inspection, provider `readConfig`, and the strict-load commit gate. The generic runtime/channel loaders consume a generation-pinned importer and never implement a second npm-aware loading loop.

Candidate-backed strict loads may fall back once to a fresh-registry selected-only attempt, but only when a structured failure belongs to the candidate provider. Candidate rejection attempts every prepared package; cleanup failures retain the original validation error and prevent fallback.

## Upgrade impact

`BREAKING:` External `npm:` refs no longer use ambient/global Node resolution. First startup, onboard, or daemon install for each package requires npm registry access and creates Dreamux-owned state under `~/.dreamux/plugins`.

`Rebuild:` Restart Dreamux after an update check so the next strict load can validate and promote the staged candidate. Published generations are immutable and are not garbage-collected by this change.

## Validation

- Resident three-seat holistic exact-head review: Architecture, Lifecycle/Correctness, and Complexity/Reuse all approved `441b967976e7509892e49ec17b4b536e2bf243cb`.
- Focused provider/config/doctor/onboard/daemon/uninstall suites: 6 files, 188 passed.
- TeamLeader rerun of provider-store and global-config suites: 2 files, 84 passed.
- `rush rebuild --from @excitedjs/dreamux`
- `rush lint --from @excitedjs/dreamux`
- `rush build --from @excitedjs/dreamux`
- `rush typecheck --from @excitedjs/dreamux`
- `rush typecheck:tests --from @excitedjs/dreamux`
- `rush change --verify --no-fetch`
- `.agents/scripts/check.sh`
- `git diff --check`
- Built CLI help smoke

Relative to the pre-hardening `ff24482` snapshot, production TypeScript is +616 lines and tests are -618 lines (combined -2). The source increase is the explicit config-owned inspection/readConfig boundary; the previous CLI/config/onboard glue and implementation-coupled tests were removed.

A skip-live full test run reached the existing `team-scheduler` same-name identity-collision flake twice; that file passed 22/22 in isolation. The final provider/config/onboard/doctor/daemon/uninstall suites pass. Live Codex app-server compatibility was not re-verified in this environment because `codex_apps` MCP startup timed out.


YourWildDad:
Closing as obsolete in favor of the narrower #302 direction. This branch adds generations, activation state, leases, background updating, and GC, while #302 explicitly excludes those mechanisms and asks for a minimal managed npm root/installer on the current loader seam.

Issue #318, opened 2026-08-08
Align workflow script API and limits with Claude Code's built-in Workflow (ultracode) orchestrator


ryanxiang7:
## Background

Dreamux's `workflow_run` script API and Claude Code's built-in `Workflow` tool (the "ultracode" orchestrator) share the same programming model: an inline JS module with a literal `meta` block, plus `agent()` / `parallel()` / `pipeline()` / `phase()` / `log()` and JSON-Schema structured output. Today the two dialects have drifted apart. Since the same person often writes both kinds of scripts, the divergence adds friction: a script written for one side does not run on the other, and the weaker side lacks several load-bearing features.

**Goal: make the Dreamux workflow script dialect a superset-compatible match of the ultracode dialect.** Small, additive parameter differences (e.g. Dreamux's `intent` / `identity` agent options) are fine and should stay. Everything else — script shape, callback signatures, runtime semantics, and scale limits — should align.

### 1. Script entry shape

| | Dreamux | ultracode |
|---|---|---|
| Entry | `export default async function run() {...}` required | top-level script body, `await` directly |
| `meta.phases` | `string[]` (`['review', 'summary']`) | `{title, detail?, model?}[]` |
| `meta.whenToUse` | not supported | supported, shown in workflow list |

### 2. `pipeline()` stage callback signature

- ultracode: every stage receives `(prevResult, originalItem, index)` — later stages can label work without threading context through stage 1's return value.
- Dreamux: stages only receive `(prevResult)`; users must manually pass context down through each stage's return value.

### 3. Structured output (`schema`) semantics

- ultracode: the subagent is forced onto a `StructuredOutput` tool call; validation happens at the tool-call layer and **the model retries on schema mismatch**. Subagents are also system-instructed that "your final text IS the return value, not a human-facing message", so prompts can ask for raw data directly.
- Dreamux: if the runtime's final text is not valid JSON, `agent()` silently returns `null` — no retry, no system-level output contract. Prompts must be written defensively ("return JSON only").

### 4. `agent()` options

- ultracode-only: `model` (per-agent model override), `effort` (reasoning effort tier), `isolation: 'worktree'` (fresh git worktree per agent, auto-removed if unchanged — enables safe parallel file edits).
- Dreamux-only: `intent`, `identity` — these are fine to keep as Dreamux extensions.
- Shared: `label`, `phase`, `schema`, `agentType`.

### 5. Missing orchestration features in Dreamux

[...]

PR #319, opened 2026-08-08
skill(workflow): bundled code-review recipe reference

ryanxiang7:
## What

Adds two bundled references to the shared `workflow` skill, linked from a new References routing table in `SKILL.md` (same convention as the `dreamux-maintenance` skill):

- `references/code-review.md` — a recipe for carrying a full code review as a single `workflow_run`: eligibility gate → context (guidance-file discovery + change summary) → six independent finder lenses with explicit coverage accounting → three-vote 0-100 confidence scoring through `pipeline` → one synthesized report. Includes a complete runnable script in the ultracode top-level dialect.
- `references/orchestration-patterns.md` — the craft guide: prompt discipline under the injected output contract, pipeline-versus-parallel decision rules with the barrier smell test, and reusable quality patterns (adversarial verify, perspective-diverse verify, judge panel, loop-until-dry, multi-modal sweep, completeness critic, no silent caps).

## Why

- Review work is the most workflow-shaped task a Team gets: a known coordination graph, heavy fan-out, one terminal result. A bundled recipe saves every TeamLeader from re-deriving the stage design, false-positive list, and scoring rubric.
- Both references are written against the current dialect from #318/#320: top-level entry, object-form `meta.phases` + `whenToUse`, three-argument pipeline stages, runtime-native structured output (failed turn → `null`; schema contract breach → rejection; direct awaits fail the run while helpers contain per-item nulls), 1000 lifetime TeamMates, 4096 helper inputs, default concurrency 16.
- The code-review script accounts for finder coverage explicitly: lens results stay index-aligned, a fully failed find phase returns an error, and partial failure is surfaced as `coverage: {complete: false, failedLenses}` in the result and in the report text — no silent `.filter(Boolean)` coverage loss.

## Notes

- Documentation-only bundled skill content; no runtime code touched. Skill injection ships the whole skill root, so `references/` is picked up without code changes.
- Rush change file included (patch: bundled skill content ships in the package).
- Branch is merged up to current `next` (includes #320).

Refs #318


packages/dreamux/skills/shared/workflow/references/code-review.md
```diff
@@ -0,0 +1,279 @@
+# Code Review Workflow
+
+`../SKILL.md` owns the run tools and the script API. This reference owns one
+recipe: carrying a full code review as a single `workflow_run`, modeled on the
+staged flow of the official Claude Code `code-review` plugin — eligibility gate,
```
YourWildDad: 不要提出处，codex 也可能使用这个技能，这个描述对于codex 有半毛钱帮助吗？

packages/dreamux/skills/shared/workflow/references/code-review.md
```diff
@@ -0,0 +1,279 @@
...
+context collection, independent multi-lens finders, per-finding confidence
+scoring, and one synthesized report.
+
+Use it when the operator asks to review a pull request or a git range and wants
+one terminal result rather than an interactive back-and-forth. For interactive
+review — where the next instruction depends on reading the previous finding —
+use `spawn` and `send` instead.
```
YourWildDad: 为什么要假设去 review github 的 PR？
我可能去review gitlab 的mr 啊，也可能是本地变更啊

packages/dreamux/skills/shared/workflow/references/code-review.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Stage design
+
+1. **Gate** — one agent checks eligibility and the run exits early when review
+   is not needed: the change is closed or a draft, is automated or trivially
+   safe, or already carries a review from this Team.
+2. **Context** — two agents in parallel: one lists the file paths (not
+   contents) of the root `CLAUDE.md` and any `CLAUDE.md` in directories the
```
YourWildDad: 不要只指定 CLAUDE.md，AGENTS.md 也包含进来

packages/dreamux/skills/shared/workflow/references/code-review.md
```diff
@@ -0,0 +1,279 @@
...
+one terminal result rather than an interactive back-and-forth. For interactive
+review — where the next instruction depends on reading the previous finding —
+use `spawn` and `send` instead.
+
+## Stage design
+
+1. **Gate** — one agent checks eligibility and the run exits early when review
+   is not needed: the change is closed or a draft, is automated or trivially
+   safe, or already carries a review from this Team.
+2. **Context** — two agents in parallel: one lists the file paths (not
```
YourWildDad: 不要用 agent，应该改成teammate，下面所有提法都是一样的

packages/dreamux/skills/shared/workflow/references/code-review.md
ryanxiang7: Fixed in 20a0072 — dropped the provenance mention; the recipe is now described on its own merits, runtime-neutral.

packages/dreamux/skills/shared/workflow/references/code-review.md
ryanxiang7: Fixed in 20a0072 — target is now forge-neutral: a single `args.target` takes a PR/MR URL or number, a local git range, or a working-tree description; prose, prompts, and the invocation examples updated accordingly (incl. gh/glab for posting).

packages/dreamux/skills/shared/workflow/references/code-review.md
ryanxiang7: Fixed in 20a0072 — guidance discovery now covers CLAUDE.md and AGENTS.md ('agent guidance files') across the context stage, finder lens, rubric, and false-positive list.

packages/dreamux/skills/shared/workflow/references/code-review.md
ryanxiang7: Fixed in 20a0072 — prose now says TeamMate throughout (stage design, budget section renamed to TeamMate budget, 200-TeamMate cap); `agent()` remains only where it names the script API call.

ryanxiang7:
Added a second reference in 00a72cd: `references/orchestration-patterns.md` — prompt discipline for workflow TeamMates, pipeline-vs-parallel decision rules (barrier smell test), and the reusable quality patterns (adversarial verify, perspective-diverse verify, judge panel, loop-until-dry, multi-modal sweep, completeness critic, no silent caps), written runtime-neutral with snippets in the current script dialect. SKILL.md References table updated.

ryanxiang7:
Updated both references for the ultracode-parity dialect from #320 (commit c7656d7): the code-review script now uses the top-level entry, object-form phases + whenToUse, and 3-arg pipeline stages; JSON-only prompt boilerplate is dropped in favor of the injected output contract and runtime-native structured output; failure semantics (null for failed turns, rejection for schema contract breaches, direct-await fails the run) and limits (1000 lifetime TeamMates, 4096 helper inputs, default concurrency 16) now match the new runtime. Merge order note: this PR now documents the #320 dialect, so #320 should land first.

YourWildDad (changes requested):
## Review — exact HEAD `c7656d740dc22e13e84e2c6cbd35e53b1a412881`

**CHANGES_REQUESTED**

I reviewed the complete PR against current `next` (`2f24a4f039283a857a7ea9246329194233f71c7f`, including #320), validated the bundled skill shape, normalized the full code-review script with the production normalizer, and ran it through the real child workflow runner with synthetic Gate → Context → Find → Verify → Report results. The happy path completes correctly, but the following issues remain.

### P1 — Finder failures can be reported as a successful clean review

`packages/dreamux/skills/shared/workflow/references/code-review.md:218-240`

Every finder is a schema-constrained call inside `parallel()`. Ordinary failed turns and contained schema-contract rejections become `null`; the script immediately removes them with `.filter(Boolean)`. If all six finders fail, `found` becomes empty and the run returns:

> `No issues found. Checked for bugs, security, history, and guidance-file compliance.`

That is the opposite of what happened: no lens completed. Partial finder loss is also silent, so the same clean-review sentence can be emitted after only a subset of the documented coverage ran. For a bundled code-review recipe this is load-bearing, not a cosmetic edge case.

Keep the positional finder results, associate them with lens names, and fail or return an explicit incomplete-review result when required lenses are missing. Only return “No issues found” after all required finder lenses settled successfully. Apply the same principle to load-bearing context collection if the recipe claims guidance/history coverage.

### P1 — Required Rush declaration is missing and CI is red

The PR changes the published `@excitedjs/dreamux` bundled skill tree but has no file under `common/changes/@excitedjs/dreamux/`. The exact-head `rush change — declaration check` fails with:

> `@excitedjs/dreamux` ... require change descriptions

This is a package-shipped model-facing capability change. Add the appropriate Dreamux Rush change declaration (not a generated changelog edit), then rerun the exact-head gate.

### P2 — The stage-design text claims streaming that the script does not perform

`packages/dreamux/skills/shared/workflow/references/code-review.md:25-35`

The reference correctly says Find uses a full `parallel()` barrier for cross-lens deduplication, but then says a finding from a fast lens is verified while slower finder work continues. The script awaits the entire finder `parallel()` at lines 218-230, deduplicates afterward, and only then starts `pipeline()` at line 249. No verification overlaps finder execution.

Either remove the streaming claim and say findings advance independently only within Verify, or change the executable graph to actually stream findings without violating the documented dedup barrier.

### P2 — The finding schema permits output that the report contract cannot satisfy

`packages/dreamux/skills/shared/workflow/references/code-review.md:116-140` and `:278-284`

`line` is optional in `FINDINGS_SCHEMA`, while the final reporter is required to emit a `file#line` citation for every issue. A runtime-valid finding may omit `line`, leaving the reporter unable to satisfy its prompt without inventing a location. Make `line` required, or explicitly support file-only findings in both the schema/key and report format.

### P2 — The ultracode scope description contradicts the implemented transform

`packages/dreamux/skills/shared/workflow/references/code-review.md:82-86`

The reference says helper functions “stay at module scope.” The #320 normalizer preserves only the `meta` export at module scope and wraps every other top-level statement, including helper declarations, inside the generated async entry function. The example works, but the ownership/scope claim is false and should be corrected.

### P2 — PR description still documents the pre-#320 dialect

The PR body still says the script uses a default-export entry, string phases, single-argument pipeline stages, explicit JSON-output boilerplate, the 200-agent cap, and that #320 is future work. Exact HEAD does the opposite. Update the public PR description to the final top-level dialect, 3-argument pipeline contract, native structured-output semantics, and 1000/4096/16 limits.

### P3 — Make channel posting conditional on an available provider tool

`packages/dreamux/skills/shared/workflow/references/code-review.md:39-42`

The shared skill is injected even when no provider reply tool exists. Replace the unconditional “Post it through the channel” instruction with the repository’s established conditional form: use a provider-exposed reply tool when one is available; otherwise return the report and let the caller choose another explicit publication path.

### Verified

- The three-way merge of `SKILL.md` with current `next` is conflict-free and leaves the References table after the full #320 Script API section.
- `skill-creator` validation passes for the reconstructed skill tree.
- The complete code-review script passes the production ultracode normalizer.
- A real child-runner probe completes all five phases with 13 synthetic agent calls and returns the expected structured finding plus markdown report.
- The branch is one commit behind current `next` (the merged #320); rebase/update before the final exact-head gate so the reviewed source and CI base are unambiguous.

ryanxiang7:
All review findings addressed at head 8b3f847 (branch merged up to current `next`, includes #320):

- **P1 silent finder loss — fixed**: lens results are index-aligned with the lens list; all-lenses-failed returns an explicit error result, partial failure returns `coverage: {complete: false, failedLenses}`, and both the no-issues report and the final report name which lenses actually ran (the reporter is instructed to append a failed-lens line when coverage is incomplete).
- **P1 missing rush change — fixed**: `common/changes/@excitedjs/dreamux/skill-workflow-code-review-reference_2026-08-09-00-00.json` added (patch, docs-only bundled skill content).
- **P2 verify-overlap prose — fixed**: stage design now states the find barrier is deliberate (dedup + coverage accounting) and describes pipeline independence per-finding, not across the find/verify boundary.
- **P2 line fabrication — fixed**: the reporter cites `file#line` only when the finding carries a line number, file path alone otherwise, and is told never to invent one.
- **P2 module-scope claim — fixed**: the entry description now matches the normalizer (helpers are wrapped into the canonical entry function; hoisting keeps them callable as written).
- **P2 stale PR description — fixed**: PR body rewritten for the post-#320 dialect and limits.
- **P3 channel wording — fixed**: report routing now says to post through the channel reply tool when one is available.

Ready for re-review.

YourWildDad (changes requested):
## Re-review — exact HEAD `8b3f84793fc918cee542404caf902ed3d4358fd7`

**CHANGES_REQUESTED**

The seven prior dispositions were implemented as described: the branch is current with `next`, the Rush patch declaration is present and its CI gate passes, finder results are index-aligned, the barrier/module-scope/report-location/channel wording and PR body are corrected, and the complete script still passes skill validation plus production normalization.

One load-bearing failure class remains, found by running the exact script through the real child runner with injected failures.

### P1 — `coverage.complete` still ignores failed context and failed verification

`packages/dreamux/skills/shared/workflow/references/code-review.md:190-215` and `:279-306`

The new `coverage` value accounts only for finder calls.

- If the guidance-discovery TeamMate fails, `ctx[0]` becomes `null`, `guidancePaths` silently becomes `[]`, the guidance lens can still settle successfully without reading the applicable files, and the terminal result says `coverage: { complete: true }` plus “Checked for ... guidance-file compliance.”
- If a finding is produced but all three verifier TeamMates fail, the first pipeline stage returns `[null, null, null]`; the second stage records `votes: 0`; `confirmed` becomes empty; then lines 304-306 return the same clean “No issues found” report with `coverage.complete: true`.

I reproduced both exact outcomes through the child runner. Track load-bearing context completion and per-finding verification completeness in the coverage result. Do not issue a clean review when guidance discovery/required context failed, or when any candidate finding received fewer than the required two settled votes. An explicit incomplete/error result is appropriate; a low confidence score from two or more settled votes can still be an ordinary rejected finding.

### P2 — “always `.filter(Boolean)`” now contradicts the coverage-preserving recipe

`packages/dreamux/skills/shared/workflow/references/code-review.md:57-66` and `references/orchestration-patterns.md:24-31`

Both references tell readers to always filter nulls before aggregation, while this fix correctly depends on preserving positional nulls until failed-lens coverage is recorded. Qualify the rule: account for positional/required coverage first, then filter only when missing entries are intentionally disposable.

### P3 — Long progressive-disclosure references need a compact contents map

The final files are 349 and 146 lines. The skill authoring contract for references longer than 100 lines asks for a concise table of contents so an agent preview can see the scope without scanning the whole artifact. Add a short contents list near the top of each reference (especially the 349-line runnable recipe).

### Re-verified

- complete finder path returns `coverage.complete: true`;
- one failed finder returns `coverage.complete: false` with the exact failed lens and an incomplete report;
- all failed finders return the explicit “change was not reviewed” error result;
- the full happy path, optional-line report contract, updated PR body, top-level normalization, skill validation, Rush change declaration, and merge-with-`next` state are correct.

ryanxiang7:
Second review round addressed at head 7c68f49:

- **P1 guidance-discovery failure — fixed**: a null discovery result is now distinct from a successful empty list. On failure the guidance lens is skipped (it must not audit an empty list) and `'guidance'` is recorded in `coverage.failedLenses`, so the run reports incomplete coverage instead of `complete: true` + 'checked guidance compliance'.
- **P1 all-verifiers-failed — fixed**: verify pipeline results are index-aligned with the unique findings; a null stage result or fewer than two settled votes routes the finding into a returned `unverified` array with `coverage.unverifiedFindings`, the no-confirmed path reports 'No issues confirmed. N finding(s) could not be verified…' instead of a clean review, and the reporter renders an Unverified section without confidence claims.
- **P2 filter(Boolean) guidance — fixed**: both references now teach positional coverage accounting first (index-aligned results, name the gap) and `.filter(Boolean)` only where item identity no longer matters, with an explicit warning that blind filtering produces false-clean results.
- **P3 Contents — fixed**: both references open with a short Contents section.

The script block re-passes a syntax check under normalizer-equivalent wrapping. Ready for re-review.

YourWildDad (approved):
## Final review — exact HEAD `7c68f493125c5902a1e267141398ada7dbfdad39`

**APPROVE**

All prior findings are resolved on this exact head. The recipe now preserves positional failure information until coverage is accounted, distinguishes failed guidance discovery from a successful empty guidance set, records failed finder lenses, returns findings with fewer than two settled verifier votes as `unverified`, and prevents those failures from becoming a false clean review. The two references also carry accurate containment guidance and compact contents maps.

Independent ownership adjudication confirms this belongs in the recipe: `parallel()` / `pipeline()` returning position-aligned `null` for contained item failures is the settled #320/ultracode-compatible public contract. Only the application script knows whether a null is disposable or a required coverage gap; changing helper result shape or whole-run failure behavior would be a separate breaking Workflow contract change, not part of this documentation PR.

### Exact-head verification

- shared skill validation passes;
- the complete embedded script passes the production ultracode normalizer;
- real child-runner probes pass for: complete no-findings, partial finder failure, all finder failure, guidance-context failure, summary-context failure, one/two/all verifier failures, reporter failure, and mixed confirmed + unverified findings;
- optional line citations are handled without fabrication;
- Rush patch declaration is present and valid;
- branch is based on current `next` (includes #320);
- CI is 8/8 green on this exact head (Ubuntu/macOS full Rush, change declaration, KB, shellcheck, commit metadata, full-history gitleaks).

No remaining findings.

PR #320, opened 2026-08-08
feat: support ultracode workflow scripts

### LINKED ISSUE #318: Align workflow script API and limits with Claude Code's built-in Workflow (ultracode) orchestrator

YourWildDad:
## Summary

- accept the issue #318 top-level Workflow dialect while preserving existing default-export module semantics
- add object-form phase metadata, three-argument pipeline stages, runtime-native structured-output enforcement, and bounded limits of 16 concurrent / 1000 lifetime agent calls / 4096 helper inputs
- inject Workflow-owned role guidance through the existing neutral owned-TeamMate launch seam without widening the public agent-runtime ABI
- add the two unmodified issue fixtures, contract coverage, bundled skill updates, maintenance guidance, and an upgrade note

## Scope

Implements requests 1-4 from #318.

Explicitly out of scope: global budget, per-agent model or effort selection, resume, nested workflows, worktree isolation, and requests 5-9.

## Validation

- exact final head: `c57bd03588162f920e675ac24f4d79eede554e99`
- `node common/scripts/install-run-rush.js lint`
- `DREAMUX_SKIP_LIVE_CODEX=1 node common/scripts/install-run-rush.js test`
- `.agents/scripts/check.sh`
- `git diff --check`
- focused exact-head Workflow/admin/MCP tests: 87/87 passed
- exact-head deterministic Dreamux tests: 914 passed, 4 opt-in/live tests skipped
- exact-head GitHub CI: 8/8 checks passed on Ubuntu and macOS
- earlier source-build live integration at implementation head `542160057a1ba855ac537cf6fef0cbf41b35c93a` for unmodified `deep-research-max.mjs`: completed, 84/84 agents, 0 failed, one terminal completion
- earlier source-build live integration at implementation head `542160057a1ba855ac537cf6fef0cbf41b35c93a` for unmodified `code-review-max.mjs`: completed through Gate/Context/Find/Verify/Report, 18/18 agents, 0 failed, one terminal completion

The ordinary live Codex suite was not used for the final deterministic gate because the local `codex_apps` MCP initialization previously timed out before test marker injection. The unmodified issue fixtures were separately exercised at the earlier implementation head by importing the checkout's built `WorkflowService` and `TeammateCollection` and connecting them to real configured runtime providers. They did not install or restart the operator's Dreamux service and did not exercise the CLI/MCP `workflow_run` entrypoint. The exact final head was covered by deterministic tests and CI, not another live runtime run. Installed-service end-to-end coverage remains intentionally unrun.

Closes #318


ryanxiang7:
## Quality review (complexity / architecture / module boundaries) — head 5421600

Max-effort multi-lens review (complexity, architecture, module boundaries, reuse & convention; findings cross-verified, duplicates merged). No blocker, no high. 2 medium + 7 low, all quality-level; correctness was reviewed separately and approved.

### Medium

**M1. `max_concurrency` bounds hardcoded in three unshared places with divergent semantics** (flagged independently by three lenses)
`packages/dreamux/src/mcp/teammate-mcp.ts` JSON schema (`maximum: 16`), the new manual throw in `workflowRunArgs`, and `MIN_MAX_CONCURRENCY`/`MAX_MAX_CONCURRENCY` + `clampMaxConcurrency` in `packages/dreamux/src/service/workflow-service/index.ts`. Consequence: `17` via the TeamMate MCP throws, while the admin-socket `workflow.run` path silently clamps to 16 — two semantics for the same input. This PR's own 8→16 bump had to touch the literals in lockstep, demonstrating the drift hazard. Sibling integer params in the same file (`limit` 1..100, `turns` 1..5) declare schema bounds only. Suggest: export the bounds from workflow-service and reference them in the MCP schema/check, and unify reject-vs-clamp.

**M2. `script-normalizer.ts` literal reader materializes values that are always discarded**
`readPlainLiteral`/`readArrayLiteral`/`readObjectLiteral`/`plainPropertyName` (~75 lines) build real JS values (null-prototype record, key stringification), but the sole caller `readUltracodeMeta` has its return value discarded; the only use is an `isRecord` check. Semantic meta validation already happens post-evaluation in `assertWorkflowScriptMeta` for both dialects. A void validator that only walks node types and throws on non-literals is equivalent and much smaller. Related: the `as VariableDeclaration` cast and null guard in `readUltracodeMeta` are unreachable (`ultracodeMetaExport` already requires `init !== null`); returning the init expression directly removes both.

### Low

1. `WorkflowScriptMeta` stays in `protocol.ts` although meta never crosses the runner IPC (validated child-side, discarded; no parent/child message carries it). Every other export there is an IPC message shape. Move it to `script-meta.ts` next to its validator.
2. Dead generality in the normalizer: the duplicate-`meta`-export branch is unreachable (acorn rejects duplicate exports at parse time), and the `export * as default from` handling only selects which error a doomed script gets (imports are disabled at link) — dropping it also removes the hand-rolled union parameter type.
3. `run.ts` structured-output path funnels the empty-result case through a synthetic throw into the JSON.parse catch, and derives the final status at the `completeAgent` call site 30 lines away. Side effect: an empty result surfaces to the script as "not valid JSON" while the accurate message only reaches the debug log. A flat if/else setting `result`/`runnerError`/`status` in one place is equivalent and clearer.
4. The 4096 helper-input cap is a magic literal duplicated in `parallel` and `pipeline` with parallel error strings; hoist one constant.
5. Phase `model` is accepted but fully inert, and meta is never persisted/projected/logged — an author writing `model: 'opus'` expecting the parity dialect's semantics gets zero signal anywhere. Docs consistently say "descriptive metadata only", but doc-only mitigation is weak for a name that collides with a live concept; a one-line structured log at validation (or rejecting the field until it means something) would make the inertness observable.
6. The eslint ignores enumerate the two fixture filenames; the directory is dedicated to non-standard-ESM dialect fixtures, so ignore `tests/fixtures/workflows/*.mjs` once instead of growing the list per fixture.
7. Docs: (a) `.agents/reference/dynamic-workflow-usage.md` was only partially translated — ~140 Chinese lines remain interleaved with the new English sections, violating root `CLAUDE.md` ("Write all repo docs … in English") and leaving the page less uniform than before; (b) the 16/1000/4096 limits are restated verbatim in both `current-architecture.md` and `dynamic-workflow-usage.md`, against `.agents/CONTRIBUTING.md`'s one-owner-per-fact rule (the bundled skill copies are a legitimate self-containment exception).

### Verified non-findings

Adversarially checked and dismissed: `errorMessage`/`isRecord` "duplication" is pre-existing or deliberate child-entry locality; the two `isUnsupportedFeatureError(…, 'outputSchema')` call sites cover exactly the two real producer paths (capability-gate throw, claude-code failed turn); the reject-vs-null asymmetry for structured output is ABI-grounded in `dreamux-types` (provider contract breach fails loud, agent fault contained as null) and layered correctly; `WORKFLOW_AGENT_SYSTEM_PROMPT` ownership stays in workflow-service with teammate-collection holding only a neutral ordered-append seam (operation-before-identity order is defensible); normalizer placement in the runner child matches the architecture reference (1MB admission check untouched parent-side); no provider special-casing; commit messages, rush change files, and the archived proposal follow repo conventions.

**Suggested disposition**: fix M1, M2, and 7(a) before merge (M1 is a semantics split across control planes; 7(a) is an explicit rule violation); the rest can follow up.

YourWildDad:
Addressed at exact HEAD `825201948cabf3cde897475d4f45a964122d17f1`.

### Medium

- **M1 — resolved.** Workflow-service now privately owns the default/minimum/maximum and one `parseWorkflowMaxConcurrency()` validator. MCP references the same owner constants for its schema, while MCP mapping, admin, and direct service admission all reject non-integer or out-of-range values before durable run creation. The prior clamp behavior and control-plane split are removed.
- **M2 — resolved.** The ultracode normalizer now walks the Acorn AST as a void structural validator. It no longer materializes discarded metadata values, and metadata semantic validation still occurs once after module evaluation and before primitives are installed.

### Low

1. **Resolved.** `WorkflowScriptMeta` moved from the IPC protocol module to its child-local validator owner.
2. **Resolved.** Removed duplicate-export and export-all-as-default generality from the legacy/default-export path. Existing `export default` and `export { run as default }` sources remain byte-for-byte unchanged; export-all remains rejected as an invalid ultracode export before any agent starts.
3. **Resolved.** Structured-output settlement now computes status/result/error in one flat path and distinguishes empty successful output from invalid JSON.
4. **Resolved.** The helper input limit has one runner-private constant shared by `parallel()` and `pipeline()`.
5. **Intentionally unchanged.** `phase.model` remains accepted descriptive compatibility metadata, as required by the approved issue #318 scope. It is explicitly documented as inert and not persisted/projected. Logging arbitrary script metadata would add a new observability/data-exposure contract, while rejecting the field would violate the accepted dialect contract.
6. **Resolved.** ESLint now ignores the dedicated `tests/fixtures/workflows/*.mjs` fixture directory pattern.
7. **Resolved.** The usage guide is fully English. It now owns the exact user-facing numeric limits; the architecture reference links to that owner rather than repeating the values. The bundled skill remains self-contained. A focused KB check now prevents Han text from re-entering current reference docs.

Additional contract coverage was added for all invalid concurrency classes, accurate empty structured-output failure, both legacy default-export forms, and export-all rejection before agent start. The Dreamux Rush change note and implemented proposal were synchronized with the final admission contract.

Validation on this exact head:

- focused Workflow/admin/MCP suite: 87/87 passed;
- `rush build`, `rush typecheck`, `rush lint`, and Dreamux test typecheck passed;
- deterministic `rush test --to @excitedjs/dreamux`: Dreamux 914 passed / 4 skipped, all selected dependency package suites passed;
- KB check, Rush change verification, and `git diff --check` passed;
- GitHub CI: 8/8 checks passed on Ubuntu and macOS, including full-history gitleaks.

Installed-service E2E remains intentionally unrun, as already stated in the PR validation section.

YourWildDad:
Final exact-head follow-up: the two resident-review P3s are now resolved at `c57bd03588162f920e675ac24f4d79eede554e99`.

- `0d499a59a571a88df3270a6b8fd0f39e631926cd` updates the implemented proposal so `WorkflowScriptMeta` is correctly owned by private `script-meta.ts`, while `protocol.ts` remains IPC-only.
- `c57bd03588162f920e675ac24f4d79eede554e99` removes the remaining Acorn-parser-invariant `declarator` and initializer runtime guards, using local non-null assertions after the existing single-`const` declaration shape checks.

The focused runner test remains 23/23; Dreamux build/typecheck and KB/diff checks pass. The resident panel is re-reviewing the final exact head, and final-head CI is in progress.

YourWildDad:
Final gate passed at exact HEAD `c57bd03588162f920e675ac24f4d79eede554e99`.

- The same resident architecture, lifecycle, and complexity reviewers all returned **APPROVE** on this exact head.
- Both follow-up P3s are resolved: archived metadata ownership is aligned with `script-meta.ts`, and unreachable Acorn declarator/initializer guards are removed.
- GitHub CI is 8/8 green on the exact head, including Ubuntu/macOS full Rush jobs and full-history gitleaks.
- The branch is clean and synchronized with the PR head.

No installed-service E2E was run; the PR validation section states the source-build and exact-head boundaries explicitly.

ryanxiang7 (approved):
Re-reviewed at exact HEAD c57bd03 against the quality findings. M1 (single limits owner via workflow-service/limits.ts, unified fail-loud admission across MCP/admin/service — verified in diff), M2 (void AST validator, no discarded materialization), and L1-L4/L6/L7 all verified resolved in the diff; the L7 fix goes further than asked with a CI-enforced Han-text check on current reference docs. L5 (inert phase.model) rejection rationale is sound: dialect-compat contract from #318 scope, documented inert, and logging arbitrary script metadata would create a new exposure contract. CI 8/8 green. APPROVE.

PR #321, opened 2026-08-09
fix(codex): adapt portable output schemas

YourWildDad:
## Summary

- compile the provider-neutral structured-output subset into Codex strict schemas before `turn/start`
- restore optional-field semantics before Codex publishes a completed turn, while preserving required nullable values
- bind private codecs to active-turn slots/native turn ids and reject incompatible folding before submission
- document the adapter-owned contract, archive the implemented proposal, and add package-scoped Rush declarations

## Exact head

`0b9d91c0ab77667a0c0a94162f4f15100f7bcb15`

## Validation

- `@excitedjs/agent-runtime-codex`: 70/70 tests; final simplification focused set 42/42
- `@excitedjs/dreamux` CI-mode tests: 916 passed, 5 skipped
- Codex and Dreamux production/test typecheck
- Codex and Dreamux build + lint
- `.agents/scripts/check.sh`
- `git diff --check`
- GitHub CI: 8/8 jobs passed on Ubuntu/macOS, including full Rush build/typecheck/lint/test, Rush change, KB, shellcheck, commit metadata, and full-history gitleaks
- resident architecture, lifecycle, and complexity reviews: APPROVE on the exact head
- external Devbox review: APPROVED on GitHub

## Live Codex gate

The opt-in real Codex model gate is included. On the current machine, Codex 0.147 accepted the compiled `turn/start.outputSchema` (`turn/start error=none`) and entered the model turn, but the ChatGPT model backend repeatedly reconnected and never produced a terminal result. This is recorded as an environment/model-service blocker rather than reported as a passing end-to-end model gate. CI keeps the existing explicit model-gate opt-in behavior.

### DIFF
--- a/.agents/archive/proposals/README.md
+++ b/.agents/archive/proposals/README.md
@@ -5,6 +5,7 @@ implementation guidance; use the linked decisions and reference docs first.
 
 | Proposal | Current pointer |
 |---|---|
+| [Codex portable output schema adapter](codex-portable-output-schema.md) | Implemented in `@excitedjs/agent-runtime-codex`; see [Provider Runtime](../../domains/provider-runtime.md#codex-portable-output-schema) and the [Codex package README](../../../packages/agent-runtime/codex/README.md#portable-structured-output) |
 | [Concrete entity name suffix length](concrete-name-suffix-length.md) | [Current architecture](../../reference/current-architecture.md), [Dispatcher orchestration](../../domains/dispatcher-orchestration.md), [Service topology](../../reference/service-topology.md) |
 | [Dynamic Workflow 编排](dynamic-workflow.md) | Implemented in PR #312 (`feat(workflow): Dynamic Workflow MVP`); see bundled skill `workflow` and `service/workflow-service/` |
 | [Workflow ultracode dialect parity](workflow-ultracode-dialect-parity.md) | Implemented issue #318 requests 1-4; see [Dynamic Workflow usage](../../reference/dynamic-workflow-usage.md), [Current architecture](../../reference/current-architecture.md), and bundled skill `workflow` |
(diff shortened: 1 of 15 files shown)


ryanxiang7 (approved):
Reviewed at exact HEAD 0b9d91c0ab77667a0c0a94162f4f15100f7bcb15 against origin/next. APPROVE.

**Scope discipline**: all runtime changes live in `@excitedjs/agent-runtime-codex` (codec + turn-manager); zero dreamux-core src changes, so the neutral ABI claim holds by construction. Docs (README, provider-runtime KB, archived proposal), two change files (codex patch, dreamux test-only none), and the version-bump-free package.json description tweak all check out.

**Codec** (`output-schema-codec.ts`): whitelist compilation is sound — non-nullable closed root object; nested closed objects / single-schema arrays; `[T,"null"]` unions only; enum canonicalization; numeric bounds with min<=max check; every unsupported keyword fails with a path-specific `UnsupportedAgentRuntimeFeatureError(outputSchema)`, which lands exactly on run.ts's existing `isUnsupportedFeatureError` reject path — local fail-loud before a provider 400, as spec'd. The optional→required+nullable compile with per-property `omitNull` restoration plans is reversible by design, and the `optional && acceptsNull` ambiguity correctly fails instead of guessing.

**Lifecycle**: codec compiled before slot admission (compile failure never claims a slot), carried per-turnId in `pendingTurns` (late completions after slot release still restore), folding gated by fingerprint equality with structured/unstructured mixing rejected, stop/restart discard the codec and drop late completions, restoration failure settles `failed` without polluting lastResult.

**My three pre-flagged review points, all landed**: (1) additionalProperties semantics is an explicit documented decision (must be literal `false`; open objects fail loud — README + KB state it); (2) optional-placeholder restore behavior is locked by codec contract tests; (3) restoration performs structural validation with tested failure paths (shape mismatch, missing assistant JSON, invalid JSON), plus a runtime-level failed-settlement test.

**Tests**: 5 codec contract tests, 7 turn-manager lifecycle tests, 2 runtime settlement tests, and a real-model live gate for admission+restoration — no existing assertion weakened. CI 9/9 green including full-history gitleaks; public-safety scan of the diff is clean.

PR #322, opened 2026-08-09
fix(codex): dispose turn collectors

YourWildDad:
## Summary

- make Codex notification registrations return an idempotent unsubscribe callback
- let each turn collector own and release its notification subscription on completion, terminal failure, explicit disposal, or direct-run cleanup
- dispose abandoned collectors after rejected `turn/start` submissions and during runtime stop

## Problem

`TurnManager` subscribes a collector before calling `turn/start`. If every submission for that slot is rejected, the slot is cleared but the collector remains registered on the resident Codex client. A later turn on the same thread is then observed by both the stale and current collectors, and repeated failures accumulate handlers and buffered turn data.

## Tests

- `rush build --to @excitedjs/agent-runtime-codex`
- Codex focused tests: 32 passed
- Codex full suite: 74 passed
- Codex source/test typecheck and lint
- `.agents/scripts/check.sh` (92 files reachable)
- `rush change --verify --target-branch origin/next`
- `git diff --check origin/next...HEAD`

The new regressions cover normal completion unsubscription, idempotent explicit disposal, rejected direct `runTurn`, rejected plain-to-structured recovery, rejected structured-to-plain recovery, and runtime stop.

### DIFF
--- a/.agents/domains/provider-runtime.md
+++ b/.agents/domains/provider-runtime.md
@@ -167,9 +167,17 @@ not call `onTurnCompleted` or mutate `lastResult`; it emits one ordinary failed
 teardown/restart, and late completion clear or discard in-memory codecs through
 the same turn lifecycle and never restore or settle twice.
 
+Each collector owns and unregisters exactly one Codex notification handler.
+Normal completion and terminal failure close it automatically; rejected
+`turn/start`, runtime stop, and direct `runTurn` cleanup dispose it explicitly.
+An abandoned collector therefore cannot buffer a later turn or accumulate
+handlers on the resident Codex client.
+
 Source:
 
 - `/packages/agent-runtime/codex/src/output-schema-codec.ts`
+- `/packages/agent-runtime/codex/src/events.ts`
+- `/packages/agent-runtime/codex/src/rpc.ts`
 - `/packages/agent-runtime/codex/src/turn-manager.ts`
 - `/packages/agent-runtime/codex/src/runtime.ts`
 - `/packages/agent-runtime/codex/tests/output-schema-codec.test.ts`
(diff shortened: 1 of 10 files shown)


ryanxiang7 (approved):
Reviewed with focus on the collector notification unsubscribe lifecycle. All paths verified against the PR head:

- **Normal completion / terminal failure**: `resolveAwaiting` and the synchronous `awaitTurn` fast paths all call `closeCollector()` after settling, which now unsubscribes and clears all three buffers. The `closed` re-entry guard makes it idempotent.
- **Rejected `turn/start`**: `recordTurnStartFailure` disposes only when `primaryFailed && pendingSubmissions === 0`, so in-flight follower submissions are never cut off; the candidate-recovery path (`primaryFailed && candidateTurnId !== null`) correctly activates instead of disposing. Since claim + counter increment happen synchronously before the awaited submit, the counter cannot race a late follower.
- **Direct `runTurn`**: `return await` inside try ensures the `finally` dispose runs after settlement; dispose after natural close is a no-op.
- **Runtime stop**: `stop()` disposes the active slot's collector before clearing `pendingTurns`; the dispose-triggered rejection lands in a microtask after `pendingTurns.clear()`, so the `pendingTurns.delete` mutual-exclusion guard drops it and each turn settles exactly once as `stopped`. No double-settlement.
- **Active-turn folding**: unaffected — the collector is per-slot, disposal only happens on full-slot failure or stop, and `slot.turnId !== null` early-returns keep a tracked turn's collector alive while followers fold. The new plain↔structured recovery regressions pin this down nicely.
- `Set`-based handler registry: self-deletion during `dispatchNotification` iteration is safe per JS Set semantics; no other `onNotification` consumers exist outside events.ts, so the signature change is fully contained.

Test diffs only extend assertions (handler-count checks) and adapt fakes to the new signature — no weakened contracts.

Non-blocking nit: the new change file lacks a trailing newline.

LGTM.

Issue #323, opened 2026-08-10
Breaking: make Workflow top-level-only with structured JSON args


YourWildDad:
## Problem

`workflow_run.args` is currently advertised with an empty MCP schema. In a live run, a JSON object argument reached the script as JSON text, so a native-form script reading `args.question` exited before starting any agent.

Dynamic Workflow also still carries a second, Dreamux-only entry dialect based on `export default async function run()`. The feature is pre-adoption, so we should remove that compatibility branch rather than preserve two script contracts.

## Proposed contract

- Support one script dialect only: literal `export const meta = ...` followed by the executable top-level body.
- Allow top-level `await` and `return`; the returned value is the run result.
- Reject default exports, named-default exports, other exports, imports, pre-meta executable statements, and non-literal metadata before any agent starts.
- Expose `workflow_run.args` as a direct JSON value. Objects and arrays reach the script as JavaScript objects/arrays; omitted args remain `undefined`; JSON-looking strings remain strings.
- Remove string-form `meta.phases`; keep native object phases `{ title, detail?, model? }`.
- Keep Dreamux's existing structured-output unsupported/invalid-success rejection semantics as an explicit provider-driven deviation from native Claude Code Workflow.

## Scope

The implementation should update the MCP schema, Workflow compiler/runner, tests, bundled workflow skill, current KB, and a breaking Rush change. It should delete the default-export compatibility path rather than add argument parsing glue.

The exact implementation proposal is in `.agents/proposals/workflow-top-level-json-args.md` on the linked draft PR.

ryanxiang7:
Reviewed the proposed contract line-by-line against the native Claude Code Workflow contract (as shipped in the harness tool definition). **Verdict: conformant — no blocking findings.** Point-by-point:

| Proposed | Native contract | Match |
|---|---|---|
| Single dialect: literal `export const meta` + top-level body | Native has exactly one dialect; scripts "must begin with `export const meta = {...}`"; no default-export mechanism exists | ✅ |
| Top-level `await` / `return`; return value is the run result | Same | ✅ |
| Reject default / named-default / other exports, imports, pre-meta statements, non-literal meta | Native: only the meta export exists; meta is a pure literal; script begins with meta. Fail-loud rejection is the right reading | ✅ |
[...]

PR #324, opened 2026-08-10
feat(workflow): use top-level scripts and JSON args

YourWildDad:
## Summary

Implements #323:

- remove the default-export `run()` compatibility dialect
- compile one native-style `export const meta` + top-level Workflow body
- expose `workflow_run.args` as direct structured JSON values
- validate JSON args once at the Workflow service boundary before durable run creation
- remove string phase metadata and the duplicate runtime metadata validator
- preserve submitted source line numbers and prior module strict semantics
- keep existing abort/helper/structured-output behavior

## Exact head

`64b4e587074910db246c4fe6d1e5c253126c06f9`

## Validation

- focused Workflow/MCP/admin/service/skill/lifecycle: 111/111
- final runner/compiler strict semantics: 34/34
- Dreamux CI-mode full package: 933 passed, 5 skipped
- Rush rebuild through `@excitedjs/dreamux`: 8/8 operations
- Dreamux source/test typecheck and lint
- `.agents/scripts/check.sh`
- `git diff --check`
- stale current-text and public-artifact safety sweeps
- GitHub CI: 8/8 checks green on Ubuntu/macOS (Ubuntu failed once on the pre-existing 1000-call timing test, then passed unchanged on rerun)
- resident architecture/lifecycle/complexity reviews: APPROVE on exact head
- external Devbox review: APPROVED on exact head

## Breaking change

Default-export Workflow entries and string phase metadata are rejected. Pass direct object/array `workflow_run.args` values rather than JSON-stringifying them. No state/config rebuild is required.

Refs #323


ryanxiang7 (commented):
Spec review at commit a8d3349 (sha256 verified = `324c82cf...ecf2`). **Conformant with the native Claude Code Workflow contract — approve as spec.** 

Everything I flagged on #323 is already incorporated: `whenToUse` retained in the metadata contract; the Delete List sweeps module-form/string-phase examples in skill + KB in the same change; the Acceptance list pins omitted-args-stay-`undefined`, JSON-looking-strings-stay-strings, and requires the two #318 fixtures plus the shared `deep-research.mjs` shape to run unmodified. The root-cause statement (untyped MCP `args` schema at the tool-projection layer, not service serialization) matches what the service/admin/runner code paths actually do, and the private-async-closure design correctly gives top-level `return` native semantics without rewriting the persisted source or changing `script_hash`. Keeping the structured-output rejection as an explicit retained deviation, separate from this cutover, is the right change discipline.

Two precision notes, neither blocking:

1. **State the unknown-key policy for `meta` and phase objects.** The spec requires `{ title, detail?, model? }` but doesn't say whether `{ title, foo: "bar" }` (or an unknown top-level meta key) fails or passes. The current validator tolerates unknown keys silently. Either choice is defensible — strict rejection matches the "fail before any agent starts" ethos; tolerance matches native's unspecified behavior — but the compiler will encode one of them, so the spec should name it rather than let the implementation decide.
2. **Line-number fidelity (nice-to-have, not scope creep).** Wrapping the body in the private closure shifts runtime error line numbers relative to the author's source. A one-line prefix pad (or documenting the offset) would make script stack traces actionable. Fine as a noted follow-up if out of scope here.

Ready to re-review the implementation at exact head against the Acceptance list.

ryanxiang7 (commented):
Re-reviewed the spec at commit 73b2400 (sha256 verified = `651ab980...4cba12`), diffed against a8d3349. **Both precision notes are resolved; the additional tightenings are sound. Spec approved from my side at this commit.**

- **Unknown-key policy** — now explicit: unknown recursively-plain-literal keys on `meta`/phase objects are accepted and ignored, reserved for future native-compatible metadata. Tolerant is the right pick for forward compatibility with the native dialect, and it's now a documented decision rather than an implementation accident. ✅
- **Line-number fidelity** — padding the closure prefix to the metadata declaration's lines so runtime stacks match the submitted script is the clean version of what I asked for, and it's pinned in Acceptance. ✅
- **Static vs dynamic import split** — compile-time rejection for static imports, existing VM-hook fail-loud for dynamic `import()`, no speculative recursive AST walker: proportionate, and consistent with the repo's reject-edge-case-machinery default. ✅
- **Single JSON-args validator owner** — one private Workflow-service recursive validator as runtime authority, MCP restricted to a top-level type union with guidance, no public JsonValue DTO, admin/MCP forward-only: this is the same single-owner discipline that the max_concurrency fix (#319 M1) established. The reject list (non-finite, explicit `undefined`, functions/symbols/bigint, cycles, sparse arrays, non-plain objects) is complete for the boundary. ✅
- **Compiler as sole metadata owner** — deleting `script-meta.ts` / `assertWorkflowScriptMeta` / namespace-meta validation removes the double-validation seam entirely; with the single dialect there is no post-evaluation meta to re-check. ✅

One observation, not a finding: v2 drops the shared `deep-research.mjs` shape from Acceptance, leaving the two checked-in fixtures. That's fine — `deep-research-max.mjs` already exercises `whenToUse` + object phases + top-level entry, so dialect coverage is unchanged; just confirming the narrowing was intentional rather than accidental.

Ready for implementation review at exact head against the updated Acceptance list.

ryanxiang7 (approved):
Implementation reviewed at exact HEAD 100b589f5412de33e3091e0372ac98908ca6e116 against the approved spec (52525a25) and its Acceptance list. **APPROVE.**

**Compiler** (`script-compiler.ts`): meta-first enforced via `program.body[0]` (comments/whitespace naturally allowed — they aren't statements); const-only single-declarator `meta` with no specifiers/source; recursively-plain-literal materialization with the good error split between "not a literal tree" and "literal but not an object"; unknown root/phase keys accepted while known-field type errors fail loudly — exactly the spec combination. The line-preserving prefix is correct: the split keeps separators at odd indexes, first meta line becomes `(` + padding, interior lines blank-padded, last line `async()=>{`; the single-line case can't underflow because any valid meta export exceeds the 11-char prefix. Static imports rejected at compile; dynamic `import()` left to the VM hook with a test proving it fails only on execution.

**Runner**: `SourceTextModule`+namespace lookup replaced by one `vm.Script` of the compiled closure; primitives installed only after compile succeeds, and the pre-meta test proves an early `agent()` call cannot spawn anything. `script-meta.ts`/`assertWorkflowScriptMeta` fully deleted — compiler is the single metadata owner per spec.

**Args**: MCP schema is the top-level JSON union with the exact anti-stringify guidance; the single private service validator covers finite numbers, explicit `undefined`, non-string keys, accessor/non-enumerable properties, sparse arrays, cycles (with correct ancestor cleanup), and non-plain prototypes; `Object.hasOwn` gating keeps omitted args out of validation and `undefined` in the script. Nested object/array round-trips are asserted at MCP, admin, and service layers.

**Delete list & docs**: normalizer and script-meta gone; the only remaining `export default async function run` occurrence in the repo is the rejection-matrix test asserting it fails; string-phase examples swept from skill/KB; SKILL.md now teaches one entry form; both #318 fixtures execute unmodified per test. Rush change leads `BREAKING: Review:`, states no rebuild, contains no `Rebuild:`. CI green.

**One low observation, non-blocking**: the previous `SourceTextModule` evaluated the body in module (strict) mode; `vm.Script` + the private closure now runs sloppy mode, so an undeclared assignment (`x = 5`) silently creates a context global where it used to throw `ReferenceError`. Neither the contract nor the fixtures depend on this, but it's a silent-failure surface for author typos. Cheap hardening if wanted: emit `'use strict';` inside the padded prefix (budgeting its 13 chars in the padding math, falling back to an extra line only if a pathological one-line meta is too short) — fine as a follow-up or a wontfix with a note.

ryanxiang7 (approved):
Re-reviewed at exact HEAD 64b4e587074910db246c4fe6d1e5c253126c06f9. APPROVE.

The strict-mode fix is the complete version: the wrapper is now a plain-invoked strict ordinary async function — `(async function(){"use strict"; ...})()` — so the directive prologue strictifies the whole body and plain invocation pins top-level and nested `this` to `undefined`. Prefix budget verified: the 32-char strict opening fits the shortest legal single-line meta (42 chars), so the padding math cannot underflow; multi-line layouts shift only the (already-padded) meta lines' columns. Regression coverage locks undeclared-assignment ReferenceError and both this-semantics in single-line and short-line layouts, with the existing line-number tests still passing.

CI note: the initial ubuntu failure was the 1000-lifetime-call test flaking under runner load (899/1000 admitted at the fixed assertion point) — unrelated to this diff (compiler prefix + runner tests only; macOS green on the same diff; the test passed on the previous head). The failed-job rerun is green, 8/8 checks pass. Recommend a follow-up hardening of that test to a deadline-bounded wait, as discussed.

PR #325, opened 2026-08-11
skill(workflow): port max-effort review mechanics into the code-review reference

ryanxiang7:
## What

Extends the bundled `workflow` skill's `references/code-review.md` with the mechanics that make a mature max-effort review practice work, so a TeamLeader loading the skill gets them without re-deriving:

- **Effort presets** (`quick` / `standard` / `max`): rounds and verifier-vote counts scale with the ask; the confirmation bar (≥2 settled votes, average ≥80) is constant, so lower effort yields fewer findings, never weaker ones.
- **Repeat-until-dry rounds** at max effort: find/verify pairs loop until two consecutive dry rounds (or the round cap), each round's prompts carrying every seen finding key; dedupe is against SEEN, not against the accepted list, with the non-convergence rationale documented.
- **Cumulative range mode** (`args.rangeMode`): resolve-first guidance for fuzzy targets ("PRs #N through head" → exact `base..head`), a seventh cross-change lens hunting bad interactions BETWEEN changes in the range (later change invalidating an earlier assumption, superseded leftovers, contradictory contracts, mid-range-only statements), and a range-scoped false-positive rule (intra-range deliberate replacement is not a finding).
- **Severity-grouped reporting**: finders estimate high/medium/low; the report leads with a verdict paragraph and groups issues by severity — with the explicit rule that severity labels never rescue low-confidence findings.

All previously landed hardening is preserved: positional lens-coverage accounting (now per-round), guidance-discovery failure handling, unverified-findings routing, citation discipline, bare-await gate semantics.

## Why

The recipe shipped in #319 covered one standard pass. Real review requests split into "quick sanity check", "normal PR review", and "audit everything since X" — the presets, dry-loop, and range mode are exactly the parts of that spectrum a script author gets wrong first (fixed passes miss the tail; naive dedupe never converges; cumulative ranges need the cross-change lens to be worth running as one diff).

## Notes

- Documentation-only bundled skill content; no runtime code touched.
- The embedded script uses the single top-level dialect (post-#324) and re-passes a syntax check under compiler-equivalent strict wrapping.
- Rush change file included (patch, bundled skill content).

Refs #319, #324

### DIFF
--- a/common/changes/@excitedjs/dreamux/skill-code-review-max-port_2026-08-11-00-00.json
+++ b/common/changes/@excitedjs/dreamux/skill-code-review-max-port_2026-08-11-00-00.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@excitedjs/dreamux",
+      "comment": "Extend the bundled workflow code-review reference with effort-level presets (quick/standard/max), repeat-until-dry find/verify rounds with seen-key deduplication, a cumulative range mode with a cross-change finder lens and range-scoped false-positive rule, finder severity estimates, and a severity-grouped report format. Documentation-only bundled skill content; no runtime behavior change.",
+      "type": "patch"
+    }
+  ],
+  "packageName": "@excitedjs/dreamux"
+}
--- a/common/changes/@excitedjs/dreamux/skill-code-review-max-port_2026-08-11-00-00.json
+++ b/common/changes/@excitedjs/dreamux/skill-code-review-max-port_2026-08-11-00-00.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@excitedjs/dreamux",
-      "comment": "Extend the bundled workflow code-review reference with effort-level presets (quick/standard/max) validated fail-loud, repeat-until-dry find/verify rounds with cross-lens seen-key deduplication, an in-script range resolver plus cross-change finder lens for cumulative range mode, all-conditions verifier prompts with distinct emphases, a required concrete failure scenario per finding, required enum severity feeding a severity-grouped report, a labeled plausible tier at max effort, and coverage facts distinguishing convergence from round-cap stops; add an ultracode orchestration techniques reference (archetypes, scouting, run chaining, worked compositions). Documentation-only bundled skill content; no runtime behavior change.",
+      "comment": "Rework the bundled workflow code-review reference into a method-first recipe (flow, verbatim scoring rubric, false-positive discipline, effort scaling, cumulative-range guidance) with a compact adaptation sketch instead of a paste-ready script, and add an ultracode orchestration techniques reference (archetypes, scouting, run chaining, compact worked compositions). Documentation-only bundled skill content; no runtime behavior change.",
       "type": "patch"
     }
   ],
--- a/common/changes/@excitedjs/dreamux/skill-code-review-max-port_2026-08-11-00-00.json
+++ b/common/changes/@excitedjs/dreamux/skill-code-review-max-port_2026-08-11-00-00.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@excitedjs/dreamux",
-      "comment": "Extend the bundled workflow code-review reference with effort-level presets (quick/standard/max), repeat-until-dry find/verify rounds with seen-key deduplication, a cumulative range mode with a cross-change finder lens and range-scoped false-positive rule, finder severity estimates, and a severity-grouped report format. Documentation-only bundled skill content; no runtime behavior change.",
+      "comment": "Extend the bundled workflow code-review reference with effort-level presets (quick/standard/max) validated fail-loud, repeat-until-dry find/verify rounds with cross-lens seen-key deduplication, an in-script range resolver plus cross-change finder lens for cumulative range mode, all-conditions verifier prompts with distinct emphases, required enum severity feeding a severity-grouped report, and coverage facts distinguishing convergence from round-cap stops; add an ultracode orchestration techniques reference (archetypes, scouting, run chaining, worked compositions). Documentation-only bundled skill content; no runtime behavior change.",
       "type": "patch"
     }
   ],
--- a/common/changes/@excitedjs/dreamux/skill-code-review-max-port_2026-08-11-00-00.json
+++ b/common/changes/@excitedjs/dreamux/skill-code-review-max-port_2026-08-11-00-00.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@excitedjs/dreamux",
-      "comment": "Extend the bundled workflow code-review reference with effort-level presets (quick/standard/max) validated fail-loud, repeat-until-dry find/verify rounds with cross-lens seen-key deduplication, an in-script range resolver plus cross-change finder lens for cumulative range mode, all-conditions verifier prompts with distinct emphases, required enum severity feeding a severity-grouped report, and coverage facts distinguishing convergence from round-cap stops; add an ultracode orchestration techniques reference (archetypes, scouting, run chaining, worked compositions). Documentation-only bundled skill content; no runtime behavior change.",
+      "comment": "Extend the bundled workflow code-review reference with effort-level presets (quick/standard/max) validated fail-loud, repeat-until-dry find/verify rounds with cross-lens seen-key deduplication, an in-script range resolver plus cross-change finder lens for cumulative range mode, all-conditions verifier prompts with distinct emphases, a required concrete failure scenario per finding, required enum severity feeding a severity-grouped report, a labeled plausible tier at max effort, and coverage facts distinguishing convergence from round-cap stops; add an ultracode orchestration techniques reference (archetypes, scouting, run chaining, worked compositions). Documentation-only bundled skill content; no runtime behavior change.",
       "type": "patch"
     }
   ],
--- a/packages/dreamux/skills/shared/workflow/SKILL.md
+++ b/packages/dreamux/skills/shared/workflow/SKILL.md
@@ -161,3 +161,4 @@ unavailable so orchestration stays deterministic.
 | --- | --- | --- |
 | Code review run | Carrying a change review (pull/merge request, git range, or working-tree diff) as one workflow run: eligibility gate, multi-lens finders, confidence-scored verification, one synthesized report. | [Code review](references/code-review.md) |
 | Orchestration and prompt patterns | Writing TeamMate prompts inside scripts, choosing `pipeline` versus `parallel`, and applying quality patterns: adversarial verify, judge panel, loop-until-dry, multi-modal sweep, completeness critic. | [Orchestration patterns](references/orchestration-patterns.md) |
+| Ultracode orchestration techniques | Shaping a whole task as workflows: the five archetypes (understand/design/review/research/migrate), scouting before fan-out, chaining runs with the TeamLeader in the loop, worked compositions, novel harness shapes. | [Ultracode techniques](references/ultracode.md) |
--- a/packages/dreamux/skills/shared/workflow/references/code-review.md
+++ b/packages/dreamux/skills/shared/workflow/references/code-review.md
@@ -2,16 +2,19 @@
 
 `../SKILL.md` owns the run tools and the script API. This reference owns one
 recipe: carrying a full code review as a single `workflow_run` — an eligibility
-gate, context collection, independent multi-lens finders, per-finding
-confidence scoring, and one synthesized report.
+gate, context collection, independent multi-lens finders repeated until dry,
+per-finding confidence scoring, and one severity-ordered synthesized report.
 
 ## Contents
 
 - [Stage design](#stage-design) — the five stages and their failure accounting
+- [Effort levels](#effort-levels) — quick / standard / max presets
+- [Cumulative range mode](#cumulative-range-mode) — reviewing many merged
+  changes as one diff
 - [Prompt discipline](#prompt-discipline) — output contract, rejection
   semantics, self-contained prompts
 - [TeamMate budget](#teammate-budget) — cost arithmetic against the run caps
-- [Script](#script) — the complete runnable module
+- [Script](#script) — the complete runnable script
 - [Invocation](#invocation) — call shape and result fields
 
 The review target is any change the Team workspace can read: a GitHub pull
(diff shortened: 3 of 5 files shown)


YourWildDad (changes requested):
Reviewed exact head `1d3b38110cd26b5f52ebeda8f27aa56fef30218c` against `next@08b8616e451be4e9d55ec491f340eac11458511f`.

I reproduced the following regressions by extracting the fenced script from this exact blob, compiling it with the production `compileWorkflowScript()`, and executing it with production-equivalent `agent` / `parallel` / `pipeline` semantics.

1. **P1 — Deduplicate the current round before verification (`packages/dreamux/skills/shared/workflow/references/code-review.md:362-373`).** `fresh = found.filter(...)` checks only the prior-round `seen` set, then updates `seen` after the full filter. The same finding emitted by all six lenses therefore starts 18 verifiers and returns six identical confirmed issues. Select fresh findings with a per-round key set (or insert into `seen` while traversing) before verification.

2. **P1 — Do not produce false-clean coverage/reporting (`code-review.md:335-336,354-369,407-419`).** Two reachable paths currently lie about completeness:
   - if a lens fails and the others return no findings, `coverage.complete` is correctly false but `report` says `No issues found. Checked all lenses across all rounds.`;
   - if round 3 still produces fresh findings, the loop exits at the cap without two dry rounds, yet `coverage.complete` is true.
   Generate the no-confirmed report from the actual coverage facts, and track whether the loop converged versus stopped at the cap with fresh findings.

3. **P1 — Preserve earlier results when a later round loses every finder (`code-review.md:354-361`).** The hard `return { error }` is now inside the multi-round loop. If round 1 confirms issues and every round-2 finder has a transient failure, the script discards all accumulated `confirmed`, `unverified`, and coverage data. Record the failed round and break to reporting; reserve the hard error for a run that never produced any review output.

4. **P1 — Validate the new control args before starting agents (`code-review.md:180-188,335-336,407-419`).** `{ maxRounds: -1 }` performs zero finder/verifier calls but returns `coverage.complete: true` and a clean report. Unknown `effort` values also silently select an unintended hybrid. Fail loudly unless `effort` is `quick|standard|max` and supplied `maxRounds` is a supported positive integer.

5. **P2 — Keep quick effort's evidence bar genuinely constant (`code-review.md:79-88,325-329`).** Quick uses `.slice(0, 2)`, removing the verifier that checks whether cited guidance/comments actually say what the finding claims. A guidance-based false positive can therefore be confirmed by quick even though standard checks the missing load-bearing fact. Two votes are fine, but every verifier must check all acceptance conditions; focus questions can add emphasis, not remove required evidence.

6. **P2 — Make cumulative-range resolution runnable or narrow the contract (`code-review.md:90-102,180-188,248-318,461-465`).** The prose says both cumulative-range differences are handled by `rangeMode`, including resolving a fuzzy target to one exact `base..head`, but the script has no resolver and passes the original fuzzy target independently to gate/context/every finder. Either add one resolve agent and thread its exact target through all stages, or require callers to supply an already-exact range and remove the claim that `rangeMode` handles resolution.

7. **P2 — Require the severity data used by the report contract (`code-review.md:207-226,343,421-434`).** `severity` is optional and accepts any string, while the report promises deterministic high/medium/low grouping. A schema-valid finding without severity reaches `confirmed`, leaving the reporter to guess. Add `severity` to `required` and constrain it with `enum: ['high', 'medium', 'low']`.

The embedded script otherwise compiles under the production compiler, the patch Rush declaration is correctly scoped, exact-head CI is 8/8 green, and the whole diff is public-safe.

ryanxiang7:
All seven review findings addressed at new head a28df5d, plus one scope addition requested by the operator.

**Fixes (mapping to the review's numbering):**
1. Fresh-finding collection now adds to `seen` inside the loop, so same-key findings from several lenses in the same round collapse to one finding and one verifier set.
2. The no-confirmed report is generated from the coverage facts (guidance discovery, per-round lens failures, round-cap stop, unverified count) — a run with gaps never claims 'checked all lenses'. `coverage.stoppedAtRoundCap` distinguishes cap-stop-with-fresh-findings from two-dry-round convergence and factors into `coverage.complete`.
3. A later round losing every finder records the gap in `roundFailures` and breaks to reporting with accumulated results; the hard error now requires that the run produced no review output at all (`seen.size === 0`).
4. `effort` and `maxRounds` are validated fail-loud before any agent starts (effort ∈ quick|standard|max; maxRounds integer 1..10).
5. Per your prescription: every verifier prompt carries ALL acceptance conditions (reproduces on changed lines / introduced not pre-existing / cited guidance actually says it) with a distinct per-verifier emphasis — quick's two votes reduce redundancy, not evidence. The effort-level prose now states this explicitly.
6. Range mode runs an in-script bare-awaited resolver whose exact resolved target threads into every later prompt; `resolvedRange` is returned. The prose now matches the runnable script.
7. `severity` is `required` with `enum: ['high','medium','low']`.

**Scope addition (operator request):** `references/ultracode.md` — the level above orchestration-patterns: five workflow archetypes (understand/design/review/research/migrate), scout-before-orchestrate, chaining runs with the TeamLeader between phases, worked compositions (including the two hard-won dedup rules), loop-until-count with a dry-pass exit, and novel harness shapes. SKILL.md References table routes to it; the Rush change comment covers both.

Embedded script re-passes the compiler-equivalent strict wrapping syntax check. Ready for re-review.

ryanxiang7:
Head moved: a28df5d → d5ab2ba, two additions from a source-fidelity audit against the original mechanisms (which the diff alone cannot reveal — flagged by the operator, verified by me against the source practice):

1. **Labeled PLAUSIBLE tier at max effort.** The original max semantics broaden coverage by *including uncertain findings, explicitly labeled* — the previous port silently dropped everything under 80 even at max, quietly narrowing what max means. Now: two settled votes averaging 50-79 → returned in `plausible`, reported under a "Plausible (unconfirmed)" heading with per-finding confidence, never mixed with confirmed. The ≥80 confirmation bar is untouched; the effort-level prose states the widened-reporting-not-relabeled-uncertainty rule.
2. **Required concrete `failureScenario` per finding.** The original finding contract requires each finding to state the concrete inputs/state → wrong outcome; finders are told a defect without a scenario is not reportable, and verifiers check the scenario as a fourth acceptance condition. This is the strongest false-positive filter in the source practice and was missing from the port.

Schema, prompts, tier routing, both return paths, report structure, invocation docs, and the Rush change comment updated coherently; script re-passes the strict-wrapping syntax check. Apologies for moving the head mid-review — please re-anchor on d5ab2ba.

YourWildDad (changes requested):
Reviewed exact head `d5ab2baaae7fd9477db578f3d5c037f942cda871` from commit-addressed GitHub blobs.

The original seven findings are materially fixed, and the Claude Code fidelity additions are directionally sound. The remaining findings below are Dreamux execution/owner-contract issues, not objections to the native Claude Code design.

1. **P1 — The documented fail-loud paths still settle the Workflow as `completed` (`packages/dreamux/skills/shared/workflow/references/code-review.md:211-225,330-331,456-460`).** These branches `return { error }`; Dreamux treats any normal return as a completed run (`runner.ts:69-74`). Invalid effort/maxRounds, failed range resolution, and initial all-finder failure therefore look successful to automated callers. Throw `Error` for the paths the reference calls fail-loud/hard-error; validate before agents as already intended.

2. **P1 — The confidence tiers classify on the rounded display value (`code-review.md:503-517`).** Votes `79,80,80` average `79.67`, are rounded to 80, and become confirmed; `49,50,50` similarly cross into plausible. Use the raw mean for tier boundaries and round only the reported `confidence` field.

3. **P1 — The two-dry convergence state still has false-complete/false-error edges (`code-review.md:428-465,522-531`).** (a) Round 1 can complete cleanly with zero findings, then an all-lens round-2 outage still hits the hard-error predicate because `seen/confirmed/unverified` are empty, discarding the completed round. Track completed/successful finder rounds directly. (b) `stoppedAtRoundCap` depends on `lastRoundHadFresh` and `MAX_ROUNDS > 1`; maxRounds=1 with fresh findings, or fresh/fresh/one-dry at cap, returns `complete:true` despite never reaching two dry rounds. Base cap status on exhausting the loop with `dry < 2`.

4. **P2 — The max-only plausible tier is not rendered as the promised report section when no issue is confirmed (`code-review.md:73-80,101-109,535-549`).** In a plausible-only run the reporter is skipped and `report` contains only a count plus “See ... plausible”; no title, path, confidence, or `failureScenario` is publishable. This also means the newly required failure scenario is not carried into the human report. Generate the same labeled Plausible section in this branch (or always use the reporter when plausible is non-empty), and require report items to preserve `failureScenario`.

5. **P2 — The new ultracode worked examples violate their own failure-plumbing owner (`packages/dreamux/skills/shared/workflow/references/ultracode.md:88-135,169-172`).** Both examples call `.filter(Boolean)` before positionally recording failed parallel inputs; in the exhaustive loop, two rounds where every finder fails are counted as two dry rounds and reported as convergence. The loop also has no explicit round/agent cap. Keep the high-level composition, but show minimal failed-position accounting and a bounded loop, or link to the single corrected owner instead of publishing another executable state machine. `orchestration-patterns.md:119-137` must not remain a contradictory owner with the old same-round dedupe bug.

6. **P2 — Multi-run chaining omits the actual Dreamux receipt/terminal boundary (`ultracode.md:68-84`).** `workflow_run` returns only `{run_id}` immediately; the structured result arrives later via terminal completion, with `workflow_status(run_id)` only for explicit recovery (`SKILL.md:19-24,36-38`). State that each chain step saves the run id, waits for terminal completion, then uses terminal `result` to build the next JSON args; otherwise a TeamLeader can feed `{run_id}` itself into the next run.

The embedded script compiles with the production compiler, same-round dedupe/range resolution/quick evidence/severity/failureScenario wiring are fixed, the Rush patch note covers both references, CI is 8/8 green, and the exact diff is public-safe.

ryanxiang7:
Round-3 findings all addressed at head 00b19ed:

1. **Fail-loud paths throw**: missing target, invalid effort/maxRounds, failed range resolution, and a finder outage with zero completed rounds now `throw`, settling the run `failed` instead of `completed`-with-error-field.
2. **Raw-mean tiering**: the aggregation stage computes a `tier` from the unrounded vote mean ('confirmed' ≥80, 'plausible' ≥50 at max, else rejected; <2 settled votes → unverified); `confidence` is rounded for display only. 79.67 stays plausible; 49.67 stays rejected.
3. **Dry-state edges**: hard error requires `completedFinderRounds === 0` (a clean round 1 followed by a round-2 outage keeps its results); `stoppedAtRoundCap = dry < 2` — convergence means two dry rounds, period, so maxRounds=1-with-findings and fresh/fresh/one-dry-at-cap both record incomplete coverage.
4. **Plausible-only reporting**: the reporter now runs whenever confirmed OR plausible findings exist, rendering the labeled Plausible section with per-item title/citation/failureScenario/confidence; the coverage-fact fallback applies only when both are empty. Report items (confirmed and plausible) carry the failure scenario.
5. **Example/owner convergence**: both ultracode worked examples account failed positions before `.filter(Boolean)`, treat an all-finder outage as an outage (never a dry round), and run under an explicit round cap with a converged/cap exit fact; the orchestration-patterns loop-until-dry owner snippet is corrected identically (in-loop seen.add, outage break, bounded loop), so the two owners now agree, with code-review.md named as the complete accounting contract.
6. **Receipt/terminal boundary**: the chaining section leads with `workflow_run` → `{ run_id }` as a durable acceptance receipt, result arriving via the pushed terminal completion, `workflow_status` for explicit recovery only.

Embedded script re-passes the strict-wrapping syntax check; CI running. Ready for re-review at 00b19ed.

YourWildDad (changes requested):
Reviewed exact head `00b19ed388258e7c953375b2a003247a912603c6` from commit-addressed GitHub blobs.

The prior six findings are fixed, and the Claude Code fidelity mechanics remain sound. Four Dreamux-facing contract issues remain:

1. **P1 — Quick and standard reviews can never report complete coverage (`packages/dreamux/skills/shared/workflow/references/code-review.md:95-105,222-227,532-542`).** Those presets intentionally run one round, but `stoppedAtRoundCap = dry < 2` applies max's two-dry convergence rule to every effort. A normal standard review whose six lenses all succeed and return no findings deterministically returns `stoppedAtRoundCap:true`, `coverage.complete:false`, and “stopped before two-dry-round convergence.” Define the required convergence by preset: quick/standard complete after their successful configured single pass; max/repeat mode requires two dry rounds. Keep outage/actual cap facts separate.

2. **P1 — The resolver still does not enforce the exact-range contract (`code-review.md:308-338`).** The prompt asks for `<baseSha>..<headSha>`, but the script accepts any non-empty string. A schema-valid `{ range: 'changes #312 through head' }` is threaded to every downstream agent and may return `coverage.complete:true`, violating the explicit promise that no agent receives the fuzzy target. Add a minimal local check for a non-empty `left..right` shape and throw otherwise; no general git ref parser is needed.

3. **P2 — A plausible finding can be displayed as `confidence: 80` (`code-review.md:505-518,560-573`).** Tiering now correctly uses the raw mean, but the public field is still `Math.round(average)`. Votes `79,80,80` yield raw 79.67, tier `plausible`, and displayed confidence `80`; the report contract simultaneously defines plausible as 50–79. Preserve one decimal (or use a tier-safe display rule) so the user-visible number does not contradict the tier.

4. **P2 — The new recovery example uses the wrong MCP call shape (`packages/dreamux/skills/shared/workflow/references/ultracode.md:80-84`).** It says `workflow_status(run_id)`, while the actual tool requires `workflow_status({ run_id })` (`SKILL.md:23-24`). This line is intended as actionable multi-run guidance and is likely to be copied literally. Keep the no-polling rule, but use the object argument.

I am not blocking on defensive validation of every JSON primitive or on adding a new test abstraction. The six requested fixes pass the production-equivalent regression matrix: fail-loud paths throw, raw means determine tiers, clean-then-outage preserves results, duplicate findings collapse, plausible-only invokes the reporter with failure scenarios, range output is threaded downstream, and the worked loops now record outages and caps. Exact-head CI is 8/8 green, the Rush patch note covers all references, and the full diff is public-safe.

ryanxiang7:
Round-4 findings addressed at head 4cdb68e:

1. **Convergence scoped to multi-round runs**: `stoppedAtRoundCap = MAX_ROUNDS > 1 && dry < 2`. A clean single-pass standard run is `complete: true` again; the two-dry requirement binds only where rounds can repeat. Effort-level prose states the rule.
2. **Resolver output shape-validated**: the resolved range must match `/^\S+\.\.\S+$/` or the run throws — a schema-valid fuzzy sentence can no longer masquerade as an exact range.
3. **Tier-safe display**: `confidence` is kept at one decimal (`Math.round(average * 10) / 10`); with integer votes the one-decimal display never crosses a tier boundary, so a plausible 79.67 shows 79.7 and the human-visible 50-79 band holds.
4. **MCP call shape**: `workflow_status({ run_id })` in the chaining reference.

Acknowledged the explicitly de-scoped items (no blanket primitive validation, no mandatory test abstraction, loop-until-count cap as non-blocking advice). Script re-passes the strict-wrapping syntax check; CI running.

YourWildDad (changes requested):
Reviewed exact head `4cdb68eda4966c88463e37f47545b00847b43e58` from commit-addressed GitHub blobs.

The four requested fixes pass the production-equivalent regression matrix, and all prior fixes remain intact. Two contract gaps remain:

1. **P2 — The “exact range” check still accepts multiple separators (`packages/dreamux/skills/shared/workflow/references/code-review.md:337-343`).** `/^\S+\.\.\S+$/` accepts `abc..def..ghi` because the right side can consume another `..`. That schema-valid resolver output is then presented to every finder as the unified exact target, but it is not one two-endpoint range and can be rejected or interpreted inconsistently by git commands. Validate exactly one `..` separator with two non-empty endpoints (while deciding explicitly whether valid `...` syntax is allowed), or return separate base/head fields and construct the range locally.

2. **P2 — The new exhaustive composition still drops verifier failures before accounting (`packages/dreamux/skills/shared/workflow/references/ultracode.md:127-145,195-198`).** Finder failures are now recorded correctly, but `accepted.push(...judged.filter(Boolean)...` silently drops a `null` pipeline item. A fresh finding whose verifier fails disappears, and the example can report convergence with no accepted item and no unverified coverage gap, contradicting this same reference's positional-accounting rule. Record null `judged[i]` entries against `fresh[i]` as unverified before filtering; the sketch can stay minimal.

I am not re-raising accepted non-blockers: `main...feature` is valid git range syntax, the outage/cap double flag is redundant but honest, and loop-until-count overflow remains optional follow-up. Single-pass coverage, raw tiering/display, receipt/terminal shape, fail-loud errors, plausible reporting, owner alignment, CI 8/8, Rush note, and public safety all pass at this head.

ryanxiang7:
Round-5 findings addressed at head 54c2621:

1. **Exact two-endpoint range validation**: the resolver output is split on `..` and must yield exactly two parts, each non-empty, not starting/ending with a dot, and whitespace-free — `v1.2..v1.3` passes, `a..b..c` and `main...feature` are rejected. The three-dot question is decided explicitly: the resolver prompt now instructs normalizing `a...b` to the equivalent `<mergeBase>..<head>` two-dot range, and the prose documents the validated output form. Truth-table checked: `a..b` ✓, dotted tags ✓, `a..b..c` ✗, `main...feature` ✗, one-sided/whitespace forms ✗.
2. **Verifier-outage accounting in the exhaustive example**: `judged` is consumed positionally against `fresh` — a null verdict routes the finding into `unverified` instead of vanishing, the exit comment reports roundFailures + unverified + converged together, and the load-bearing details list now names the rule (a run must not display converged while silently missing verdicts).

Script re-passes the strict-wrapping syntax check; CI running.

YourWildDad (changes requested):
Reviewed exact head `54c262130e915c321a437b6d109554ac577deec0`.

One previously requested core fix is not yet closed:

**P2 — The verifier-outage example checks only the final pipeline value (`packages/dreamux/skills/shared/workflow/references/ultracode.md:144-147`).** Dreamux passes each stage's returned value to the next stage; a normal failed `agent()` resolves to `null`, it does not throw. Thus `verifyStage` can return `null`, then `aggregateStage(null, finding)` can return a truthy `{ accepted: true }`, and the final `judged[i]` enters `accepted` instead of `unverified`. This directly violates the new line 164-166 promise that missing verdicts are positionally routed to `unverified`. Short-circuit before aggregation (for example, have the aggregate wrapper return `null` when its input is `null`) or split the two stages so the null positions are recorded before aggregation.

This is the only remaining finding. The exact-range truth table and all prior fixes pass; CI is 8/8 green, Rush/KB/public safety pass, and no new defensive edge cases are being requested.

ryanxiang7:
Direction change at head 97ea5ab, per operator: a skill teaches an approach and the model writes the script for the actual situation — it does not ship a paste-ready executable. (For reference: the original Claude Code /code-review skill is pure instructions, and the built-in Workflow doc's canonical example is a ~20-line structural sketch. Porting those as a certified 400-line script was my over-engineering, and it created the entire production-hardening review surface of the last four rounds.)

The references are now method-first (473 lines total across the three files, down from 1015):
- **code-review.md** (143 lines): the flow, the verbatim 0-100 rubric, the false-positive discipline, effort scaling, cumulative-range guidance (resolve-first + cross-change lens), a failure-semantics reminder, and a compact adaptation sketch explicitly marked as a shape, not a template.
- **ultracode.md**: worked examples reverted to compact source-style sketches; the one hard-won convergence rule (dedupe vs SEEN, not accepted) stays in prose; failure-plumbing deferred to its owner.
- **orchestration-patterns.md**: loop snippet reverted to the compact source form.

This moots the outstanding round-6 finding (the verifyStage/aggregateStage null-propagation was in example machinery that no longer exists) and the certified-executable review dimension generally. @ reviewer: please re-anchor the final gate on 97ea5ab with the method-doc bar — accuracy of the described approach, rubric fidelity, owner boundaries, public safety — rather than executable certification. CI running.

YourWildDad (changes requested):
Reviewed exact head `97ea5abb8b36c01f8277665cc9a41d5e24821887` as method documentation, not as a production script.

The scope correction is sound: the references are compact, explicitly non-template, owner boundaries are clear, the rubric/false-positive/effort/range/failure semantics are accurate, CI is 8/8 green, and the diff is public-safe. Two small method-level omissions remain:

1. **P2 — Preserve the concrete failure-scenario discipline (`packages/dreamux/skills/shared/workflow/references/code-review.md:24-52,61-69`).** The compression removed the Claude Code mechanism that every finding states the specific input/state and wrong outcome, and that verifiers check whether that scenario actually reproduces. Without it, vague “may race/may break” concerns can still be scored and reported. Add one method sentence: a finding without a concrete failure scenario is not reportable, and every verifier must validate the scenario. No schema or executable template is needed.

2. **P2 — Make channel publishing conditional (`code-review.md:53-59`).** The shared workflow skill may run without a channel reply tool, but the Report step currently says to post through it unconditionally. Restore the existing owner truth: “when a channel reply tool is available”; otherwise return the report or use the confirmed follow-up `send`/forge path.

These are the only remaining findings. Do not restore the paste-ready script or production-hardening surface.

ryanxiang7:
Both method sentences restored at head 07b180f: (1) the failure-scenario discipline is back in the flow — every finding states concrete inputs/state → wrong outcome, no scenario means not reportable, and verifiers check the scenario as part of scoring; (2) report posting conditions the channel reply tool on availability, with the return-to-caller and confirmed send/forge paths as alternatives. Method prose only — no schema or template restored. CI green; ready for the text-regression pass and exact-head APPROVE.

YourWildDad (approved):
Reviewed exact head `07b180fee38d658c90ea24fb6fd6ad0c857f6b6e`.

APPROVE. The method-document scope correction is sound, and the final prose delta closes both remaining omissions: Find now requires a concrete failure scenario and Verify checks it; report publishing uses a channel reply tool only when available and preserves caller / confirmed follow-up alternatives. The delta does not restore the paste-ready script or add new abstractions. Three resident reviewers approved this exact head, CI is 8/8 green, the Rush note and links are valid, and the diff is public-safe.

PR #333, opened 2026-08-13
refactor(mcp): adopt official protocol server

YourWildDad:
## Summary

- replace the five hand-written stdio protocol shims with one official `@modelcontextprotocol/server@2.0.0` runner
- pin protocol support to `2026-07-28`, `2025-11-25`, and `2025-06-18` while preserving scoped process and admin authorization boundaries
- publish canonical tool metadata, schemas, structured results, and method-plus-code public error projection
- return ordinary successful objects only as `structuredContent` with `content: []` to avoid duplicate model context
- restore operation-local no-poll/wait-for-push reminders as a single text block only for accepted Team, TeamMate, and workflow submissions; reminders never enter `structuredContent`
- make Channel provider catalogs fail loud and align built-in Feishu live/sessionless result contracts
- update current-state architecture references and add breaking Rush change files for dreamux, dreamux-types, and feishu-channel

## Validation

- CI run `31725865786`: all Ubuntu and macOS Rush update/build/smoke/typecheck/lint/test jobs passed
- Rush change declaration, knowledge-base, shellcheck, commit-metadata, and full-history gitleaks gates passed
- local Dreamux run: 86 test files passed, 968 tests passed, 3 opt-in skips; one unrelated `team-dissolve-contract` fixture teardown hit `ENOTEMPTY` while removing its temporary `.git` directory
- local built CLI smoke passed
- Codex `0.147.0` live suite passed, including official Feishu MCP discovery and an actual Feishu MCP reply call through the model gate
- Claude Code built MCP handshake passed; the real-model suite remains opt-in
- `.agents/scripts/check.sh`: 94 files reachable
- exact-head architecture, lifecycle, and complexity resident reviews approved `1596746f9cadd76fd7a33123a951fefd52b606be`

## Alpha

Published from exact head `1596746f9cadd76fd7a33123a951fefd52b606be` by release run `31726189887`:

- `@excitedjs/dreamux@1.0.0-alpha.g1596746f9cad`
- `@excitedjs/dreamux-types@1.0.0-alpha.g1596746f9cad`
- `@excitedjs/feishu-channel@5.0.0-alpha.g1596746f9cad`


ryanxiang7 (approved):
Approving. Reviewed the full diff against the `next` base (merge-base `6204d71`, 53 files, +5030/-3598, head `1596746`).

## What I verified

- **SDK assumptions checked against the real artifact**, not from memory: unpacked `@modelcontextprotocol/server@2.0.0` and `@modelcontextprotocol/core@2.0.0` and confirmed `serveStdio(factory, { legacy, transport, onerror })`, `StdioServerHandle.close()`, `McpRequestContext.era`, `fromJsonSchema`, and the modern/legacy split of `supportedProtocolVersions` all match how `src/mcp/server.ts` uses them.
- **The `content: []` design holds on the wire.** `appendTextFallbackForNonObject` only appends a text block when `structuredContent` is a non-object (array/primitive/null), and `normalizeContentlessToolResult` only fills in `content: []` when `content` is absent. Since every projector returns a record, nothing is auto-appended on either era, so `1596746`'s de-duplication does what it claims.
- **Schema constraints are really enforced.** The default validator on Node is a bundled AJV 8.18, so `additionalProperties: false`, `anyOf`, union `type`, `pattern`, and `const` are all live. That is what makes "descriptor-bound scope cannot be overridden by model input" true for `dispatcher_id` / `team_id` / `caller_kind`, even though scope is spread before `mapped.params` in `callTool` — and `channel-mcp.test.ts` locks it with the `dispatcher_id: 'evil'` case.
- **Nested input whitelists were not lost** in the `collaboration-space-mcp` rewrite: the old hand-written `containerArgs` / `repoArgs` projections became equivalent closed nested schemas with the same required fields.
- **Feishu result contracts line up**: `listKnownChatBots` -> `readChatBots` returns `{ chat_id, known, trusted }`, `WireChatBot` is exactly `{ open_id, name? }`, and the sessionless path in `provider.ts` returns the identical shape — all consistent with the declared output schemas.
- **Built and ran the suite locally**: `rush build` and `rush test` both clean; the seven MCP test files pass 76/76.
- **Test coverage was not weakened.** Per the repo rule about not trusting a green run from rewritten tests, I diffed the case lists: `mcp-contract-whitelist` keeps all 14 cases, and `teammate-mcp` / `team-mcp` each gained cases (sanitized unmapped errors, empty projected `run_id`, failed receipts carrying no guidance). The line-count drop comes from replacing hand-rolled JSON-RPC framing with a real MCP client.

`mcp-protocol-conformance.test.ts` is a strong gate — both eras, counter-offer, unsupported-revision rejection, output-schema validation, cancellation after acceptance, stdin EOF shutdown, injected-transport start failure, and stdout non-contamination.

## Non-blocking notes

1. **The restored no-poll reminders may not reach a Claude Code caller.** Before this PR, `appendStructuredTaskDispatchSuccessReminder` put the reminder into **both** the text block and `structuredContent`. `1596746` correctly keeps structured results pure, so the reminder is now text-only. If a host surfaces only `structuredContent` to the model, the reminder lands nowhere for that host, and the guidance survives solely through the role prompt. That is probably acceptable — `model-facing-writing.md` already says the general rule belongs in role prompts — but the operation-local reinforcement this commit deliberately restored will then only be effective on hosts that render text content. Worth confirming per supported runtime, since the same asymmetry applies to any future text-only signal.
2. `.agents/proposals/mcp-protocol-conformance.md` is registered under **Active Proposals** in `.agents/root.md`, but this PR is its implementation. `.agents/CONTRIBUTING.md` says an implemented proposal moves to `archive/`, and this PR already archives five other proposals — this one looks like it was missed.
3. SDK-level output-schema failures return `isError: true` with the raw `Output validation error: ...` text, which bypasses the `SANITIZED_TOOL_ERROR` policy this module otherwise enforces. For side-effecting tools (`spawn`, `create`, `reply`) that also means the model sees an error for work Dreamux already accepted. The proposal acknowledges the risk and the submission schemas are test-locked, so this is a residual to watch rather than a defect.
4. `channelMcpServerDescriptors` still does `if (tools === undefined || tools.length === 0) continue;`, silently skipping a provider that publishes an empty catalog, while `validateChannelToolCatalog` and the `channel-mcp` CLI treat an empty catalog as fail-loud. Worth aligning the two postures.

On the `@excitedjs/feishu-channel` output-shape change (`reply` now returns `{ message_ids }` instead of `{ status, message, message_ids }`): I checked the old handlers and `status` was hard-coded `'ok'` in all three tools with failures thrown rather than reported, and no core code reads `status` / `message` — so this is a dead-field removal rather than a live contract break. It is correctly declared `major` in the rush change file; only an external consumer reading those two fields would need updating.

PR #335, opened 2026-08-14
feat(cron): remove manual run-now capability

### LINKED ISSUE #334: Solution review: remove cron run-now capability
# Remove Cron Run-Now

## Outcome

Retire the broken manual cron fire capability completely. Scheduled jobs continue
to fire only from their configured schedules.

## Change

- Remove `cron_run_now` from the cron MCP tool catalog and call mapping.
- Remove `scheduler.cron.run_now` from the admin method registry.
- Remove `SchedulerCommands.runNow`, `SchedulerService.runNow`, and Team scheduler
  forwarding. Leave no alias, replacement endpoint, test-only hook, or unused
  helper.
- Remove the manual-dispatch option and keep the existing timer behavior as the
  only dispatch path. Preserve `dispatchAdmitted` and its lifecycle-generation
  shutdown guard.
- Make the timer-only dispatch orchestration return `Promise<void>` so no
  manual-run status contract remains; keep submission outcomes internal to the
  scheduler's persistence and rearm decisions.
- Replace tests that invoked `runNow` with deterministic timer-driven tests.
  Retain coverage for defer-until-idle, held fires, delete/stop races, lazy runtime
  start, dispatch-error rearm, the queued-admission generation race, and
  same-job held-fire collapse after an update rearms the timer.
- Add retirement contracts: MCP rejects `cron_run_now` without forwarding an
  admin request, and the admin socket returns `UNKNOWN_METHOD` for
  `scheduler.cron.run_now`. Keep registration coverage for the four retained
  scheduler admin methods.
- Update current bundled skills, documentation, and the Rush breaking change note
  to expose only create/list/update/delete.

## Unchanged

- Cron job schema, persistence, timing, scheduler ownership, and normal timer
  semantics.
- The official-SDK MCP server architecture on `next`.

## Task-record organization

- Root task discovery at `.agents/tasks/README.md` and route this lineage through
  `mcp/scheduler/remove-cron-run-now`, with no repository-name layer.
- Keep a README index at each domain level. Accept slash-separated domain paths
  in the task initializer and reject a broken root-to-leaf index chain.
- Record the routing rule in its own current decision instead of rewriting the
  provenance of the earlier lean-task decision.

## Verification

- Focused scheduler, Team scheduler, cron MCP, skill, admin namespace, MCP
  whitelist, and protocol-conformance tests.
- Rush build, lint, and package tests; KB validation and `git diff --check`.
- A production-source scan confirming the removed symbols are absent.

YourWildDad:
## Summary

- remove the `cron_run_now` MCP tool, `scheduler.cron.run_now` admin method,
  scheduler command/service entry, Team forwarding, and manual-dispatch branches
- keep timer-driven scheduling as the only dispatch path, preserving held-fire,
  error-rearm, and lifecycle-generation stop-race behavior
- replace manual-run tests with deterministic fake-timer coverage and explicit
  MCP/admin retirement contracts
- update bundled skills, current documentation, the breaking Rush change, and
  capability-domain development-task records

## Why

The manual run-now path can wait for the same conversational runtime that is
currently awaiting the tool result. The capability is retired completely rather
than retaining a broken compatibility path or adding a replacement.

## Breaking change

Cron MCP now exposes only `cron_create`, `cron_list`, `cron_update`, and
`cron_delete`. Direct callers of `scheduler.cron.run_now` now receive
`UNKNOWN_METHOD`. Cron job formats and timer-driven scheduling are unchanged.

## Validation

- focused scheduler, Team scheduler, cron MCP, skill exposure, admin namespace,
  MCP whitelist, and protocol-conformance tests: 92 passed
- `rush rebuild --to @excitedjs/dreamux`
- `rush lint --to @excitedjs/dreamux`
- `rush typecheck:tests --to @excitedjs/dreamux`
- `rush test --to @excitedjs/dreamux`
- task initializer nested-domain exercise and task-record check
- `.agents/scripts/check.sh`
- `git diff --check origin/next`
- production and current-documentation symbol scans

Closes #334


ryanxiang7 (approved):
Approving. Reviewed the full diff against `next` (merge-base `6b8ec14`, 43 files, +1094/-288, head `426c4fe`).

## The premise checks out

I traced the deadlock the PR is built on rather than taking the description at face value: `runNow` -> `dispatchAdmitted` -> `dispatch` -> `deferUntilIdleAndSubmit` -> `runtime.waitIdle()`. The old `manual: true` flag only skipped the held-fire collapse check; it never bypassed the idle wait. So a `cron_run_now` call made by the conversational agent that owns the target scheduler waited on its own runtime going idle, which cannot happen while that runtime is blocked on the tool result — it would park until `MAX_DEFER_MS` (1h) and then report `missed`. That is a deadlock by construction, not a tuning problem, so removing the capability rather than patching it is the right call.

Removing it also had a low blast radius: pre-PR the only entry points were the MCP tool and the raw `scheduler.cron.run_now` admin method — no CLI command or script called it.

## Behavior preservation

The interesting part of this diff is what happens to the branches that were previously guarded by `!opts.manual`, and they all collapse correctly:

- held-fire collapse (`this.heldFires.has(jobId)`) becomes unconditional — same behavior the timer path always had;
- `rearmAfterDispatchError` becomes unconditional — same;
- the lifecycle-generation fence in `dispatchAdmitted` is preserved, only its `{ id, status }` return degrades to `void`;
- inside `deferUntilIdleAndSubmit` / `submitHeld`, every `return '<status>'` becomes a bare `return` with no change to `clearHeldFire` / `armMissed` / abort ordering. The two `return this.submitHeld(...)` -> `await this.submitHeld(...)` rewrites are in scopes with no `try`/`finally`, so they are equivalent.

The Team-side removal is also clean. The deleted `runNow` wrapper was the only path that entered through `deps.availability.admit` (the Team route-lifecycle gate) and then carried an hour-long wait inside it — which is exactly what its comment said. The timer path never used that gate: `SchedulerService.admit` for a Team is `deps.admitOperation` (Dispatcher admission), while `availability.admit` wraps only the short mutations and the final `scheduledInput` submit. So dropping the wrapper removes a special case without leaving any long wait under the Team lock.

## Tests

Per the repo rule about not trusting a green run from rewritten tests, I diffed the case lists rather than the assertions:

- `scheduler.test.ts`: **no cases removed**, and three added that target precisely the branches which lost their `!manual` guard — transient-dispatch-error re-arm, the post-`stop()` generation drop, and same-job held-fire collapse. They drive the scheduler through its public API with fake timers and a pass-through `CapturedAdmissions` gate rather than reaching into private methods, so they exercise the real admission path.
- `team-scheduler.test.ts`: identical case list; the churn is in-body rewiring.
- Retirement is contract-locked at every layer it existed on: `tools/list` shape, `tools/call` rejection with no admin forwarding, `adminMethods` absence plus the `RETIRED_METHODS` list, and `not.toContain('cron_run_now')` in both bundled skills.

Locally: `rush build` and `rush test` clean, the six scheduler/cron/skill/namespace files pass 79/79, `.agents/scripts/check.sh` reports 104 files reachable, and `init_task.py check --domain mcp/scheduler --slug remove-cron-run-now` passes.

I also grepped the whole tree for `run_now` / `runNow` / `run-now`: nothing left in production source or current documentation — only archived proposals (correctly historical) and the new task records.

## Non-blocking notes

1. **Two changes in one PR.** The capability-domain task routing work — the new `capability-domain-task-routing.md` ADR, re-rooting `.agents/tasks/` away from the fixed `dreamux/` segment, and `init_task.py` (+92/-38) — is a workflow/tooling change riding along with a product removal. It is coherent and properly recorded, and I understand this task is the first consumer of the new layout, but it would be easier to revert or bisect on its own. The `assert_domain_chain` rewrite is actually a strengthening of `check` (it now validates the whole root-to-leaf link chain, where the old version only checked the root index), so no concern about the change itself.
2. **Breaking-note form differs from #333.** That change file used `BREAKING: Review: … no rebuild is needed`; this one leads with `BREAKING:` and states "No rebuild is required" without a `Review:` marker. Root `CLAUDE.md` scopes the `Review:` clause to a same-shape semantic change that retains its state version, which arguably does not apply here since no persisted state is involved — and `rush change --verify` only checks that a file exists. Flagging it as a consistency question, not a violation.
3. **Follow-up from #333:** `.agents/proposals/mcp-protocol-conformance.md` is still listed under Active Proposals in `.agents/root.md` even though #333 has merged. This PR demonstrates the convention working correctly — it archives `scheduled-tasks.md` and files `remove-cron-run-now-mcp.md` under `archive/proposals/` — so the MCP one looks like a leftover worth sweeping up separately.

PR #336, opened 2026-08-15
test: isolate collaboration worktree fixture

YourWildDad:
## Summary

- isolate the collaboration-space provisioning test from the developer's real Git checkout
- initialize the test source repository inside the existing temporary fixture root
- prevent repeated test runs from leaving prunable worktree registrations in the host repository

## Validation

- `node common/scripts/install-run-rush.js lint --to @excitedjs/dreamux`
- `packages/dreamux/node_modules/.bin/vitest run tests/dispatcher-collaboration-space.test.ts`
- `packages/dreamux/node_modules/.bin/tsc -p packages/dreamux/tsconfig.tests.json --noEmit`
- repeated the focused test twice and confirmed the host repository worktree, matching-branch, and prunable counts remained unchanged

## Limitations

- The full Rush test run reached the changed test successfully, but two unrelated live Codex tests timed out while reconnecting to the external service.

### DIFF
--- a/common/changes/@excitedjs/dreamux/dreamux-worktree-residue-audit_2026-08-15-05-01.json
+++ b/common/changes/@excitedjs/dreamux/dreamux-worktree-residue-audit_2026-08-15-05-01.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Isolate the collaboration provisioning test Git fixture so repeated runs cannot leave stale worktree registrations in the host checkout.",
+      "type": "none",
+      "packageName": "@excitedjs/dreamux"
+    }
+  ],
+  "packageName": "@excitedjs/dreamux",
+  "email": "15624809+YourWildDad@users.noreply.github.com"
+}
--- a/packages/dreamux/tests/dispatcher-collaboration-space.test.ts
+++ b/packages/dreamux/tests/dispatcher-collaboration-space.test.ts
@@ -1,6 +1,14 @@
+import { execFile } from 'node:child_process';
 import { afterEach, beforeEach, describe, expect, it, vi } from 'vitest';
-import { mkdirSync, mkdtempSync, realpathSync, rmSync } from 'node:fs';
+import {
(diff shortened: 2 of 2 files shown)


ryanxiang7 (approved):
Reviewed the diff against the worktree provisioning path — LGTM.

**Why the change is right:** `ProvisionManager` runs `git worktree add -b dreamux/<slug> <path> HEAD` in the source repo (`packages/dreamux/src/service/worktree/manager.ts:109-114`), and the managed worktree lands under the dispatcher workspace (`<workspace>/.workspace/worktree/...`), which `afterEach` deletes via `rmSync(root, ...)`. With `repo = process.cwd()` that left both a prunable worktree registration *and* a stray `dreamux/*` branch in the developer's real checkout on every run. Pointing the fixture at an isolated repo under `root` fixes both, and since `worktree add` bases off `HEAD`, a single-commit repo is sufficient.

**Consistency:** the fixture mirrors the existing pattern in `packages/dreamux/tests/collaboration-space-repo-close.test.ts:70-82` (temp source repo, local `user.email`/`user.name`, README + init commit), and `HOME` is already redirected in `beforeEach`, so global git config can't leak in. This was also the last `process.cwd()`-as-repo site in the test suite.

**Non-blocking nits (fine to ignore or fold into a later cleanup):**
- `git init --initial-branch=main` diverges from the four sibling fixtures that use plain `git init -q`, and raises the floor to git >= 2.28. The branch name is irrelevant here since provisioning bases off `HEAD` — worth either dropping the flag or standardizing the other fixtures on it.
- This is now the fifth copy of the same temp-source-repo setup; a `createSourceRepo(root)` helper in `packages/dreamux/tests/helpers/` would be a nice follow-up.
- `realpathSync(sourceRepoPath)` is a no-op because `root` is already realpath'd — harmless, and matches the sibling test.

CI is green on both ubuntu-latest and macos-latest (full rush build/typecheck/lint/test), which covers the concern about the new git invocations being portable.

PR #339, opened 2026-08-16
fix(transcript): harden native read contracts

YourWildDad:
## Summary

- enforce an exact source allowlist so `transcript_path` can appear only on direct TeamMate spawn/send receipt surfaces
- pin direct receipt path nullability to established runtime-session association across submitted, duplicate, failed, ambiguous, and stopped admissions
- cap model-visible transcript tool names at 256 Unicode code points and align `truncated` with clipping in returned content only
- stop Codex pagination at confirmed transcript origin while retaining exact-budget continuation into older lineage segments
- repair stale current-source pointers and clarify the existing Dreamux Types breaking note

Builds on #338.

## Validation

- combined single-writer development + four-seat non-Codex review workflow: `run-08ea99ea-91c9-46c0-bda1-71937f0079e7`
- complete corrected-tree non-Codex xhigh review: `run-4c2aa8c3-9ae5-49ef-853c-ee1954d155c1`
- final scoped non-Codex xhigh review: `run-1ac4b860-df9d-47e2-9767-9bf0f3b9cf67`, zero findings and no partial coverage
- Dreamux Utils: 22/22 tests; source/test typecheck and lint passed
- Codex: focused transcript 27/27, full package 124/124; typecheck and lint passed
- Dreamux: focused contract suites 87/87; deterministic full suite 1064 passed, 5 skipped; source/test typecheck and lint passed
- Rush build and change verification against `next` passed
- task/KB validation and `git diff --check` passed
- repository artifact and task-added-line publication red-line scans returned zero matches

## Known limitations

- two opt-in Codex model-service gates remain externally unhealthy as recorded in #338; this follow-up does not change their app-server or protocol prerequisites
- canonical gitleaks was unavailable locally; targeted scans passed and CI must run the canonical full-history gate

## Upgrade note

No rebuild is required. This change adds enforcement and bounded transcript projection fixes without changing persisted Dreamux state, native rollout formats, cursor envelopes, or the public `last` query surface.


ryanxiang7 (approved):
Final independent review at exact head `747ea077ed52573d5b7688e61630c4371d97d481` (base `next`, merge-base `8ed949e`). Reviewed the exact diff and current source, not the PR summary. Locked-artifact SHA-256 for `requirement.md` and `technical-design/final.md` both match. **No blocker. Approved.**

Verified independently:

- **Static gate is real, not a paper gate.** `hitsInSource` scans the whole file with a global regex rather than line-by-line, so a second occurrence on the same line is still counted — the 4 expected hits in `teammate-mcp.ts` in fact require counting both occurrences on line 423. `productSourceFilesUnder` walks every `packages/**` path containing a `src` segment (7 product trees, 309 `.ts` files); no `.tsx/.mts/.js` product source exists and nothing under a `src/` tree is non-`.ts` except `.md`, so there is no current blind spot. The allowlist reflects the true direct-receipt surface rather than legalizing an existing leak: `teammateReceiptSchema` and `projectTeammateSubmission` are bound only to `spawn` and `send` (teammate-mcp.ts:302/312/385/391).
- **Codex pagination.** A reviewer raised a P1 claiming the deadline path loses parent-segment history under a multi-segment lineage. Rejected on evidence and withdrawn by that reviewer: `nextCursor` is gated on `cursorTarget` (reader.ts:140-154), which is null whenever `candidates` is empty, so `hasOlder` is load-bearing only in the non-empty-candidates domain, and the base commit emits the identical empty page with a null cursor for the same inputs. The `!parsed.sawCompleteMarker` special case (reader.ts:333-339) is untouched pre-existing code. In the domain where `hasOlder` does matter, every stop site is conservatively correct: `segmentOriginReached` implies the segment was read from byte 0 with every window line parsed, and it is required to be the last lineage segment before a null cursor is emitted, so exact record/byte-budget exhaustion with a parent still pending keeps its continuation.
- **`truncated`.** The rule lives once, in `budgetTranscriptTurns`; moving `truncated ||= bounded.truncated` inside the accepted branch stops an omitted candidate from setting the page flag, while the single-oversized-turn fallback still forces `true`. Provider and core pass it through verbatim.
- **Tool-name cap is consistent across all three layers** — same constant, same unit, same boundary: `clipCharacters` clips by code point (`[...input]`), `clipTurnToBudget`'s fallback is also code-point based, and core's `isBoundedString` verifies `[...value].length <= 256`. Clipping cannot split a surrogate pair, and core cannot fail-loud on a page the shared utility produced. The new core test pins code-point semantics with 257 astral characters rather than UTF-16 units.
- **Change files / KB / red line.** Three patch change files plus the amended still-pending `dreamux-types` note, which now names the `AgentRuntime.getLast()` removal — verified genuinely absent from source. `@excitedjs/dreamux-utils` already has a pending `major` change file, so the `patch`/`minor` question has no effect this cycle; the no-rebuild claims hold since no persisted format changed. Both stale KB Source entries are fixed; a full scan of `/packages/...` references across `.agents/domains` and `.agents/reference` shows no remaining dangling path and no within-block duplicate; `.agents/scripts/check.sh` passes. Full-diff red-line scan is clean, and `git diff --check` is clean.
- **No test weakened.** The only deletions in test files are import/signature/helper edits; `fake-team-runtime.ts` is a backward-compatible signature widening. The pre-existing codex test gained an assertion.
- Required CI is green on this exact head, and the full local suite reproduces (dreamux-utils 22/22, agent-runtime-codex 124/124, dreamux 1066 passed / 3 skipped live-Codex gates).

Non-blocking, recorded for follow-up rather than fixed here:

1. The gate matches literal text, so a hypothetical wholesale receipt passthrough (`return { ...receipt }`) would leak with no new literal. Accurate for AC1 as written; no such passthrough exists today.
2. The negative test feeds synthetic hits into the comparator, so it proves the comparison logic rather than the scanner; a fixture-tree test would close that.
3. `productSourceFilesUnder` admits files only on a literal `src` path segment and asserts no minimum coverage, so a future package laid out under `lib/` would be silently skipped.
4. A repo-wide invariant lives in a leaf package's suite, so a filtered `rush test --to` run never executes it. Full CI does.
5. The core-side cap tightens the provider contract (a third-party provider returning a name over 256 code points now hard-fails); risk judged negligible and `AgentRuntimeTranscriptBlock` documents no bound for `name`/`text`/`input`/`output` either, so this is a pre-existing documentation gap.
6. A fork whose parent segment holds no completed turn still yields one empty terminal page — pre-existing, unchanged by this PR, already recorded as deferred in `verification.md`; the README's framing reads slightly broader than what shipped.

PR #342, opened 2026-08-17
fix(claude-code): gate settlement on lifecycle terminality; parse top-level command_lifecycle

YourWildDad:
## Problem

The resident `claude` CLI emits `command_lifecycle` as a **top-level `type`**:

```json
{"type":"command_lifecycle","command_uuid":"...","state":"completed"}
```

But `parseLine` only recognized it as a `system` subtype:

```json
{"type":"system","subtype":"command_lifecycle",...}
```

Top-level envelopes fell through to `kind:'other'`. PR #338's settlement gate (`tryComplete`) waited for every accepted command to reach a terminal `command_lifecycle` state before settling the turn — but those envelopes were never parsed, so the gate never opened. The turn hung until the 600s idle reap even though the terminal `result` envelope had already arrived.

## Fix

**Settle on lifecycle terminality, not on counting results.** A Dreamux logical turn can span several CLI commands (a live steer is written into the turn already running), and the CLI answers them unpredictably — probed against a live 2.1.231 resident session:

- **Commands fold.** A message that arrives while the in-flight turn is inside a tool call is absorbed into that turn at the next query-loop boundary: the CLI issues `started` for each queued command, answers them together, and emits a **single** `result` (3 commands → 1 result, observed). A folded command's uuid never appears on any `result`, so counting one result per submitted command deadlocks.
- **`result.user_message_uuid` is not a completion ledger.** It is present only sometimes, is not reliably the first-submitted uuid of a fold, and is absent entirely on the `error_during_execution` artifact a `priority:'now'` interrupt produces.
- **`command_lifecycle` is the only 1:1 signal.** Every submitted uuid reaches a terminal state (`completed`/`cancelled`/`discarded`), folded commands included. Its ordering against `result` is not stable, so only eventual arrival may be assumed.

Consequently a turn settles when **every submitted command has reached a terminal lifecycle state and at least one `result` has been seen**, taking the last result seen (last-result-wins). A turn that did run a command keeps waiting for its result, since terminality does not imply the result has already been emitted.

Two escapes keep that gate from hanging:

- a build without `msg_lifecycle_v1` has no lifecycle signal at all and settles on its first `result`;
- a turn whose commands all ended without ever running (`cancelled`/`discarded`, or a failed steer write) can never be answered and fails immediately, because the idle deadline is not an acceptable backstop there — it reaps the resident child, and unrelated stream lines re-arm it.

**Parse the real wire shape.** `parseLine` now recognizes `command_lifecycle` as a top-level `type` (the resident CLI's actual shape), keeping the `system`-subtype path for backward compatibility with older streams and the test fixture.

**Cross-talk guard.** `result.user_message_uuid` survives solely to reject a result naming a command the pending turn never submitted (it belongs to an already-settled turn). A `result` that arrives with no pending turn is logged instead of dropped silently.

## Behavior change

A command reported `cancelled`/`discarded` no longer fails the turn: it is marked terminal and logged, because an interrupted command coexists with another that answers normally (that is exactly what a `priority:'now'` interrupt looks like). Only a turn where **no** command ever ran fails loudly, with the last abnormal reason.

## Test coverage

- `stream.test.ts`: top-level `command_lifecycle` parse, `system`-subtype back-compat, unrecognized state → `null`, `user_message_uuid` extraction.
- `rpc.test.ts`: a fold of three commands answered by one result; unfolded commands settle on the last result; result-before-lifecycle ordering; the interrupt artifact does not settle; a settled turn's late result does not settle the next turn; no-lifecycle build settles on first result; a discarded command does not hang or fail the turn; fail-loud when nothing ever ran.

126 tests pass, typecheck and lint clean.

Refs #341


ryanxiang7:
## Review: the parse fix is right, the settlement model is not

I verified the protocol claims against the actual CLI rather than against the fixture, and the two halves of this PR land differently.

**The `parseLine` half is correct.** I extracted the stream-json schema out of the `claude` 2.1.231 binary: `command_lifecycle` really is a top-level `type`, and its `command_uuid` is documented as *"the client-supplied uuid on the inbound message"*. The root-cause analysis in #341 and the `parseCommandLifecycle` refactor both hold. The 600s hang on a plain unsteered send is genuinely fixed.

**The settlement half rests on a protocol assumption that does not hold.** The PR body and the `deferSteeredResult` doc comment both say a priority steer "can make Claude Code close the interrupted run and drain the queued steer in the same stdout flush". I probed this against a live resident CLI (`--print --input-format stream-json --output-format stream-json --verbose --permission-mode bypassPermissions`, v2.1.231) with a real mid-stream injection: message A (no priority, long counting task), then message B (`priority:"next"`) written to stdin ~0.2 ms after A's first `content_block_delta`. Reproduced 3/3 runs:

```
queued(A) → started(A) → queued(B) → result(A) → completed(A) → started(B) → result(B) → completed(B)
```

- A is **not** interrupted. It streams for another ~4.3 s after the steer lands and completes fully (`stop_reason: end_turn`, `terminal_reason: completed`, full 1113-char output).
- There are **two** `result` envelopes, `num_turns: 1` on each — neither claims to cover both commands.
- They are **2127 / 2282 / 2752 ms apart** and in **different stdout chunks**, every run. Each `result` was the only complete line in its flush; across a 42-chunk run exactly one flush ever carried two complete lines, and those were `stream_event`s. The gap is floored by B's full model round-trip.

So the `setImmediate` fold window can never close over the two results. Two consequences follow, and both are reachable today:

### 1. A steered send silently loses its answer and returns the wrong text

`acceptTextInput` returns `active.runtimeTurn` to B's caller (runtime.ts), so B's sender is bound to A's turn. With this change: `result(A)` arrives → `deferSteeredResult` → one tick later the turn settles with **A's** text. B's sender receives A's answer. ~2.3 s later `result(B)` arrives, finds `pending === null`, and is dropped with a warning. B's actual answer is gone.

This misses the #341 acceptance criterion "并发 send 合并一条回推" — and it trades a loud hang for silent data loss, which is the worse failure mode under the repo's fail-loud posture.

### 2. A late `result` settles the *next* turn (cross-talk)

The `result` case only guards on `this.pending === null`. In the ~2.3 s window after the turn settles, the runtime considers itself idle (`runActiveTurn`'s `finally` clears `activeTurn`), so a new send goes down the `submitTurn` path and installs a fresh `pending`. The stale `result(B)` then settles turn C. **C's sender gets B's answer.** This is a new correctness bug — it cannot happen on `next`, where the turn never settles at all.

The existing test encodes the drop as expected behaviour (`logs a warning when a result arrives with no pending turn`), and the fold test feeds every chunk synchronously, so neither can catch this. That is the same "fixture models a protocol the CLI does not speak" trap #341 correctly called out about `fake-claude-stream.mjs`.

### The probe also hands you the fix

`result` carries **`user_message_uuid`, and it is exactly the client-supplied `uuid`** — `aaaaaaaa-…` for A and `bbbbbbbb-…` for B, every run. Full key list (23 keys, identical on both):

```
api_error_status, duration_api_ms, duration_ms, fast_mode_disabled_reason, fast_mode_state,
is_error, modelUsage, num_turns, permission_denials, request_sent_wall_ms, result, session_id,
stop_reason, subtype, terminal_reason, time_to_request_ms, total_cost_usd, ttft_ms,
ttft_stream_ms, type, usage, user_message_uuid, uuid
```

(There is no `command_uuid` on `result`; the envelope's own `uuid` is server-generated and `session_id` is shared across both turns.)

So `parseResult` can extract `user_message_uuid` and settlement can be attributed properly:

- Keep the set of command uuids submitted into this logical turn (the initial one plus each steer) — this is what the removed `commands` map was for, but keyed on results instead of lifecycle states.
- On `result`, look up `user_message_uuid`. Not in this turn's set → it belongs to an already-settled turn: warn and drop, **do not** settle the current turn. That closes the cross-talk hole.
- Settle once every submitted uuid has produced its `result` (last one wins, which `TurnAggregator` already does). That makes turn A wait for `result(B)` and hand B's sender the answer B asked for.
- Fall back to today's settle-on-first-result when `user_message_uuid` is absent, so older CLIs and the fixture keep working.

Worth noting this is *not* a return to the removed gate: attributing on `result` avoids the two things that made `tryComplete` wrong. Lifecycle terminal states are unusable as a settlement gate — the probe shows `completed(A)` arrives **after** `result(A)` (by 2.2–137 ms) — and the `resultCount >= completedCommandCount` comparison would deadlock whenever the CLI folds several queued commands into one turn (the schema's `command_lifecycle` description explicitly describes folds). Keying on `user_message_uuid` sidesteps both.

### Smaller points

- **`cancelled` / `discarded` are now fully silent.** My probe never triggered either (7 commands, all `queued`→`started`→`completed`), but the schema is explicit: `discarded` = "the session ended with the command still queued", i.e. definitely never answered; `cancelled` = possibly unanswered, and the schema tells resenders to correlate rather than blindly resend. Dropping the old reject path is defensible, but these should at least log a warning — otherwise a steer the CLI threw away leaves no trace anywhere.
- **Dead branch.** `TurnAggregator.outcome()` cannot return `null` once a `result` has been accepted, so `else pending.reject(new Error('claude turn ended without a result'))` in the `result` case (and the mirror in `deferSteeredResult`) is unreachable.
- **`steered` never resets**, so a steer whose write failed still costs every later `result` an extra tick. Harmless, but the flag no longer means what its doc comment says.
- `steerWriteUnconfirmedError` → `ambiguous` is a genuine improvement — distinguishing "written but unconfirmed" from "capability undecided" is the right call, and `classifySteerFailure` maps it correctly.

### Suggestion

Land the `stream.ts` parse fix — it is correct and independently valuable. Rework the settlement half onto `user_message_uuid` attribution before merging, and replace the synchronous-flush fold test with one that injects the second `result` in a **separate** `onStdoutChunk` call after the first has already had a chance to settle; that is the shape the CLI actually produces.

Probe artifacts (raw chunk logs with per-chunk timestamps and byte counts, plus the driver script) are available if you want the primary data.

ryanxiang7:
Pushed `a19e2ca` onto this branch, reworking the settlement half against the probe data from my review above. The `stream.ts` top-level `command_lifecycle` parse is kept as-is — that half was correct.

**What changed.** Settlement is now attribution-based rather than timing-based. `parseResult` extracts `user_message_uuid` (the client-supplied `uuid` of the message a result answers — the only usable key, since `result` has no `command_uuid`, its own `uuid` is server-generated, and `session_id` is shared session-wide). `PendingTurn` keeps its submitted command uuids in submission order plus a per-uuid outcome map; the turn settles only once every submitted uuid has produced a result, and takes the last submission's outcome. A result whose uuid this turn never submitted is warned about and dropped instead of settling it. The `steered` / `settleImmediate` / `deferredOutcome` / `deferSteeredResult` machinery is gone.

This is deliberately not a new mechanism: it is the contract `finalizeSlotIfReady` in `@excitedjs/agent-runtime-codex` already implements — fold then wait, `accepted.at(-1)` wins, attribute-by-id-or-drop — keyed on `user_message_uuid` instead of a native turn id. `runtime.ts:284` already claimed alignment with the Codex active-slot semantics; that claim is now true at the rpc layer too.

**Waiting needs an exit.** A `cancelled`/`discarded` command, or a failed steer write, leaves the waited-on set with a warning (mirroring Codex's `recordTurnStartFailure`) rather than hanging the turn on a `result` that is not coming. Dropping one command never fails the turn — that part of this PR was right. Emptying the fold entirely does fail it, loudly, with the last drop reason. Letting that case fall through to the idle deadline would be worse than a hang: `reapOnTimeout` kills the resident child, and every inbound stream line re-arms the timer, so a healthy session could be reaped long after the turn became unanswerable.

**On the tests.** The fold test now delivers the second `result` in a **separate** `onStdoutChunk` call with the event loop turning in between, because that is what the CLI actually does — the previous version fed both synchronously, which is why it passed against a premise the wire does not honor. Added regressions for cross-talk (a settled turn's stale result must not settle the next turn), a missing `user_message_uuid` (degrades to settle-on-this-result, no hang), a discarded steer (no hang, no rejection), and the empty-fold rejection (asserts it does not wait on the idle reap and that `reapOnTimeout` is never called).

One existing test needed a fixture correction: `completeRpcCommand` in `runtime-activity.test.ts` did not echo `user_message_uuid`. Only the fixture's wire shape changed — the `resultText: 'final result'` fold assertion is untouched and still green. That assertion is load-bearing, and correcting the fixture makes it actually exercise the contract rather than pass by accident.

Verified independently of the implementing agent's report: `rush build`, `rush test` (no FAILURE section; the three SUCCESS-WITH-WARNINGS packages only write expected stderr on error paths), `rush lint`, both tsconfigs' `tsc --noEmit`, and `.agents/scripts/check.sh` — all clean. Package suite is 123 tests. The new regressions were also confirmed to fail against the previous `rpc.ts`, so they test the new behavior rather than having been written to match it.

The probed wire facts are recorded in `.agents/domains/provider-runtime.md`, since this contract has now been guessed wrong twice — once via the fixture, once via the fold premise. The change file description is updated to match the final behavior.

ryanxiang7:
Correction. `a19e2ca` was wrong, and `c269b35` replaces its gate. Flagging this prominently because the intermediate commit would have hung turns in normal use — do not read the middle of this branch as a working state.

**What I got wrong.** `a19e2ca` settled a turn once every submitted command uuid had produced its own `result`. That assumes one result per command. A second probe against the live CLI disproved it: **the CLI folds several queued commands into one turn emitting a single `result`.** Reproduced 3 commands → 1 result. It triggers whenever messages are queued while the in-flight turn is executing a **tool call** — at the next query-loop boundary the CLI issues `command_lifecycle started` for each queued command, absorbs them into the running turn, answers them together (`"APPLE\n\nOK_1\nOK_2"` in one run), and emits one result. Since a real agent turn makes tool calls, that is the common case. My first probe missed it only because its test turn was single-shot text with one loop boundary.

Three further facts rule out the obvious repairs:

- **A folded command's uuid never appears on any `result`.** Uuids cannot count completions.
- **The uuid a result does carry is not reliably the first-submitted one.** When two commands folded into a new turn, the single result carried the *later*, higher-priority one.
- **`priority:"now"` genuinely interrupts** (unlike `"next"`): the original command goes `cancelled` and the CLI emits a result with neither a `result` nor a `user_message_uuid` key. So "missing uuid ⇒ settle now" is wrong too — it would settle on an interrupt artifact.
- Also worth knowing: **`priority` is not required to fold.** A message with no priority field folded as well, so this is not steer-specific.

**The corrected gate.** `command_lifecycle` is the only signal that stays 1:1 with submitted commands — every uuid reaches a terminal state, folded ones included. So: settle when every submitted command is terminal **and** at least one result has been seen, taking the aggregator's last result. No ordering is assumed between lifecycle and result; the probe saw folded `completed` land both before and after it.

**This restores the skeleton of the gate this PR removed**, which is the part worth saying plainly: waiting for every command to reach a terminal state was right. The original failed for three specific reasons, none reintroduced — it never saw the envelopes at all (the top-level parse fix from this PR stays, and remains the correct root-cause fix), it additionally required one result per completed command (which folding deadlocks), and it rejected the turn on `cancelled`/`discarded` (wrong: an interrupted command coexists with another that answers normally).

Terminality with no result yet is a transient rather than a failure, so the turn keeps waiting once any command has actually run; only a turn where no command ever ran fails loud. `user_message_uuid` is kept solely as the cross-talk guard.

**Tests** cover the probed orderings as distinct cases, each delivered across separate `onStdoutChunk` calls with the event loop turning in between: the 3→1 fold settling at the driver's `completed` (asserting it is *not* settled earlier), the unfolded two-command case settling on the last rather than the first result, a fold whose result precedes the folded commands' `completed`, the interrupt artifact not settling the turn, cross-talk, the no-lifecycle degrade, and no-command-ever-ran failing without waiting on the reap. Verified against the previous implementation: the two fold cases fail there by **5s timeout**, reproducing the hang rather than merely asserting a different value.

One more fixture correction in `runtime-activity.test.ts`: `completeRpcCommand` now emits `result` *before* the command's `completed` (the order the probe observed for a command the CLI runs on its own) and uses the real top-level `command_lifecycle` shape. Assertions untouched.

Independently verified: `rush build`, `rush test` (0 FAILURE sections), `rush lint`, both tsconfigs, `.agents/scripts/check.sh` — all exit 0; package suite 126 tests. The probed wire facts, including folding, are recorded in `.agents/domains/provider-runtime.md`.

Standing caveat, stated rather than hidden: the gate assumes a turn's last terminal event does not precede the result answering it. Four probed scenarios hold to that. If it were ever violated the turn would settle with the previous result rather than hang.

YourWildDad:
Review of the settlement gate rewrite (c269b35). I went through the three asks; here's where I land.

## Ask 1 — the "last terminal ≥ result" assumption: one counterexample candidate

The gate is `all terminal ∧ sawResult → settle(last result)`. In the fold case this is exactly right (N commands, 1 result, uuid unreliable). In the **unfolded** case it can settle one result early:

1. A runs alone → `result(A)` arrives → A terminal.
2. B steers in *between turns* → B runs alone → B's `completed` lifecycle arrives in a chunk **before** `result(B)` (the terminal-before-result ordering the KB itself records).
3. Gate: all terminal ∧ sawResult(`result(A)`) → settle with A.
4. `result(B)` lands later → dropped as a late result. B's caller gets A's answer.

The KB's "ordering not stable" observation is what makes this reachable — but only if terminal-before-result occurs for a **solo** command, not just inside a fold. The fold tests and `keeps waiting when a command completed but its result has not arrived yet` both cover single-command or fold shapes; neither covers "unfolded N commands, last terminal before its result".

If the probe saw terminal-before-result only inside folds, this gate is sound as-is. If it can happen solo, the gate needs a way to tell "result for the last command" from "result for an earlier one" — `user_message_uuid` is the only on-wire candidate, and it's exactly the field that's unreliable in folds, so the distinction would need to be: uuid matches the *last* terminal command → settle; uuid matches an *earlier* terminal command → keep waiting; no uuid → current behavior.

**Question:** did the 2.1.231 probe see terminal-before-result on a command that ran alone, or only inside folds?

## Ask 2 — `ranAnyCommand`: correct

The split is right. `completed` means the CLI ran the command; a result may still be in flight (terminal-before-result), so waiting is correct. `cancelled`/`discarded`/failed-write means the command never ran; no result can exist for it, and with *zero* ran commands the turn is unanswerable — failing immediately is strictly better than the idle deadline, which would reap a healthy resident child and gets re-armed by unrelated stream lines. The "one ran, one didn't" middle case (`fails a multi-command turn only once the last unrun command is terminal`) settles on the ran command's result — the right call.

## Ask 3 — `completeRpcCommand`: correcting to the real protocol, with one nit

The three changes all match the KB wire facts: `user_message_uuid` is the CLI echoing the client uuid, result-before-`completed` is the observed solo-command order, and the top-level `command_lifecycle` shape is the resident CLI's actual emission. Assertions unchanged → this is fixture correction, not implementation accommodation. The helper now exercises the result-before-terminal path; the terminal-before-result path has its own dedicated test, so both orderings are covered.

Nit: the helper only emits the top-level shape now. The `system`-subtype shape is still parsed (backward compat) and has parser-level tests, but no runtime-activity coverage. Not blocking — the parser tests pin it — just noting the asymmetry.

## Other notes

- `deferSteeredResult` is gone, and with it the setImmediate same-chunk-only fold: the lifecycle gate subsumes it and handles cross-chunk results naturally. Good deletion.
- The cross-talk guard (result naming an unsubmitted uuid → warn + drop, with a dedicated test) closes the "settled turn's late result settles the next turn" hole.
- The `error_during_execution` interrupt artifact (no uuid, no `result` key) correctly does not settle — treating "missing uuid" as settle-now would have settled on interrupts.
- The no-`msg_lifecycle_v1` degrade (settle on first result) is the right escape for builds with no lifecycle signal at all.

Net: the gate is a strict improvement over my first-result model — it handles the fold (the case my model got wrong) and keeps the no-hang property. The only open question is Ask 1's solo-command terminal-before-result reachability; everything else I'd ship.

ryanxiang7 (approved):

PR #311, opened 2026-07-31
feat: trust Feishu groups and improve maintenance

YourWildDad:
## What changed

- reinterpret Feishu V3 `group.allow_chats` as trusted human groups for `allowlist` and `follow-user` while keeping P2P and `/introduce` sender-scoped
- fail closed on unknown inbound classifications and invalid group policies without changing the V3 state shape or public gate ABI
- refactor `dreamux-maintenance` into a concise seven-route entrypoint plus seven single-owner references
- add a staged managed-daemon self-upgrade SOP with changelog-driven migration review, private recovery ownership, restart-resume notification, exact marker handling, rollback, and original-Channel reporting
- synchronize public docs, KB ownership, contract tests, archived proposals, and Rush change notes

## Why

Selected Feishu groups need trusted-member delivery under the existing mention gate, while normal DMs, unlisted follow-user groups, and `/introduce` retain sender-level authorization. The maintenance guidance also needed progressive disclosure and an executable, fail-closed upgrade path instead of a monolithic skill.

## Impact

Existing non-empty `allow_chats` entries under `allowlist` or `follow-user` must be reviewed before deployment. `access.json` remains V3 and needs no rebuild. The maintenance refactor changes bundled model guidance and tests only; it adds no runtime, schema, API, or CLI behavior.

## Validation

- rebased onto current `origin/next` (`7bfca29`) and resolved the two overlapping KB sections without dropping either side
- Rush update, build, typecheck, typecheck:tests, lint, and full test passed
- final full test: 126 files (125 passed, 1 skipped), 1425 tests (1422 passed, 3 skipped)
- focused maintenance contracts: 3 files, 28 tests passed
- maintenance skill validator passed
- KB reachability: 88 files
- `rush change --verify` and `git diff --check` passed
- two fresh-context plan-only evaluations passed for lifecycle routing and non-private upgrade refusal without sentinel leakage
- exact final HEAD `587fc10a138b80e86d3b94867fa85bac8581393f` approved by architecture, correctness, and complexity reviewers

One initial full-test run hit the upstream Team identity HOME-isolation flake in `team-scheduler.test.ts`; the failed case passed alone and the subsequent full run passed.

PR #345, opened 2026-08-25
fix(release): keep the 0.x version line for pending change files

YourWildDad:
## What

Rewrite all 15 pending change files of type `major` to `minor`, keeping every `BREAKING:` note intact, and record the rule in `CLAUDE.md`: while Dreamux stays on the 0.x version line, change files must never use type `major`.

## Why

Rush bumps a 0.x package straight to 1.0.0 when a pending `major` change file is applied. The 2026-08-16 release run consumed the then-pending majors and pushed a version-bump commit that set `@excitedjs/dreamux` (and siblings) to 1.0.0 on `main` — a version state the project does not want yet. That release run failed before publishing, so 1.0.0 was never published to npm (`latest` is still 0.21.0); the bump commit is being discarded from `main` separately.

Without this fix, the next release run would reproduce the same unwanted 1.0.0 jump. With it, the next stable release is a normal 0.x minor bump (e.g. `@excitedjs/dreamux` 0.21.0 → 0.22.0), and the breaking-change documentation in the generated changelog is unchanged — the `BREAKING:`/`Review:` notes still lead each entry per the repository's changelog rules.

## Scope

- `common/changes/**`: 15 files, `"type": "major"` → `"type": "minor"` only; comments untouched.
- `CLAUDE.md`: one paragraph under *Changelog Responsibility* stating the 0.x no-major rule; the operator decides when the project leaves 0.x.

No source, test, or config changes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,20 +58,35 @@ jobs:
         run: git fetch origin "${{ github.base_ref }}" --depth=1
       - name: Verify Rush change files
         run: node common/scripts/install-run-rush.js change --verify --target-branch "origin/${{ github.base_ref }}" --no-fetch
-      # While Dreamux stays on the 0.x version line, a pending `major` change
-      # file would make Rush bump the package straight to 1.0.0 (how the
+      # While a package stays on the 0.x version line, a pending `major`
+      # change file would make Rush bump it straight to 1.0.0 (how the
       # 2026-08-16 release run produced an unwanted, never-published 1.0.0
-      # version state). Breaking changes are recorded as `minor` with a leading
-      # `BREAKING:` note instead (CLAUDE.md, Changelog Responsibility). The
-      # scan covers the full pending set, not the PR diff: the invariant is
-      # that no major is pending at release time, wherever it came from. The
-      # operator removes this gate when the project leaves 0.x.
+      # version state for the core packages). For 0.x packages, breaking
+      # changes are recorded as `minor` with a leading `BREAKING:` note
+      # instead (CLAUDE.md, Changelog Responsibility). Packages already past
+      # 1.0.0 (e.g. @excitedjs/feishu-channel) use real semver majors and are
+      # exempt. The scan covers the full pending set, not the PR diff: the
+      # invariant is that no 0.x major is pending at release time, wherever
+      # it came from. The operator lifts a package out of this gate by moving
+      # its version past 1.0.0.
       - name: Forbid major change files on the 0.x line
         run: |
           node -e '
             const fs = require("fs");
             const path = require("path");
-            const root = "common/changes";
+            const versions = new Map();
+            const collect = (dir) => {
+              for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
+                if (entry.name === "node_modules") continue;
+                const p = path.join(dir, entry.name);
+                if (entry.isDirectory()) collect(p);
+                else if (entry.name === "package.json") {
+                  const pkg = JSON.parse(fs.readFileSync(p, "utf8"));
+                  if (pkg.name && pkg.version) versions.set(pkg.name, pkg.version);
+                }
+              }
+            };
+            collect("packages");
             const bad = [];
             const walk = (dir) => {
               for (const entry of fs.readdirSync(dir, { withFileTypes: true })) {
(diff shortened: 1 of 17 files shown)


ryanxiang7:
## Review: REQUEST-CHANGES

The 0.x-policy direction is right, and the 12 change files for 0.x packages (dreamux, agent-runtime-*, dreamux-types, dreamux-utils) are correctly downgraded to `minor` with their BREAKING notes intact. But the sweep also hit `@excitedjs/feishu-channel`, which is not a 0.x package — and CI is red because of it.

### Blocker — CI red: the sweep broke the trusted-allow-chats release contract

`packages/dreamux/tests/feishu-allow-chats-release-contract.test.ts` fails on both platforms: it deliberately pins `common/changes/@excitedjs/feishu-channel/feishu-trusted-allow-chats_2026-07-31-15-22.json` to `type: "major"` (contract added in #311). This PR downgraded it to `minor`.

`@excitedjs/feishu-channel` is a published package at **4.1.1** (5.0.0 betas already exist on npm). For a 4.x package, a BREAKING authorization-semantics change must ship as a major — that is the entire point of the #311 pin. Downgrading it would release the breaking change as 4.2.0.

The same applies to the other two swept feishu-channel files (`mcp-protocol-conformance_2026-08-13-00-00.json`, `unified-teammate-lifecycle_2026-08-16-00-00.json`) — both are BREAKING changes to the same published 4.x package.

### Major — the CI gate is not scoped to 0.x packages

The new "Forbid major change files on the 0.x line" step scans **all** pending change files and forbids `major` unconditionally. Even after the feishu-channel files are restored to `major`, this gate would fail on them. The gate must be package-aware: only forbid `major` for packages whose own `package.json` version is `0.x`, so published `>=1.0.0` packages keep normal semver.

### Minor — CLAUDE.md wording

"While Dreamux stays on the 0.x version line, change files must never use type `major`" should be scoped to 0.x packages; as written it conflicts with the published 4.x package.

### Fix direction

1. Restore the three `@excitedjs/feishu-channel` change files to `"type": "major"` (the contract test enforces the trusted-allow-chats one; the other two are BREAKING changes to the same published 4.x package).
2. Scope the CI gate to packages whose `package.json` version is `0.x`.
3. Scope the CLAUDE.md rule wording to 0.x packages.
4. Keep the 12 downgrades for 0.x packages as-is — those are correct.

Everything else checks out: no source changes, BREAKING comment text preserved verbatim in all 15 files, and the gate script logic is otherwise sound (recursive scan, `::error` annotations, `existsSync` guard).

YourWildDad:
Review round addressed in abcb6908 — all three points taken:

1. **Blocker — feishu-channel majors restored.** `@excitedjs/feishu-channel` is a published 4.1.1 package with 5.0.0 prereleases on npm, so it is not on the 0.x line. The three feishu-channel change files are back to `major`, byte-identical to `next`; the issue #311 release-contract pin is honored again. `feishu-allow-chats-release-contract.test.ts` passes locally (7/7).
2. **Major — gate is now version-scoped.** The CI step resolves each pending change file's `packageName` against `packages/**/package.json` and only forbids `major` when the package's current version starts with `0.`. A pending major for an unknown package also fails (cannot prove it is past 0.x). Verified both ways locally: current tree passes with the feishu-channel majors present; flipping a 0.x change file to major fails with a file-anchored error.
3. **Minor — CLAUDE.md narrowed.** The rule now reads per-package: 0.x packages record breaking changes as `minor` with a leading `BREAKING:` note; packages past 1.0.0 (e.g. `@excitedjs/feishu-channel`) use real semver majors.

The other 12 downgrades are unchanged. PTAL @ryanxiang7

ryanxiang7 (approved):
Re-reviewed abcb6908 — all three findings from the previous review are resolved:

1. **Blocker fixed:** the three @excitedjs/feishu-channel change files (feishu-trusted-allow-chats, mcp-protocol-conformance, unified-teammate-lifecycle) are back to `type: major`, matching the published 4.1.1 package and the #311 contract test pin.
2. **CI gate scoped per package:** the gate now resolves each pending change file's packageName against `packages/**/package.json` and only blocks majors for 0.x packages; packages past 1.0.0 are exempt and unknown package names fail conservatively. I independently exercised the script on the PR head tree: passes with the three feishu-channel majors present, fails (with the correct file-level error) when a 0.x package or an unknown package carries a pending major.
3. **CLAUDE.md wording scoped per package**, with feishu-channel named as the past-1.0.0 example.

Also checked the full pending set on the PR head: exactly the three feishu-channel majors, everything else minor/patch/none — no stray majors. CI is green on ubuntu (build/test/lint, rush change declaration, gitleaks); macos is still running but this change is platform-independent.

Approving. Cleared to merge and run promote-next for 0.22.0.

PR #346, opened 2026-08-25
ci(release): push protected branches with the release deploy key

YourWildDad:
## Problem

Stable releases were blocked by branch protection: the `promote-next` push of `main` and the release version job's bump/next-sync pushes all run as `github-actions[bot]`, and GitHub offers no way to exempt the Actions app from a pull-request rule — classic protection silently drops it from `bypass_pull_request_allowances`, and rulesets reject it as a bypass actor (422). The operator had to lift protection by hand for every release; the 2026-08-16 release broke the main-ancestor-of-next invariant exactly this way when its next-sync push was rejected.

## Repository settings (already applied and verified)

- Read-write deploy key **release-pipeline** registered; private half stored as Actions secret **RELEASE_DEPLOY_KEY**.
- New repository ruleset **release-branch-pr-gate**: requires a PR with 1 approving review on `main` and `next` for every ordinary actor (including admins and the bot), with **deploy keys** as the only bypass actor.
- Classic protection on both branches keeps force-push/deletion blocks and `enforce_admins`, but no longer carries the PR rule (the ruleset owns it).
- Empirically verified on a scratch ruleset-covered branch: ordinary-credential push → rejected with GH013; deploy-key push → accepted.

## Workflow changes

- `promote-next.yml`: the `main` push authenticates over SSH with the deploy key (host keys pinned from the GitHub meta API; fail-loud when the secret is missing). Because a deploy-key push is an ordinary push, `on.push.main` fires release.yml by itself — the explicit `gh workflow run` dispatch step (the GITHUB_TOKEN anti-loop workaround) is removed, and permissions shrink to `contents: read`.
- `release.yml`: the version job's bump push and next-sync push use the same deploy-key SSH identity; the bump commit's `[skip ci]` footer becomes the load-bearing guard against retriggering. Top-level and version-job permissions shrink to `contents: read` (publish jobs keep their own `id-token: write`).
- `.agents/reference/release-process.md`: documents the two-layer protection model, the five-step promote shape, the push-triggered coupling, and updated failure/recovery rows. KB gate green.
- `type: none` change file for the paper trail; no package behavior or version change.

## Notes for review

- The next stable release is the first end-to-end exercise of the new push path; the deploy-key bypass itself is already proven against the ruleset.
- Key rotation = replace the deploy key + secret; no workflow change.

### DIFF
--- a/.agents/reference/release-process.md
+++ b/.agents/reference/release-process.md
@@ -33,14 +33,36 @@ debug dump; see [Promotion step](#4-promotion-stable) below.
 
 | Branch | Purpose | Lifetime | Protected |
 |---|---|---|---|
-| `main` | Stable npm `latest` head; one commit per release plus `[skip ci]` version bumps | permanent | push restricted to `github-actions[bot]` only |
-| `next` | Default PR base; beta channel head | permanent | squash-merge only via PR; bot allowed to fast-forward on release |
+| `main` | Stable npm `latest` head; one commit per release plus `[skip ci]` version bumps | permanent | ruleset PR gate; only the release deploy key pushes directly |
+| `next` | Default PR base; beta channel head | permanent | squash-merge only via PR (1 approval); release deploy key may fast-forward on release |
 | `feature/<slug>` | In-progress work for one PR | lives from branch-off to PR merge; PR squash merge deletes it | no |
 | `team/<slug>` | Reserved for TeamMate team integration branches; merged via PR | same as feature | no |
 
 PRs target `next`, never `main`. The GitHub repo default branch is
 `next`. Hotfixes, like every other change, land on `next` first and
 ship through the normal promote path.
+
+## Branch protection model
+
+Two layers guard `main` and `next`:
+
+- **Classic branch protection** on each branch blocks force pushes and
+  deletions and applies to administrators (`enforce_admins`). It carries
+  no pull-request rule.
+- **Repository ruleset `release-branch-pr-gate`** (targets
+  `refs/heads/main` and `refs/heads/next`) requires a pull request with
+  1 approving review for every ordinary actor — including
+  administrators and `github-actions[bot]` — and lists **deploy keys**
+  as its only bypass actor.
+
+The bypass exists because GitHub cannot exempt the Actions app from a
+pull-request rule (neither classic protection nor rulesets accept it as
+a bypass actor). The release pipeline therefore pushes over SSH with the
+repository's read-write deploy key `release-pipeline`; its private half
+lives in the Actions secret `RELEASE_DEPLOY_KEY`, and both pushing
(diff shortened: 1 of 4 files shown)


ryanxiang7 (approved):
Re-reviewed 877a847a — the long-term release-pipeline mechanism is sound. Verified end to end:

**Workflow changes**
- All three protected-branch pushes (promote-next ff of `main`, release.yml version-bump push, next re-sync) now go over SSH with the `release-pipeline` deploy key; `GITHUB_TOKEN` is reduced to `contents: read` everywhere. The publish job keeps its own `id-token: write`, so npm trusted publishing is unaffected.
- The deploy-key setup fails loudly on a missing `RELEASE_DEPLOY_KEY` (promote-next gates the push steps on it; release.yml's version job sets it up unconditionally so a release that cannot push dies before change files are applied).
- Host keys are pinned from the meta API instead of TOFU; the private key is chmod 600, written from an env var, never echoed.
- The removed dispatch step is correctly replaced by push-triggered coupling: deploy-key pushes are ordinary pushes and visible to `on.push` (only GITHUB_TOKEN-originated events are suppressed), and the `[skip ci]` footer on the bump commit is the load-bearing anti-loop guard for both the main push and the next re-sync. Manual dispatch remains the documented retry hook.
- gitleaks (full history) is green — the private half of the deploy key never entered the repo.

**Live repo settings (verified via API, not just the PR description)**
- Ruleset `release-branch-pr-gate`: active, targets `main`+`next`, requires 1 approving review, sole bypass actor is DeployKey.
- Classic protection on both branches: PR rule removed, force-push/delete blocks and enforce_admins retained.
- Deploy key `release-pipeline` registered with write access.

**Gates**
- Full CI on the head SHA: build/typecheck/lint/test green on ubuntu and macos, shellcheck, kb check, commit-metadata, gitleaks all green (manually dispatched — see below). The rush-change declaration job is pull_request-only and was skipped on the dispatch; I verified locally that the `type: none` change file satisfies both `rush change --verify` and the 0.x-major gate.
- KB link/orphan check also run locally on the PR head: clean.

**One anomaly to chase (not a merge blocker for this PR's code):** no `pull_request` workflow run was created when this PR was opened — zero check runs on the head SHA until I dispatched ci.yml by hand. If that reproduces on the next PR, the repo's PR gates are silently not running and the cause (ruleset interaction? how the head branch was pushed?) needs a fix.

**Residual risk, acknowledged by the SOP doc:** the deploy-key-push → `on.push.main` link has not been exercised end-to-end yet — the first post-merge promote-next is the live test. The failure mode is benign (release.yml simply doesn't fire; the doc's manual-dispatch retry hook covers it), and the alternative (keeping the GITHUB_TOKEN push) is now impossible under the ruleset anyway.

Approving. Cleared to merge.

PR #347, opened 2026-08-26
feat(channel): render conversations as live Feishu COT cards

YourWildDad:
## Summary

Render each channel-facing agent conversation — the dispatcher's own agent and TeamLeaders — as one live Feishu chain-of-thought card that updates in place. Team members publish the same neutral conversation events for Channel providers to consume, but Feishu COT deliberately renders only dispatcher and TeamLeader conversations; team-less dispatcher-spawned teammates publish nothing.

- **Core (neutral):** a display-only conversation projection covers the dispatcher agent, TeamLeaders, and Team members (scope union `{team_name: null, role: 'dispatcher'} | {team_name: string, role: 'team_leader' | 'team_member'}`). Every turn that enters conversation projection publishes exactly one terminal display fact (`completed` / `failed` / `stopped`) from its own submission settlement, even when a runtime folds several submissions into one completion; turns rejected by the participation predicate publish no display events. All projection entry points are fail-open: display can never change admission, settlement, completion delivery, or shutdown. Projected text is sanitized and bounded; buffers carry explicit caps with drop-newest behavior.
- **Feishu channel (presentation):** an inbound turn's card is pinned to its inbound message. A TeamLeader keeps one active card correlated by turn id and accepts activity only for its single admitted turn; its next card may anchor to the leader's latest same-target Reply receipt, with the team-group binding notification as the pre-inbound fallback. Two bounded lifecycle fences reject late submitted facts and fallback anchors: a leader-wide fence set by Team close and endpoint-scoped route fences set by unbind/replacement, matched against the anchor's authoritative accepting binding (not visible-target fallbacks) so one endpoint's teardown never disables the leader's other live endpoints; Team restart clears both kinds and a matching re-bind clears its route fence. Late receipts or notifications can never recreate or revive presentation state (full conversation-target identity checks: chat + target type + target key). Team-member events are explicitly ignored. Dispatchers keep independent per-chat, per-turn state, ignore foreign-channel origins, and never render internal (scheduled/completion/control) turns. COT I/O is deadline-bounded with settled-state reaping.
- **Runtime display fields:** failed tool results populate the neutral `error` field from provider result detail on both runtimes (`null` only when no detail exists).
- **Transport:** a stateless COT message surface over the official Feishu chain-of-thought endpoints (Lark SDK `^1.73.0`), plus a synchronous fail-open per-message creation observer on text sends.
- **Reaction supersession:** the automatic inbound received/in-progress reaction flow is removed — the COT card is the progress surface. The model-facing `react` tool and `addReaction` remain. This formally supersedes the issue #63 automatic-reaction tri-state contract (decision record: `.agents/decisions/feishu-cot-conversation-display.md`); the issue #63 **non-blocking inbound gate is unchanged** and its live coverage remains load-bearing.

## Validation

- Full workspace build, src+tests typecheck, and lint green.
- Full workspace tests: 151 files, 2073 passed, 4 skipped (pre-existing skips plus one deliberately dormant display-cleanup case with an active replacement assertion).
- Codex live gates: 7/7 real runs (no skip variable).
- Previously locked reaction suites are rewritten against the new contract, not deleted; automatic-reaction absence is positively asserted.
- Change files included for all six published packages (feishu-channel is a real major with every public removal named; 0.x packages use minor with `BREAKING:` notes per repository policy).
- `.agents` knowledge updated in the same PR (`.agents/scripts/check.sh` passes).

## Limitations / follow-ups (tracked in the task README)

- `collaboration-routing.ts` sits at the 700-line lint cap; split before further edits.
- Leader conversations keep the single-active-card model; per-turn leader presentations are a candidate refinement.


ryanxiang7 (changes requested):
Three review seats (channel, core, KB/contracts) went through this PR; I verified every finding below against the diff myself. The implementation quality is high — lifecycle tests, fail-open boundaries, and the #63 supersession chain are all in good shape. Two accuracy/robustness gaps need fixing before merge, both small.

## Must fix

**1. [major] feishu-channel change file under-reports the breaking surface** — `common/changes/@excitedjs/feishu-channel/feat-feishu-cot-cards_2026-08-25-22-10.json`

The BREAKING comment only names the two reaction constants, but the diff also removes:
- `FeishuBot.removeReaction` — a method on the publicly exported `FeishuBot` interface (`bot.ts`, old lines 33/76-77). Any integrator implementing `FeishuBot` fails to compile; the comment's "addReaction surface remain available" wording actively implies the reaction surface is otherwise intact.
- `FeishuChannelSession.state` public readonly field and the exported `FeishuChannelState` type (`feishu-channel.ts`, old lines 26/72).

The `major` bump and the 5.0.0 line are correct; the migration note must enumerate all three removals.

**2. [major] The "late receipts after Team close / unbind / route replacement can never recreate presentation state" invariant has a gap** — `packages/channel/feishu-channel/src/feishu-cot-adapter.ts:103-145`

The identity checks cover Reply receipts, binding notifications, old-card revival, and session close, but `onTurnSubmitted` with a presentable `channel_origin` has no fence: after `onTeamState` (which only interrupts the current card, line 351-355) or `onBindingRoute` (same, 359-373), a late submitted-with-origin event runs `ensureLeaderState` + `advanceAnchor(..., 'done')` and opens a brand-new card. The claim is committed in three places — decision record `.agents/decisions/feishu-cot-conversation-display.md:52`, `.agents/reference/channel-runtime.md:454`, requirement `.agents/tasks/channel/feishu-cot-conversation-cards/requirement.md:24` — so the docs and the code currently disagree.

Exposure is narrow (needs core to emit submitted-with-origin after close/unbind; the matching settled event closes the stray card by turn_id, so it self-heals), but the invariant as documented is false. Fix either way:
- (preferred) adapter-side fence: remember closed team keys / unbound routes and reject submitted-with-origin for them; or
- downgrade the claim in the three docs + PR body to "depends on core event ordering; a late submitted-with-origin may open a transient card that its settled event closes".

## Should fix (this PR or immediate fast-follow)

- **[minor]** `packages/channel/feishu-transport/README.md:8,28` still documents `removeReaction` as a transport capability — stale published docs after the interface removal.
- **[minor]** `packages/channel/feishu-channel/src/feishu-channel.ts:412-417`: the team_leader `sendReply` path eagerly calls `targetRouter.resolveTarget(...)` *before* `sessionSendReply`, and resolveTarget throws on target conflicts (`feishu-target-router.ts:296`) — a display-only COT-anchor lookup can then break a user-visible Reply. The eager call is redundant: the observer at :427 already has a lazy `target ?? resolveTarget(...)` fallback. Drop the eager pre-resolution (or move it inside the fail-open observer). The new `feishu-cot-reply-routing.test.ts` covers dispatcher/neutral conflicts but not the team_leader case — add one.
- **[minor]** Runtime `error`-field divergence: claude-code now sets `error: null` on failed tool calls with the text in `result` (`runtime-submissions.ts:240`), while codex still fills `error` from `item.error` (`turn-manager.ts:453-467`). The projection normalizes via `error ?? result` and no other core consumer reads `activity.error`, so there is no live bug — but the `RuntimeActivity` contract is now inconsistent across runtimes and the claude-code change file comment doesn't mention the semantic change. Align the runtimes (or document the divergence in dreamux-types) and amend that change file comment.
- **[minor]** PR body says "every admitted turn publishes exactly one terminal display fact"; a dispatcher turn whose `channelOrigin` JSON snapshot fails (circular/BigInt metadata) publishes zero events, settled included (`conversation-projection.ts:150`). Deliberate fail-open, rare trigger — scope the claim to turns that freeze an origin (and check the docs don't repeat the stronger wording).

## Nits (no action required)

- `collaboration-routing.ts` sits at exactly 700 lines, the lint cap — any addition now breaks lint; the tracked split follow-up should land soon.
- `turn.settled` timestamp uses projection time, not settlement time (cosmetic).
- The dormant `it.skip` ANSI case is documented with an active replacement assertion — acceptable.
- Leader `pendingTurns` entries can linger past settle when no user message arrives; bounded (512, drop-newest) and harmless.

## Verified clean

- #63 supersession is legitimate: the live gate still runs real codex and still asserts the folded non-blocking semantics (2 turn/starts → 1 turn/completed); the reaction assertions are inverted to positive absence (`bot.reactions`/`bot.reactionOps` empty), and `setReactionError(new Error('automatic reaction attempted'))` makes any automatic-reaction regression fail loudly. The domain doc, decision record, and `entity-owned-teammate-lifecycle` doc are mutually consistent; no tri-state mandate left standing.
- Projection injection is constructor-level (dispatcher + TeamLeaders only; teammates excluded by construction at the `eventScope` chokepoint); all four projection entry points are fail-open with tests; caps/drop-newest and sanitization are tested; type surface stays Feishu-free.
- Change file types are correct (feishu-channel `major` on 5.0.0; five 0.x packages `minor` with `BREAKING:` notes); the CI major-gate passes on this set.
- No sync IO in package source; no secrets/internal identifiers in the diff; gitleaks green.
- CI: all gates green. The Ubuntu rush job's first run was a runner-overload flake (12 blanket 5s timeouts in untouched `workflow-service.test.ts` + MaxListenersExceededWarning; macOS passed; six prior Ubuntu runs on next green) — re-run passed.

Happy to re-review the delta and approve once the two majors are addressed.

ryanxiang7 (approved):
Re-review of fix delta (665408aa → f0c44b48): both majors resolved.

- M1 (change file comment): now enumerates all removals (RECEIVED_REACTION_EMOJI / IN_PROGRESS_REACTION_EMOJI exports, FeishuBot.removeReaction, FeishuChannelSession.state, FeishuChannelState) with consumer actions per caller/implementer category.
- M2 (late-receipt revival): the LeaderLifecycleFence closes every revival path I flagged — global fence on Team close, endpoint-level route fences on unbind/route-replace matched against the anchor's authoritative binding (no target fallback, no topic false positives), fallback notification path fenced, single admittedTurnId validated at every event entry point, fence clearing on restart/rebind, dual 512 caps with LRU eviction, 12 dedicated regression tests asserting the real invariants.

The four should-fix items are also in: transport README cleaned, sendReply target resolution moved into the receipt observer (conflict no longer breaks a sent Reply, with a team_leader conflict test), claude-code error field aligned to codex semantics (disclosed in the change file), and the 'exactly one terminal fact' claim narrowed to turns that enter conversation projection (decision record + PR body). The operator-aligned scope expansion (team_member neutral publishing restored; Feishu COT renders dispatcher/leader only) is correctly wired and disclosed across the three change files.

One non-blocking follow-up: the Team-close fence is leader-keyed, so a same-team leader replacement whose old state was already reaped would not be fenced (currently unreachable — leader identity is de facto immutable per team, and the dissolve path is backstopped by route fences). Filing a follow-up issue to make the fence team-granularity as defense-in-depth.

CI 8/8 green on f0c44b48. Approving.

PR #350, opened 2026-08-28
Refactor/minimal agent runtime provider

YourWildDad:


ryanxiang7 (changes requested):
Blocking findings for exact head `fd00db650f439053871625e176643f0e8b3b8773`:

1. The Feishu COT adapter/session was materially rewritten, while all five direct COT suites (78 tests in the base) were deleted without equivalent replacement coverage. The surviving session-binding test stubs `FeishuCotSessionSeam` and does not exercise anchor lifecycle, outbox batching/merge, settle mapping, routing fences, fail-open behavior, or error redaction. This leaves the Stage 9 replacement matrix incomplete for a frozen, load-bearing contract.

2. The canonical architecture/KB still contradicts the new public contracts. In particular, `.agents/reference/current-architecture.md` still documents the removed `completionInput`/`channelInput` split, optional `disableFeatures`, and turn-level `outputSchema`; `.agents/decisions/cli-and-package-naming.md` still lists five removed `*-mcp` subcommands. This violates the same-PR Knowledge Delta requirement for a boundary and CLI rewrite.

3. The new Team/TeamMate/Scheduler Command and MCP adapters duplicate parsing, schemas, domain invocation, and result projection across parallel `commands.ts` and `mcp-delegate.ts` implementations. Provenance, authorization, and model-facing descriptions can remain adapter-specific, but the shared domain operations/codecs need one owner to avoid contract drift.

4. Several replacement tests assert private source text rather than observable contracts, including fixed-width method-body slicing in `team-dissolve-contract.test.ts`, import-string scanning in `channel-service.test.ts`, and a repository-wide ban on the ordinary word `generation` in `workflow-service.test.ts`. These can fail on behavior-preserving refactors and pass semantic regressions that retain the searched text.

CI is green, but these findings concern deleted/rewritten tests and therefore cannot be cleared by that green run alone. Please restore behavior-level coverage, align the current KB with the exact head, consolidate the duplicated domain-owned surface, and replace source-shape assertions with behavioral or proper static-boundary checks.

ryanxiang7 (changes requested):
Re-review of exact head `aa769c607d650ac44f6320522782ac15dc22f792`: **REQUEST CHANGES**.

The two requested focus areas are clean:

- Classified Command/MCP failures preserve `code` / native `message` / `action`; unclassified failures become `INTERNAL` with the native message. The UUID, sanitization, and per-operation allowlist paths are gone.
- Feishu bot `open_id` lookup does not negatively cache failure/empty results, retries on the next inbound message before the @ gate, caches only success, and coalesces concurrent inbound lookups through one in-flight promise. The new tests exercise these behaviors.

However, three previously accepted findings remain open on the full PR:

1. **The rewritten Feishu COT adapter/session still lacks equivalent direct behavioral replacement coverage.** The original PR deleted the direct adapter/events/lifecycle/outbox/state/reply-routing suites (85 tests across six suites), while the corresponding production implementation remains. `fd00db65..aa769c60` adds no replacement under `packages/channel/feishu-channel/tests/`; the new `feishu-transport/tests/self-identity.test.ts` covers only the new bot-identity recovery behavior.

2. **The canonical KB still contradicts the new head contracts.** The fix range does not touch `.agents/reference/current-architecture.md`, so the previously reported stale flattened-submission, `disabledFeatures`, session-bound `outputSchema`, generic `dreamux mcp`, and admin/stop ownership descriptions remain. This boundary/CLI rewrite must satisfy the same-PR Knowledge Delta rule.

3. **The accepted source-text mirror-test finding is unchanged.** The relevant test blobs are identical from `fd00db65` to `aa769c60`: `team-dissolve-contract.test.ts` still scans production files and uses a fixed 400-character method window; `channel-service.test.ts` still asserts import text; `core-event-catalog.test.ts` still matches private callback source; and `workflow-service.test.ts` still scans imports and bans the ordinary word `generation`. These assertions lock private implementation shape instead of observable behavior or a proper static boundary.

All GitHub checks are green on this exact head, but these findings concern missing/replaced tests and stale contract documentation, so that green run does not independently clear them.

ryanxiang7 (approved):
Approved at exact head `aa769c607d650ac44f6320522782ac15dc22f792`.

The final production-code re-review found the Command/MCP error contract and Feishu bot identity recovery correct, correctly owned, and behavior-tested. Lifecycle/correctness review found no regression, and all GitHub checks are green on this head.

Per the operator's explicit decision, the remaining KB and unit-test quality findings are non-blocking for this merge and are tracked in #352: direct COT behavioral replacement coverage, canonical provider-boundary KB reconciliation, and replacement of source-text mirror tests.

PR #353, opened 2026-08-31
Clean up provider-boundary residue from #350

YourWildDad:
Follow-up to #350. A review of the merged refactor found a set of leftovers: one real boundary leak, dead machinery the refactor stopped using, messages that named things the code cannot see, duplicated validation, and two defaults that hid a mistake instead of reporting it. Each item is a separate commit.

## Provider boundary

**Narrow channel factory context to ref-only** (`bf8e05f0`)

The minimize-provider-boundaries design states that provider registration identity is Core-owned loader metadata, kept in the `RegisteredProvider` wrapper and never handed to the implementation side. The published contract says the same in the type: `ProviderFactoryContext` carries only the canonical ref.

The channel loader did not honor it — it projected `({ ref, descriptor }) => ({ ref, descriptor })`, so Core's internal seed descriptor travelled to every external channel factory, while the sibling agent-runtime loader passed `{ ref }` only. The `descriptor` field's JSDoc still read *"Descriptor the provider must expose back to Dreamux"*, echo-era wording contradicting the module's own header.

Both loaders now type their factory context as the published `ProviderFactoryContext`, deleting the two structurally identical local aliases and making the asymmetry unrepresentable. Zero behavior change: the built-in Feishu factory takes no context argument at all. The loader test pins the received context's keys as exactly `['ref']`.

**Delete the echo-era provider descriptor assertion** (`55c1605d`) — `assertProviderDescriptorShape` validated that a provider echoed Core's descriptor back. Nothing echoes anything back now, and it had zero callers.

## Dead machinery

**Delete the dead inbound-turn shapes** (`d0d95a39`) — `InboundTurnInput` / `InboundAttachment` described a turn that a runtime-neutral layer pre-rendered; that layer is gone. No consumer in any package, and deliberately withheld from the root export, so the published surface is unchanged. The root-export guard now asserts the carve-out is gone rather than locking it in.

**Drop the producer-less `dispatcher.status` field** (`0dbbb754`) — `last_lost_thread_id` was declared as the literal `null`, schema-typed `NULL`, required, and hardcoded to `null` in every response. Nothing produced or read it, in any casing.

**Remove the dead skill-source fences from the MCP delegates** (`44dab9c3`) — both delegates called `normalizeSkillSources` with a literal `null`, which returns immediately before validating, so the fence never ran and the guarded spread was always empty. The live fence on the Command path is untouched.

## Messages that borrowed what the code cannot see

**Say what the Codex runtime knows when it rejects an approval** (`8661bf86`) — the warn log and the handler's module doc explained the rejection with *"Feishu outbound is MCP reply-only"*. A runtime package sits behind `AgentRuntimeProvider` and cannot see which Channel, if any, is attached. Both now state the runtime-owned truth.

**Write the Codex runtime source in English** (`2adbd616`) — two runtime errors that reach a caller were written in Chinese, so a non-Chinese-reading operator hit a wall exactly where the message should have told them what to change. Same meaning, same remedy, in English; issue-section citations in module docs likewise. CJK test fixtures that prove escaping works are untouched.

## Correctness

**Key provisioning idempotency to the inbound message** (`8fa96b96`)

Automatic provisioning minted a fresh UUID per `team.create` attempt, so the request id carried no identity Core could recognize twice. If Feishu redelivered one message, the second delivery presented a new id and Core built a second Team for one conversation.

The request id is now the inbound message id, used bare. A redelivered message replays the same id and Core answers with the Team the first attempt made; when that attempt died between `team.create` and the routing bind, the replay answers `existing` and this run installs the binding, recovering instead of duplicating.

Identity is scoped to the message rather than the topic deliberately: Core keeps a request's acceptance record permanently, so a thread-scoped id would replay `closed` forever once a Team was dissolved and that topic could never be provisioned again. Two costs are accepted rather than defended — a *different* message after a partial failure creates a second Team, and a replayed id whose policy snapshot changed raises an idempotency conflict, reporting the message unsubmitted so it falls back to the Dispatcher.

**Require an explicit state root for a Feishu session** (`a6bdbc09`) — the provider defaulted `state_root` to `'.'`, so a mis-wired host wrote durable routing state into whatever directory the process started in, surfacing only later as state that had vanished. A missing or empty root now fails session creation. No impact on the normal Dreamux path, which always injects one. Adds the provider's first construction tests.

## Vocabulary and duplication

**Speak session, not thread, on the Dispatcher status surfaces** (`4acccaf5`) — the status projections named their field `thread_id` / `threadId`. Thread is Codex's word; Core must not describe a runtime-neutral fact in one runtime's vocabulary, and Dreamux had already settled on `session` everywhere else. The values were never Codex-specific — every site already read `identity.session.id`. Renamed across `DispatcherSummary`, `DispatcherRuntimeStatus`, `LiveDispatcherRuntimeStatus`, their producers, and the `dispatcher.status` output field plus its schema. No aliases. No persisted file, key, or filename is touched: the on-disk identity schema already said `session`.

**Report the real package version to MCP clients** (`449ac7f8`) — all four MCP servers announced version `0.4.0`, hand-written in four places; the package is at 0.22.0. `serverInfo.version` now comes from the package manifest via one shared constant, so it tracks releases automatically. A static JSON import keeps it a load-time constant — no filesystem read, no sync-IO concern — and the relative path resolves identically from `src/` and compiled `dist/`, verified by running the compiled module.

**Give the scheduler one cron validator** (`79108061`) — the five schedule rules were implemented twice and had already drifted: the reload path folded in the minimum interval while the command path kept it separate, and the two worded an unknown timezone differently. The rules now live in one validator with the failure classification as the parameter, so a caller mistake is still a `RuleViolation` and corrupt state is still a `LegacyStateError`, messages unchanged. A redundant third validation pass at scheduler start is removed — verified redundant first: `assertCurrent` applies the same rules and the store's parser already rejects an empty action prompt.

## CI gate repairs

**Assert the ref-only factory context without a cast** (`b87101be`) — the loader assertion added above held the received context in a nullable local, which TypeScript keeps narrowed to `null` (the factory that assigns it runs inside the loader, not in that flow), so each read needed a cast and the widening one to `object` failed `rush typecheck:tests`. `rush test` never saw it: vitest transpiles through esbuild without typechecking. The contexts are collected into a `ProviderFactoryContext[]` instead, so both assertions read the element directly with no cast.

**Point the channel-input decision at what still exists** (`38f92e99`) — the `channel-input-runtime-assembly` decision record linked `turn.ts`, deleted above, breaking `.agents/scripts/check.sh`. The record and its prose stay — the routing/display split it settled is still the live invariant — with the dead path demoted to inline code and a dated note recording that the pre-render shapes are gone and that Core now renders the one provenance envelope, linking the surviving owners.

## Gates

- `rush update`, `rush build`, `rush smoke-built-cli`, `rush typecheck`, `rush typecheck:tests`, `rush lint` — clean across all 8 packages.
- Full `rush test` green, with the live Codex integration exercised against real codex 0.151.0 (not skipped).
- `.agents/scripts/check.sh` — KB OK, 146 files reachable.
- `rush change --verify` against `origin/next` — clean.
- Change files for `@excitedjs/dreamux`, `@excitedjs/dreamux-types`, `@excitedjs/agent-runtime-codex`, and `@excitedjs/feishu-channel`.

## Breaking changes

Two items break a caller, and their change files say so with a leading `BREAKING:`:

- **The session-vocabulary rename** — `dispatcher.status` renamed its `thread_id` output field, and the Dispatcher rows in `dispatcher.list` and `server.status` renamed theirs. No alias is kept, so a script still reading `thread_id` reads undefined and must read `session_id` instead. Recorded as `minor` on `@excitedjs/dreamux`: a pending `major` would bump a 0.x package straight to 1.0.0, so the repo records 0.x breaks as `minor` with a leading `BREAKING:` note.
- **The Feishu state-root guard** — an omitted `state_root` used to write durable routing state into the process working directory and now throws. The published context type still marks the field optional, so nothing fails to compile; the break is at runtime. `@excitedjs/feishu-channel` is past 1.0.0, so this is a real semver `major`. Dreamux hosts always inject a state root and are unaffected.

No config, state, cache, run, or log path semantics changed; no persisted format changed; no rebuild is required for either.


ryanxiang7 (approved):
Approving after a full cross-team review.

Reviewed the 15-commit batch as one change: the provider-boundary residue cleanup, the two CI gate repairs, and the release-contract corrections. Method: 7 independent finder lenses with 3-vote confidence scoring, plus a commit-by-commit adjudication of every PR-body claim against the code.

All findings from the first round were addressed in 3aa0f72:
- The `dispatcher.status` `thread_id` to `session_id` rename is a no-alias breaking change; its change file is now `minor` with a leading `BREAKING:` note, per the 0.x rule.
- The Feishu `state_root` fail-loud guard is a runtime break for embedders (the published type still marks the field optional); its change file is now `major` with a leading `BREAKING:` note, per real semver for a post-1.0 package.
- The channel-input decision note now matches the current architecture: the structured turn contract is gone, and routing ids reach the model only as rendered attribute text inside the provenance envelope.

CI is green on both platforms (rush chain, rush change declaration check, KB link check, gitleaks, shellcheck). No remaining issues from this review.

Issue #352, opened 2026-08-31
Restore COT contract coverage and reconcile provider-boundary KB


ryanxiang7:
## Context

PR #350 (`aa769c607d650ac44f6320522782ac15dc22f792`) completed the minimal Agent Runtime / Channel provider boundary refactor. Its production-code re-review found no remaining correctness blocker, but identified test and knowledge-base debt that the operator explicitly accepted as non-blocking for that merge.

### Restore direct Feishu COT behavioral coverage

Add equivalent behavior-level coverage for the still-live COT adapter/session/outbox/events/state/reply-routing implementation after the original direct suites were removed. Cover at least:

- reply-anchor lifecycle and routing fences;
- outbox merge and append batching;
- turn-settlement projection;
- conversation activity bounds;
- failure isolation and error redaction;
- session/state recovery and close behavior.

Do not mechanically restore obsolete implementation-coupled assertions; test observable contracts through the current architecture.

### Reconcile the canonical knowledge base

Update current architecture and decision references to match the landed contracts, including:

- the flattened text submission surface;
- required `disabledFeatures`;
- session-bound `outputSchema`;
- the unified `dreamux mcp` CLI surface;
- current admin command and Team stop/cleanup ownership.

Run `.agents/scripts/check.sh` and remove or mark superseded descriptions rather than leaving conflicting current-state claims.

### Replace source-text mirror tests

Replace tests that read production source and assert private text shape, including:

- fixed-width method-body slicing and private call-name matching;
- import-string presence/absence checks;
- full private callback source matching;
- broad bans on ordinary words such as `generation`.

Use observable behavior tests, dependency-injection/call-count tests, or proper AST/lint/static-boundary checks as appropriate.

## Acceptance criteria

- The current COT behavior listed above has direct executable contract coverage.
- Canonical KB statements agree with current source and `.agents/scripts/check.sh` passes.
- The identified source-text mirror assertions are removed or replaced with behavior/static-boundary checks.
- Relevant Rush build/typecheck/lint/test gates pass.

## Provenance

Follow-up from the final review of PR #350. The operator classified this scope as non-blocking for that PR; this issue preserves the accepted cleanup trail.

PR #354, opened 2026-08-31
test: restore Feishu COT contract coverage

YourWildDad:
## Summary

- restore direct behavior-level coverage for the current Feishu COT adapter and session contracts
- cover visible-message anchors, outbox budgets, settlement, routing fences, failure isolation, and close behavior through real public seams
- replace brittle source-text mirrors with runtime assertions, type-level contracts, and TypeScript AST boundary queries

This PR addresses the test-coverage portion of #352. The canonical KB follow-up remains outside this PR.

## Validation

- `.agents/scripts/check.sh`
- `node common/scripts/install-run-rush.js change --verify`
- `git diff --check`

The focused additions cover 116 Feishu COT cases and 80 rewritten boundary cases. Mutation checks demonstrated that the restored suites or type gates reject all 13 targeted regressions.

### DIFF
--- a/.agents/tasks/architecture/minimize-provider-boundaries/README.md
+++ b/.agents/tasks/architecture/minimize-provider-boundaries/README.md
@@ -623,6 +623,12 @@ architecture merely because it is possible.
   gating, with one in-flight lookup shared by concurrent messages. Scope is the
   Feishu transport implementation, focused tests, and its Rush change note;
   the separate empty-TeamLeader recovery defect is explicitly excluded.
+- Post-merge COT test-coverage correction: Approved by the operator on
+  2026-08-31 for a new PR branched from the merged `next` head. Scope is the
+  test-related portion of issue #352: restore direct behavior-level coverage
+  for the current Feishu COT adapter/session contracts and replace the accepted
+  source-text mirror assertions with behavior-level or proper static-boundary
+  coverage. The canonical-KB portion of #352 is not part of this correction.
 
 ## Delivery
 
--- a/common/changes/@excitedjs/dreamux/test-feishu-cot-coverage_2026-08-31-11-42.json
+++ b/common/changes/@excitedjs/dreamux/test-feishu-cot-coverage_2026-08-31-11-42.json
@@ -1,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "Restore and strengthen test coverage without changing published package behavior.",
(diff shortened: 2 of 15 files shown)


ryanxiang7 (changes requested):
Re-review of exact head `0e11d036a32a586b46ba73d4ef54979f82a2a842`: **REQUEST CHANGES**.

The restored Feishu COT coverage is clean: the tests drive the real adapter/session/client boundaries, observe public behavior and wire requests, do not backfill adapter private fields, and the fixtures do not copy the production state machine. Three of the four source-mirror replacements are also valid behavior/type/module-boundary tests. The two `type: none` Rush declarations are accurate, there is no product-source change, and all GitHub checks are green.

One prior finding is not fully closed:

- `packages/dreamux/tests/core-event-catalog.test.ts:574-638` parses the production implementation and locates private members/calls by the names `publishAgentState` and `onPersisted`, depends on call argument positions, then uses `initializer.getText(source)` to reconstruct source text and compare the private callback expression (`'teammate.state'` and `identity.team_id`). This is still an implementation mirror, now expressed through AST. A behavior-preserving rename, inline, helper extraction, or equivalent expression rewrite would fail the test, while preserving those nodes would not prove that the actual event is published correctly.

AST is appropriate for a true module/type/dependency boundary, as used elsewhere in this PR. This particular assertion is about runtime event publication and should instantiate the owner path and observe the emitted `teammate.state` payload (or be reduced to a genuine public type/module boundary) instead of inspecting a private callback body.

Please replace this remaining AST implementation-shape assertion with an observable behavior test, then request exact-head re-review.

YourWildDad:
Closing at the operator's request: this coverage pins the per-turn admission contract that has since been withdrawn.

A TeamLeader conversation now presents activity from two facts only — where a card would open, and whether one is open — so there is no per-turn gate left for these cases to assert. Concretely, this file calls `adapter.onTurnSubmitted(...)`, which no longer exists, and it pins three behaviors that were deliberately removed: admitting a completion or cron fire only once its own text arrives, refusing to narrate an out-of-band turn, and ignoring a settlement for a turn the card is not showing. The last of those is the reported bug rather than a contract worth keeping.

Equivalent coverage for the current contract, including the two accepted grouping edges, lives in `packages/channel/feishu-channel/tests/feishu-cot-mid-turn-message.test.ts`.

PR #356, opened 2026-08-31
fix(state): restore the legacy persisted session identity layout

YourWildDad:
## Why

PR #350 (`2ed5f5ea`, unreleased) reshaped persisted state so far that an
existing 0.22.0 user could not upgrade without losing facts. The operator ruled
out both read-old-write-new and a startup migration, so this PR takes the other
option: **restore the old persisted layout** where the new one bought nothing.

Three divergences were found. This PR fixes two of them; the third is
deliberately deferred.

### 1. The session identity is a plain string again

#350 replaced the persisted `session_id: string | null` with a nested
`session: { id }` object carrying a provider-boundary shape. Nothing in Core
needed that nesting: Core stores the id, checks it for presence, and hands it
back to the same provider. It never parses, indexes, or branches on it.

So the identity contract collapses back to one nullable string, end to end, and
persists under the legacy field name `session_id`. The in-memory contract stays
camelCase (`sessionId`) and the persisted JSON stays snake_case (`session_id`),
matching the rest of the repo's split.

### 2. `role` and `transcript_locator` are inert residue, not rejections

#350 added both to the removed-field reject list. Neither field is consulted by
this version — `role` is derived from the owning directory, and
`transcript_locator` was replaced by the Activity seam's opaque id. Rejecting
them costs the operator a full rebuild to delete two fields nothing reads.

They are now inert residue: no path creates, opens, validates, or deletes them.

### 3. The rejection rule itself is now written down

Reviewing the list surfaced the real rule, which was implicit before. A field
earns rejection only when **both** hold:

1. a released build actually wrote it, **and**
2. accepting the record would silently discard a fact this reader cannot see.

`checkpoint` and `session_ref` qualify — `session_ref`'s resumable id sits one
level below where `session_id` is read, so accepting such a record silently
drops a live session. `role` / `transcript_locator` fail (2).

An earlier revision of this branch also rejected the nested `session` object.
That was the same mistake inverted: `git log` confirms nested `session: { id }`
exists only inside unreleased #350, so it fails (1) — a shape no released build
ever wrote cannot reach a real upgrade, and gating on it would be a permanent
tax for an impossible input. It is also redundant, because `session_id`'s own
`string | null` type check is the better gate: "no usable id found" already
degrades correctly to "start a fresh session", while a *present but corrupt*
value (a number, an object, an empty string) still fails validation and is
reported as an unreadable record rather than being coerced into the same "no
session" answer.

`assertNoRemovedRecordFields` remains the single chokepoint with exactly one
caller (`identity-store.ts`); its list is now 5 entries, not 7.

### 4. Deferred: the Feishu routing pair

`channel-bindings.json` and `collaboration-spaces.json` remain a rebuild. The
operator's decision is to design that half as a read-only projection when this
code is synced downstream, rather than to churn it twice.

## Verification

Both real state trees were audited with the compiled `dist/` readers, not by
inference:

| | identities read ok | team records | startup detector |
|---|---|---|---|
| legacy (pre-#350) layout | 1692 / 1692, `session_id` round-tripped | 175 ok | exactly 2 findings — the Feishu pair |
| current layout | 1700 / 1700 | 176 ok | clean |

Cron: 39 jobs, all `prompt-agent`, no removed `deliver` field.

Of the 1700 identities, only 19 carry a resumable id and only 5 of those are
running, which bounds the real-world loss (see `Review:` below).

Suites: the 3 legacy-state test files pass (68 tests), `tsc --noEmit -p
tsconfig.tests.json` is clean, `rush build` exits 0, the core suite is 963
tests / 65 files green including the codex 0.151.0 live integration, full
`rush test` exits 0, and `.agents/scripts/check.sh` reports KB OK (146 files).

While writing the corrupt-`session_id` test I asserted the wrong contract: I
expected `read()` to reject. It does not — it deliberately downgrades a
non-`LegacyStateError` parse error to warn + `null` so one unreadable entity
never sinks a whole collection scan. Per the standing rule, the **test** was
fixed rather than the source: it now pins the distinction that actually matters
(a corrupt value must not silently collapse into the same `null` a legacy record
produces) by asserting both the `null` and the warning. The `readSessionId` doc
comment, which had overclaimed "fails loud", was corrected with it.

## Changelog

The #350 change file is rewritten rather than supplemented, since #350 is
unreleased and the two notes describe one upgrade.

Its `Rebuild:` now names **only** the Feishu pair plus the
`collaborationSpace` config key and the re-bind. Per repo rule, using `Review:`
instead of `Rebuild:` for the same-shape half required explicit operator
approval, which was given. The `Review:` half states that identities, Team
records, and cron jobs are not rebuilt and need no action; that `session_id` is
kept; that `role` / `transcript_locator` are inert residue; that an agent whose
session was recorded in the short-lived nested shape simply starts fresh (so a
running agent may need its context restated); and that a dissolve still waiting
for its Team to go idle is not resumed and must be re-submitted.

Type stays `minor` with a leading `BREAKING:` — `dreamux` is on the 0.x line,
where a `major` file would jump it to 1.0.0 (CI-enforced).

## Docs

`.agents/reference/state-and-paths.md` was doubly wrong (it listed `session_id`
itself among the removed fail-loud fields) and is corrected.
`service-topology.md`, `service/CLAUDE.md`, and the `dreamux-maintenance`
skill's `service-lifecycle.md` follow the same contract.
`self-upgrade.md` needed no change — its `session_id` reference is the CLI
`dispatcher.status` output field, which this revert makes consistent again.


ryanxiang7 (approved):
Reviewed the session-identity restore and the change-file declaration.

Code: the legacy persisted session identity layout (plain `sessionId` string) is restored consistently across `dreamux-types`, both agent runtimes, and core; `AgentRuntimeSessionRef` and the `TSession` generic are removed cleanly. #350 was unreleased, so the runtime/types packages have no net published change.

Change files: the rush change declaration check only counts Added files (`git diff --diff-filter=A`), so the modified #350 dreamux entry did not register. The fix relocates the corrected dreamux note to a new file (Added + Deleted under `--no-renames`) and adds `none` entries for the three runtime/types packages. This avoids keeping the stale #350 "delete the whole state dir" rebuild instruction, which is false and destructive after this PR. The final dreamux note correctly scopes `Rebuild:` to the channel-binding store and uses `Review:` (no rebuild) for the identity restore, per the operator's decision.

All CI checks green on 34c2d62.

PR #362, opened 2026-09-01
feat(channel): ask the user a question as a Feishu card

YourWildDad:
Adds the `ask_user_question` MCP tool, offered to both the Dispatcher and a
TeamLeader. The model sends a decision card into a Feishu chat; the user picks
an option or writes their own answer; the answer comes back as an ordinary
inbound message.

## Arguments mirror AskUserQuestion

Field names and descriptions are lifted from Claude Code's own AskUserQuestion
so a model reads the two as one tool: 1-4 `questions`, each with a `header`
chip, a `question`, and 2-4 `options` of `label` + `description`.

Four differences:

- **`chat_id` is required.** A chat tool needs a destination; AskUserQuestion
  has no such concept.
- **`message_id` is offered.** When the question came out of something someone
  said, the card is addressed under that message — into its topic, under the
  anchor the target router already holds — exactly as `reply` is addressed.
  Without one the card addresses the chat, which in a topic group opens a topic
  of its own: right for a question belonging to no particular message, wrong
  for one that does. A message id from another chat is ignored rather than
  obeyed, `reply`'s rule, so a stale id cannot redirect a question into a
  conversation it was not meant for.
- **No `multiSelect`.** The operator ruled multi-select out of this channel, so
  every question takes exactly one answer. The two sentences in the upstream
  descriptions that referenced the field are dropped rather than left dangling.
- **No `preview`.** A preview belonging to the focused option has to sit beside
  the options, so it cost every question a two-column `column_set` and reshaped
  the layout. The operator removed it on sight:
  "这个 preview 有点复杂了，给他去掉，他也会影响卡片的布局".
  The field, its schema property, its parse branch, and the tool-description
  paragraph that taught the model to use it went with the column.

## The tool does not return the answer

It returns `{ request_id, status: 'asked', next }` as soon as the card is sent,
and `next` instructs the model to end its turn. A click settles the round
server-side and the answer reaches Core as an ordinary inbound submission — the
path a typed reply takes, so nothing downstream learns a card produced it.

Blocking the tool call was the alternative, and it loses the case that matters:
a person who answers after lunch. The MCP client, not Dreamux, bounds how long
a tool call may run, so a blocking handler would be abandoned mid-question by a
timeout this repo does not own.

## Why the card looks the way it does

Feishu has no radio component. `select_static` is its only native single-select
and its options carry a label and nothing else, so an option's `description`
would vanish the moment the dropdown closed. Each option is therefore its own
`interactive_container` — label, then description — which is also what the
open-source Feishu ask-user renderers converged on.

That choice is what makes the server authoritative: an `interactive_container`
holds no client state, so "which option is selected" exists only in the
session's registry, and the selection is visible solely because every click
repaints the whole card.

There is deliberately no `form`. A form batches its inputs until submit, which
would cost each question the ability to settle on its own; without one, an
`input` carries its own callback and Feishu draws it a send button, so
committing free text is the same explicit click as choosing an option.

## Expiry

Feishu stops accepting interaction on a card 15 minutes after it is sent. Past
that cutoff the card still looks live but every click is dropped, and the model
would wait for an answer that can no longer be given. A round therefore closes
itself before the cutoff, repaints the card as closed, and tells the model that
no answer came and to stand still until the user's next message.

`editCard(messageId, card)` is added to `@excitedjs/feishu-transport` for that
repaint. It generalizes the `im.message.patch` call `editText` already made.

## Operator rulings recorded

Quoted in `.agents/domains/channel.md` and at the code sites they govern, so a
later reader meets a decision rather than an apparent gap:

- multi-select dropped: "废弃多选（大概率完全用不到），只把单选做好即可"
- anyone may answer: "不需要限制，所有人都可以点" — the clicker's open_id rides
  along as `sender_id`, but no authorization check stands between a group
  member and the card.

## Boundary fix on the way through

`feishu-channel.ts` sat two lines under the `max-lines` gate, so any new
capability would have broken it. Rather than widen the gate, `outboundTarget`
moves to `FeishuTargetRouter` — it reads state the router already owns, and had
its fallback reimplemented outside its owner. The file ends up smaller than it
started (698 → 696).

## Verification

`rush build`, `rush lint`, `rush typecheck`, `rush typecheck:tests`,
`rush test`, and `rush smoke-built-cli` all green, including the live Codex
tests. `.agents/scripts/check.sh` passes. 25 tests cover the round state
machine (pick, free text replacing a pick, clearing an answer, cancel, double
submit, unknown round, a round whose card never sent, an empty submit, expiry,
and expiry losing a race to a click), how the card is addressed, the tool's
argument rejections, and the anchor regression below.

**Verified live** against a real bot on the `0.23.0-alpha.g7bf178cf6e11`
build: the full MCP tool → card → click → inbound injection chain, and both
refusals — a submit with nothing chosen stopped by its toast, and dismissing the
card ending the round.

**Still not exercised live:** the expired-card repaint, which needs the full
14-minute wait, and addressing a card under a `message_id` in a topic-mode
group.

## Note for review

One bug worth a second pair of eyes, since it was caught late: the settlement
originally used its synthetic dedup id (`ask_user_question:<hex>`) as the
submission `anchor.messageId`. That anchor becomes a COT presentation's
`originMessageId` and is handed to Feishu's API, so every answered round would
have produced a broken COT card. It now carries the question card's real
message id — from the click event, or from what the send recorded when a round
expires — with a regression test pinning the distinction.

## Review round

Seven findings from @YisenFE's review, in `022f5195`:

- **A failed send no longer leaves a round behind.** `open` builds the round
  and its card but registers neither; `activate`, called once the send lands,
  puts the round in play and starts its clock. Previously a send that threw
  left an armed round whose TTL would, 14 minutes later, tell the model no
  answer came for a question the user was never asked. Making the round's
  existence a construction property removes the rollback rather than adding
  one.
- **`message_id`**, above — in a topic-mode group the card used to open a topic
  of its own no matter what, so the answer came back somewhere the asking agent
  was not.
- **The repaint no longer waits on delivery to Core.** Feishu gives a card
  callback a few seconds before the click looks dead, and waking an agent can
  outlast that. Delivery already swallowed and logged its own failures at error
  level, so nothing was lost by detaching it.
- **提交 with nothing chosen is refused with a toast.** The button sits directly
  under the questions, so it is also the one pressed by accident, and a round
  settles exactly once: spending that settlement to report every question as
  unanswered is worse than saying nothing. A partial answer still submits.
- The free-text branch is unconditional, so the unreachable
  `return { kind: 'ignored' }` after it is gone.
- The expiry callback drops a `.catch` that could not fire — both halves of the
  expiry path already handle their own failures.
- The submitted card gains the `summary` its two siblings carry, so a
  notification names it.

https://claude.ai/code/session_01K8SjnH3Kt7uSUjgLFZ7m8D

### DIFF
--- a/packages/channel/feishu-channel/src/feishu-session-ops.ts
+++ b/packages/channel/feishu-channel/src/feishu-session-ops.ts
@@ -40,7 +40,13 @@ import { AsyncMutex } from './lib/mutex.js';
 import type { FeishuChannelSessionOptions } from './feishu-channel.js';
 import type { PeerBot } from './chat-bots-store.js';
 import type { FeishuTargetRouter } from './feishu-target-router.js';
-import type { FeishuInboundDelivery } from './feishu-submit.js';
+import { CHANNEL_REMINDER, type FeishuInboundDelivery } from './feishu-submit.js';
+import type {
+  AskUserExpiry,
+  AskUserRegistry,
+  AskUserSettlement,
+} from './feishu-ask-user.js';
+import type { AskUserQuestionSpec } from './feishu-ask-user-card.js';
 import { FeishuOperationError } from './feishu-bounded-operation.js';
 import {
   alwaysActiveSessionFence,
@@ -71,6 +77,8 @@ export interface SessionHandle {
    * learns what a binding or a space policy is.
    */
   delivery: FeishuInboundDelivery;
+  /** Open ask-user rounds; a question outlives the tool call that asked it. */
+  askUser: AskUserRegistry;
 }
 
 /** Build a package-private handle from a session's internal fields. */
@@ -81,6 +89,7 @@ export function sessionHandle(input: {
   botDisplayName: string;
   targetRouter: FeishuTargetRouter;
   delivery: FeishuInboundDelivery;
+  askUser: AskUserRegistry;
   sessionFence?: FeishuSessionFence;
 }): SessionHandle {
   return {
@@ -90,6 +99,7 @@ export function sessionHandle(input: {
     botDisplayName: input.botDisplayName,
     targetRouter: input.targetRouter,
     delivery: input.delivery,
+    askUser: input.askUser,
     sessionFence: input.sessionFence ?? alwaysActiveSessionFence(),
   };
 }
@@ -359,10 +369,151 @@ export async function approvePairingByToken(
   });
 }
 
+/**
+ * Hand a settled ask-user round to Core as an ordinary inbound submission.
+ *
+ * The answer travels the path a typed reply travels, so nothing downstream has
+ * to learn that a card produced it. A delivery that fails is logged and
+ * dropped, exactly as the inbound path treats a message Core would not take:
+ * re-delivering risks a second turn for one answer, and the user can say it
+ * again.
+ */
+async function deliverAskUserSettlement(
+  h: SessionHandle,
+  settlement: AskUserSettlement,
+): Promise<void> {
+  const { target } = settlement;
+  try {
+    const outcome = await h.delivery.deliver({
+      target,
+      containerChatId: target.kind === 'topic' ? target.chatId : null,
+      submission: {
+        attrs: {
+          chat_id: target.chatId,
+          ...(target.threadId !== undefined
+            ? { thread_id: target.threadId }
+            : {}),
+          ...(settlement.cardMessageId !== undefined
+            ? { message_id: settlement.cardMessageId }
+            : {}),
+          // Anyone in the chat may answer the card — an operator ruling, not a
+          // gap ("不需要限制，所有人都可以点"), so there is no check on who
+          // clicked. Carrying the clicker keeps the fact the model would
+          // otherwise lose.
+          ...(settlement.operatorOpenId !== undefined
+            ? { sender_id: settlement.operatorOpenId }
+            : {}),
+          ask_user_request_id: settlement.requestId,
+        },
+        text: settlement.text,
+        reminder: CHANNEL_REMINDER,
+        sourceId: settlement.sourceId,
+        anchor: {
+          chatId: target.chatId,
+          // The question card, never `sourceId`: this id is handed to Feishu as
+          // a COT presentation's origin, and a synthetic one would be sent to
+          // the API as if it were real. Empty when the card id is unknown,
+          // which the COT layer already treats as "no anchor".
+          messageId: settlement.cardMessageId ?? '',
+          target,
+        },
+      },
+    });
+    log(h).info(
+      {
+        dispatcher_id: h.opts.dispatcherId,
+        chat_id: target.chatId,
+        ask_user_request_id: settlement.requestId,
+        ask_user_outcome: settlement.outcome,
+        status: outcome.status,
+      },
+      '[ask-user] answer delivered',
+    );
+  } catch (err) {
+    log(h).error(
+      {
+        dispatcher_id: h.opts.dispatcherId,
+        chat_id: target.chatId,
+        ask_user_request_id: settlement.requestId,
+        err: errInfo(err),
+      },
+      '[ask-user] answer delivery failed',
+    );
+  }
+}
+
+/**
+ * Send a question card and return the round's id.
+ *
+ * Nothing is awaited beyond the send. The click that answers arrives on the
+ * card-action route, and the answer reaches Core as an inbound submission, so
+ * the tool that called this is long finished by the time the user decides.
+ *
+ * With no message to thread under, the target names the chat itself — all the
+ * tool is given, a question being addressed to a conversation rather than to
+ * one message in it.
+ */
+export async function askUserQuestion(
+  h: SessionHandle,
+  input: {
+    chatId: string;
+    questions: readonly AskUserQuestionSpec[];
+  },
+): Promise<{ request_id: string }> {
+  const target = h.targetRouter.outboundTarget(input.chatId, undefined);
+  const opened = h.askUser.open({ questions: input.questions, target });
+  const sent = await sendCard(h, {
+    target: h.targetRouter.notificationTarget(target),
+    card: opened.card,
+    mode: 'inbound',
+  });
+  const messageId = sent.messageIds[0];
+  if (messageId !== undefined) h.askUser.attachMessage(opened.requestId, messageId);
+  return { request_id: opened.requestId };
+}
+
+/**
+ * Close out a round that ran out of time.
+ *
+ * Both halves are best-effort and independent: the model is told there is no
+ * answer, and the card is repainted so the user is not left looking at a
+ * question that silently stopped working. A failed repaint must not cost the
+ * model its notification, which is why the patch is awaited separately.
+ */
+export async function expireAskUserQuestion(
+  h: SessionHandle,
+  expiry: AskUserExpiry,
+): Promise<void> {
+  await deliverAskUserSettlement(h, expiry.settlement);
+  const { messageId } = expiry;
+  if (messageId === undefined) return;
+  try {
+    await h.bot.editCard(messageId, expiry.card);
+  } catch (err) {
+    log(h).warn(
+      {
+        dispatcher_id: h.opts.dispatcherId,
+        message_id: messageId,
+        ask_user_request_id: expiry.settlement.requestId,
+        err: errInfo(err),
+      },
+      '[ask-user] expired card repaint failed',
+    );
+  }
+}
+
 export async function handleCardAction(
   h: SessionHandle,
   event: FeishuCardActionEvent,
 ): Promise<FeishuCardActionResponse | Record<string, never>> {
+  const applied = h.askUser.apply(event);
+  if (applied.kind !== 'ignored') {
+    if (applied.kind === 'settled') {
+      await deliverAskUserSettlement(h, applied.settlement);
+    }
+    return applied.response;
+  }
+
   const action = String(event.actionValue[DREAMUX_ACTION_KEY] ?? '');
   if (action !== DREAMUX_PAIRING_CARD_ACTION) return {};
 
--- a/packages/channel/feishu-channel/src/feishu-session-ops.ts
+++ b/packages/channel/feishu-channel/src/feishu-session-ops.ts
@@ -449,26 +449,31 @@ async function deliverAskUserSettlement(
  * card-action route, and the answer reaches Core as an inbound submission, so
  * the tool that called this is long finished by the time the user decides.
  *
- * With no message to thread under, the target names the chat itself — all the
- * tool is given, a question being addressed to a conversation rather than to
- * one message in it.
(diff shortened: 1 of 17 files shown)


packages/channel/feishu-channel/src/feishu-session-ops.ts
```diff
@@ -359,10 +369,151 @@ export async function approvePairingByToken(
...
+export async function askUserQuestion(
+  h: SessionHandle,
+  input: {
+    chatId: string;
+    questions: readonly AskUserQuestionSpec[];
+  },
+): Promise<{ request_id: string }> {
+  const target = h.targetRouter.outboundTarget(input.chatId, undefined);
+  const opened = h.askUser.open({ questions: input.questions, target });
+  const sent = await sendCard(h, {
```
YisenFE, line 471: **A failed send leaves the round armed, and the model hears from it 14 minutes later.**

`open()` arms the TTL timer before the card is sent, and nothing takes the round
back if this `await` throws. `sendCard` throws on any Feishu error — a `chat_id`
the bot is not a member of, an oversized card, a transport failure — and on a
stale fence. The tool call fails and the model is told so, but the round stays in
`rounds` with a live timer.

Fourteen minutes later `onExpire` fires and `deliverAskUserSettlement` injects
`The question card expired with no answer from the user. Stop where you are and
take no further action...` into the chat, for a card nobody ever saw. There is
not even a repaint to explain it, because `expiry.messageId` is undefined.

A model guessing a `chat_id` from a stale context is the ordinary way in. Closing
the round when the send fails would keep the registry's invariant that an open
round is a card someone can look at.

packages/channel/feishu-channel/src/feishu-session-ops.ts
```diff
@@ -359,10 +369,151 @@ export async function approvePairingByToken(
...
+ * one message in it.
+ */
+export async function askUserQuestion(
+  h: SessionHandle,
+  input: {
+    chatId: string;
+    questions: readonly AskUserQuestionSpec[];
+  },
+): Promise<{ request_id: string }> {
+  const target = h.targetRouter.outboundTarget(input.chatId, undefined);
```
YisenFE: **In a topic-mode group the answer never reaches the agent that asked.**

With `replyToMessageId` undefined this is unconditionally
`chatTarget(chatId, 'group')` — the tool takes no `message_id`, so unlike `reply`
it can never recover a topic target from the observed-message ledger. In a topic
group, which is exactly what a Collaboration Space is, that costs two things:

1. `notificationTarget` on a non-topic target is a bare `{ conversationId }`, so
   the card is a fresh top-level send. Per `.agents/domains/channel.md`, a
   top-level send to the container chat "in a Feishu topic group creates a new
   topic". The question does not appear in the conversation the Team is bound to.
2. The settlement is delivered against a `group` target, and `resolutionChain`
   for a group is one entry long — it can never match a `topic\0chat\0thread`
   binding key. The answer routes to the group's binding or to the Dispatcher.
   The topic-bound TeamLeader that asked was told by `next` to end its turn and
   wait, so it waits forever, while some other agent receives
   `The user answered the question card:` for a question it never asked.

`reply` already solves this by taking an optional `message_id` and feeding it
through this same `outboundTarget`. Giving `ask_user_question` the same argument
fixes both halves with the capability that exists, rather than a new one.

packages/channel/feishu-channel/src/feishu-session-ops.ts
```diff
@@ -359,10 +369,151 @@ export async function approvePairingByToken(
...
+}
+
 export async function handleCardAction(
   h: SessionHandle,
   event: FeishuCardActionEvent,
 ): Promise<FeishuCardActionResponse | Record<string, never>> {
+  const applied = h.askUser.apply(event);
+  if (applied.kind !== 'ignored') {
+    if (applied.kind === 'settled') {
+      await deliverAskUserSettlement(h, applied.settlement);
```
YisenFE: **The card repaint waits on the full inbound delivery.**

`applied.response` is the repaint Feishu shows the user, and it is now returned
only after `deliver` has resolved a route, possibly provisioned a Team, and had
Core admit the submission. Feishu's card-callback window is a few seconds; Team
provisioning is not bounded by it.

When it overruns, the round is already gone from the registry but the user still
sees the live card with nothing submitted. Clicking again gets
`这轮提问已失效，直接说你的想法就行。` — the answer was in fact delivered, and
nothing on screen says so.

Nothing in the response depends on the delivery: `deliverAskUserSettlement`
returns void and swallows its own errors. Detaching it would keep the repaint at
the speed of the registry rather than the speed of Core.

YisenFE (commented):
Read the whole change against `next`: the tool definition, the card builder, the
round registry, the ops wiring, the transport `editCard`, and the KB entry. The
design holds — a round that outlives the tool call is the right shape for a
question a person answers after lunch, the server-authoritative card is the only
way `interactive_container` can work, and moving `outboundTarget` onto
`FeishuTargetRouter` puts it with the ledger it reads.

Three things I would want settled before this merges, plus some minor notes.
Details are inline; the first two are correctness, the third is a user-visible
failure mode under load.

Minor notes, none blocking:

- `apply`'s trailing `return { kind: 'ignored' }` (`feishu-ask-user.ts`) is
  unreachable: all four members of `DREAMUX_ASK_ACTIONS` return earlier, and the
  membership test above already rejects everything else.
- `void sessionExpireAskUser(this.handle, expiry).catch(() => undefined)` in
  `feishu-channel.ts` guards a function that already catches both of its halves.
  One of the two is defense without a scenario.
- Submitting with nothing chosen is accepted and delivers
  `→ (left unanswered)` for every question. That is a decision either way — a
  legitimate "none of these" or a misclick that costs the round — but right now
  it is neither refused with a toast nor mentioned anywhere. Worth naming.
- The submitted card drops `config.summary`, which the live and closed cards both
  carry, so the notification line for the final state falls back to the default.

packages/channel/feishu-channel/src/feishu-session-ops.ts
YourWildDad: Fixed in 022f5195. Rather than roll the round back in a catch, `open` no longer registers it: it builds the round and the card and returns an `activate(messageId)` that the ops layer calls only once `sendCard` has resolved. A send that throws now leaves nothing in the map and no timer armed — "an open round is a card someone can look at" became a construction property instead of something a rollback has to restore. `attachMessage` is gone; recording the card id and starting the clock are the same act.

Test: `leaves nothing behind when the card never reached the chat` — opens without activating, asserts the click is answered as a dead round, then fires the manual timers and asserts `onExpire` never ran.

packages/channel/feishu-channel/src/feishu-session-ops.ts
YourWildDad: Fixed in 022f5195. `ask_user_question` now takes an optional `message_id`, mirroring `reply`, and it threads through `FeishuToolSession.askUserQuestion` into `outboundTarget(chatId, messageId)` — so a question that came out of a message is addressed into that message's topic, under the anchor the router already holds. With no `message_id` the old behaviour stands, which is right for a question that belongs to no particular message.

Kept `reply`'s rule that a message id from another chat is ignored rather than obeyed, and put it under test (`ignores a message from another chat rather than redirecting the card`) so the fallback reads as deliberate.

packages/channel/feishu-channel/src/feishu-session-ops.ts
YourWildDad: Fixed in 022f5195 — `void deliverAskUserSettlement(...)`, so the repaint returns at registry speed. `deliverAskUserSettlement` returns `void` and wraps its whole body in try/catch, logging a failure at error level, so there was never anything to report back here; the await only spent the callback window on waking an agent.

Left the comment honest about what the repaint does and does not confirm: it says the click was applied, not that Core took the answer. That was already true when the await was there, since the failure was swallowed either way.

packages/channel/feishu-channel/src/tools/ask-user-question.ts
```diff
@@ -0,0 +1,224 @@
...
+        ...nonEmptyString,
+        description:
+          'Feishu chat id from the inbound <channel source="feishu"> block. ' +
+          'The card is sent to this conversation.',
+      },
+      message_id: {
+        ...nonEmptyString,
+        description:
+          'Optional id of the message this question came out of. The card is ' +
+          'sent under it, in the same topic, exactly as a reply would be.',
```
YisenFE, line 172: Minor, and only the model-facing string — the routing this describes is right.

"sent under it" is not what happens on either path. `notificationTarget` returns
a bare `{ conversationId }` for a non-topic target, so in an ordinary group the
card threads under nothing and lands exactly where it would have without a
`message_id`. For a topic it does thread, but under
`targetAnchors.get(targetKey(target))` — the newest message this session has
seen in that topic — which is often not the message the model named.

The KB entry has this exactly right ("under the anchor the target router already
holds for it"); this description is the one place that promises the model a
`replyTo` it will not get. What the argument really buys is the topic, which is
the part that matters.

YisenFE (approved):
Re-reviewed the whole change, not just the deltas. All three findings are
fixed, and two of them are fixed better than what I asked for.

**A round exists only once its card does.** I suggested a rollback on send
failure; `open` / `activate` makes it a construction property instead, so
there is nothing to remember to undo. `open` builds the round and the card and
registers neither; `activate` — reachable only after the send lands — puts it
in play and starts the clock. The test drives the exact shape that was broken
(the send threw, `activate` was never reached) and asserts both halves: the
round is unanswerable, and no clock is running to report a question nobody was
asked.

**`message_id` closes the topic hole.** The card now follows the named message
into its topic, and the settlement is delivered against a topic target, so
`resolutionChain` reaches the topic binding and the answer gets back to the
TeamLeader that asked. The cross-chat case is `reply`'s rule and is now pinned
by a router test rather than left to read like an accident. Parsing keeps the
field absent rather than `''`, which matters because the router would look `''`
up as an id — nice that the test says so out loud.

**The repaint no longer waits on Core.** Detached, with the reason on the line.
`deliverAskUserSettlement` logs its own failures at error level and returns
void, so nothing was lost by not awaiting it.

The four minor notes are all addressed, and the empty-submit refusal went
further than the note asked: refusing rather than settling is the right call,
since a round settles exactly once and spending it on "everything unanswered"
is worse than saying nothing. Following it with a commit to make the toast
quote a label the card actually draws — one exported constant, asserted against
the card's own strings — is the kind of fix that cannot rot.

Docs keep up: the KB entry is precise about the anchor, the product catalog
gains the user-visible entry the behavior needs, and the change file mentions
`message_id`. CI green.

Approving. One inaccuracy inline, cosmetic and not worth another round trip.

Residual risk, unchanged from the original description and worth stating in the
merge decision rather than hiding: the MCP tool → card → click → inbound chain
and the expired-card repaint are still not exercised live. The specific shape I
would watch first is `action.input_value` on a form-less `input` — the free-text
path depends entirely on it, and nothing else in this repo reads that field.

PR #357, opened 2026-08-31
Fix Feishu COT conversation card lifecycle

YourWildDad:
## Summary

Give every TeamLeader, and the Dispatcher, one standing anchor and at most one open Feishu COT card, and close that card on the runtime's own native turn end rather than on a logical settlement.

- One recipient owns one anchor and one card, whichever Feishu chat supplied it; the target-keyed and per-turn dispatcher presentation modules are deleted
- Only a Channel user message Core reports as admitted moves an anchor; on that success the superseded card closes as interrupted and its successor opens under the new message. A Reply is outbound only
- Both built-in runtimes report one provider-neutral end per runtime-native turn through a new optional `nativeTurn` seam, and Core publishes it as the actor-scoped `teammate.native_turn.ended`. `teammate.turn.settled` no longer closes anything
- No source whitelist and no body-suppression ledger: once a recipient has an anchor, every projected input displays
- Projected operator paths are renamed rather than blanked — the workspace reads `.` and this host's home reads `~` — from prefixes resolved once at start and injected into the projection as a value

## Simplification round

A later pass deleted what review proved unnecessary: the Claude per-window deduplication flag (a synthesized end now fires only from the call that actually settled a still-open submission, so an ordinary success reports exactly one end), the Channel's per-turn body-suppression ledger, and the process-global home-prefix cache with its test reset hook and start-order dependency.

## Validation

Rush build, lint, and the full test suite with the real Codex live suite intentionally skipped; source and test TypeScript checks for the five affected packages; Rush change verification; task-record and knowledge-base checks; CI green on every pushed head.

## Accepted best-effort losses

Recorded with their reasoning in the task's verification record: facts published synchronously inside the admitting call may land on the predecessor card or nowhere; an exceptionally early native end may be ignored; a tool result crossing an anchor replacement may be dropped; a card whose I/O is still in flight when the session closes may be abandoned unfinished; a provider that omits the optional seam gets cards that never close.

Closes #360

PR #363, opened 2026-09-02
Keep internal paths out of published and committed artifacts

YourWildDad:
## Why

The beta release for `next` aborted at `release.yml`'s manifest gate ([run 33581749711](https://github.com/excitedjs/dreamux/actions/runs/33581749711)). Nothing was uploaded.

- The home-prefix resolver's docstring illustrated trailing-separator normalization with a literal absolute home path. That comment is compiled into `dist`, `dist` is packed, and the gate refuses any `/home/<user>/` token outside its allow-list. It now says the same thing with a `<home>` placeholder.
- The same round also used a real internal mount path and account name as the example home in the projection privacy tests. Tests are not packed, so the release gate could not have caught it — but it is a public-repository red-line violation on its own. It now uses the neutral fixture home the rest of that file already uses.

Assertions are unchanged; only fixture literals move. The workflow and its allow-list are deliberately untouched: a guardrail that fires is not something to edit around.

## Validation

Rush build, lint, and the full test suite with the real Codex live suite skipped; the ten package `tsc --noEmit` checks; knowledge-base check; `git diff --check`. The release gate's own scan was also run locally over built `dist` output and prints nothing.

No change file: this moves a comment and test fixtures, with no behaviour or published-contract change.

## Follow-up worth considering

CI has no equivalent scan. `gitleaks` is configured with three Feishu id rules and nothing for internal paths, so a path literal can reach the default branch and only surface at publish time — and one that never reaches `dist` never surfaces at all.

### DIFF
--- a/common/changes/@excitedjs/dreamux/fix-release-manifest-internal-paths_2026-09-02-02-13.json
+++ b/common/changes/@excitedjs/dreamux/fix-release-manifest-internal-paths_2026-09-02-02-13.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "comment": "Replace a literal absolute home path in a compiled docstring, and an internal mount path used as a test fixture, with neutral placeholders. Comment and fixture text only: no behaviour, published contract, or path semantics change.",
+      "type": "none",
+      "packageName": "@excitedjs/dreamux"
+    }
+  ],
+  "packageName": "@excitedjs/dreamux"
+}
--- a/packages/dreamux/src/platform/home-paths.ts
+++ b/packages/dreamux/src/platform/home-paths.ts
@@ -45,7 +45,7 @@ function lexicalHomePath(): string {
 }
 
 /**
- * A trailing separator is stripped so `/home/me/` and `/home/me` are one
+ * A trailing separator is stripped so `<home>/` and `<home>` are one
  * prefix — and so a home of `/` normalizes to the empty string and is dropped
  * rather than matching every absolute path in sight.
  */
--- a/packages/dreamux/tests/cot-projection-privacy.test.ts
+++ b/packages/dreamux/tests/cot-projection-privacy.test.ts
@@ -213,7 +213,7 @@ describe('conversation projection: secret redaction', () => {
   });
 
   it('treats a trailing period as prose punctuation for home and workspace paths', () => {
-    const home = '/data00/home/dyzhu';
+    const home = '/home/me';
     const cwd = `${home}/work/repo`;
     expect(
       redactText(`see ${home}. edit ${cwd}/a.ts.`, cwd, [home]).value,
@@ -221,7 +221,7 @@ describe('conversation projection: secret redaction', () => {
   });
 
   it('distinguishes dot-suffixed home siblings from workspace-adjacent siblings', () => {
-    const home = '/data00/home/dyzhu';
+    const home = '/home/me';
     const cwd = `${home}/work/repo`;
     const value = [
       `${home}.bak/notes.md`,
@@ -238,7 +238,7 @@ describe('conversation projection: secret redaction', () => {
   });
 
   it('renames workspace-adjacent siblings through the containing home prefix', () => {
-    const home = '/data00/home/dyzhu';
+    const home = '/home/me';
     const cwd = `${home}/work/repo`;
     expect(
       redactText(`${cwd}.git/config ${cwd}-old/x`, cwd, [home]).value,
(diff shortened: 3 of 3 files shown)


common/changes/@excitedjs/dreamux/fix-release-manifest-internal-paths_2026-09-02-02-13.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "comment": "Replace a literal absolute home path in a compiled docstring, and an internal mount path used as a test fixture, with neutral placeholders. Comment and fixture text only: no behaviour, published contract, or path semantics change.",
+      "type": "none",
+      "packageName": "@excitedjs/dreamux"
+    }
+  ],
+  "packageName": "@excitedjs/dreamux"
+}
```
ryanxiang7, line 10: Confirmed clean: this declaration carries **no `email` field**, so the generated changelog will not embed a
contributor address. Worth keeping an eye on — `rush change` writes that field from local git config, which is
how an unintended address reaches a committed artifact in the first place.

`type: none` is also the correct classification here: comment and test-fixture text only, no behaviour, published
contract, or path semantics change, so no version bump is warranted.

ryanxiang7 (approved):
# Review — PR #363 "fix(release): keep internal paths out of published and committed artifacts"

*Review by the **Devbox review seat**, against head `0dcafbd0`. Findings below are described
rather than quoted: this is a public repository, and restating the removed strings in a review
comment would re-publish exactly what the PR removes.*

## Verdict: APPROVE

The diff is comment- and fixture-only, no assertion was touched, and a full red-line sweep of
the branch tree is clean. CI is 8/8 green on the PR head. Merging this is safe.

## What I verified

**The diff is what it claims — with one correction.** It is **three** files, not two: the two
source/test files plus a new Rush change declaration (a required artifact, and itself clean —
see the inline note). The substantive changes are:

- `packages/dreamux/src/platform/home-paths.ts:48` — one docstring line, an absolute
  home-shaped example replaced with a neutral `<home>` placeholder. Comment only; the function
  below it is untouched.
- `packages/dreamux/tests/cot-projection-privacy.test.ts:216,224,241,249` — four
  `const home = ...` fixture assignments swapped to a neutral value.

**Zero assertion changes — verified, not assumed.** Every `expect(...)` body in those four
tests is byte-identical before and after; only the fixture *input* changed, and each
expectation is either a template derived from `home`/`cwd` or a literal that never contained
the old value. Because a fixture swap can silently change what a test proves, I did not rely on
inspection: I extracted the real `redactText` from this branch and re-ran all four cases under
the new fixture. **4/4 pass**, and each still exercises the same boundary property it was
written for — trailing-period prose punctuation, dot-suffixed home siblings versus
workspace-adjacent siblings, renaming a workspace-adjacent sibling through the containing home
prefix, and a doubled separator not being treated as a URL scheme boundary.

**Red-line sweep of the branch tree — clean.** I swept the whole tree at `0dcafbd0` (not just
the diff) for internal mount prefixes, the operator account name, private-relay and internal
mail domains, internal registry and host names, and real Feishu `ou_` / `oc_` / `cli_` id
shapes. The only surviving hits are **guardrail definitions naming the pattern they reject** —
the packaging red line in `.github/workflows/release.yml:356,365` and three knowledge-base
pages that document it, plus the rule bodies in `.gitleaks.toml`. A scanner has to name what it
scans for; those are correct and must stay. No real internal identifier remains anywhere on the
branch.

**Change-file hygiene.** The new declaration carries no `email` field, and `type: none` is the
right classification for a comment/fixture-only change.

## Scope note (not a defect in this PR)

This is a forward fix, and it is the correct one for the branch. For completeness, since it
bears on any follow-up decision: the neutralized test fixture value was introduced into `next`
by the #357 squash commit, so it is present in that branch's tree and history until something
other than a forward fix addresses it. That is outside this PR's scope and is not a reason to
hold it — landing this stops the content from propagating further and unblocks the release
gate. I am reporting the state, not recommending a remedy.

Also worth recording: the post-merge release run for #357 failed at its **"Pack and verify
manifest (no registry upload)"** step, with the version-bump and publish jobs skipped. The
packaging red line did its job — **nothing was published to the registry**. The exposure is
confined to the repository; no released artifact carries it.

## CI status

`gh pr checks 363` — **8/8 pass**, every check run bound to head `0dcafbd0`: commit-metadata
author-email, gitleaks (full history), knowledge-base link/orphan check, rush change
declaration check, rush bootstrap/build/typecheck/lint/test on ubuntu-latest and macos-latest,
and shellcheck on both. The earlier rush-change-declaration failure is resolved, and the two
`rush` jobs that were still pending when I started have since completed green — so the four
edited tests are verified by the real suite as well as by my direct re-run.

**Approved.**

PR #364, opened 2026-09-02
The Channel takes its own COT anchor, and hides its own body

YourWildDad:
## Why

Live testing after #357 showed two symptoms with one cause: the operator's own message was echoed into the card, and every message after the first opened a small card under the *previous* message that closed immediately.

Both come from waiting for Core before moving an anchor Core does not own. `teammate.turn.submitted` and the user body are published synchronously inside `team.submit`, before it returns, so anything the Channel does with the returned `turn_id` is already too late — the body lands on the predecessor anchor, opens a card there, and the anchor move then closes it.

## What changed

- **The Channel takes the anchor when it submits.** Nothing Core publishes synchronously can land on the predecessor. The optimistic anchor is retired only on an outcome that proves no turn exists — `rejected`, `failed`, `stopped`. `duplicate` keeps it, because the turn it names is the one already running; `ambiguous` and transport errors prove nothing and change nothing. The predicate is an allowlist so a future status cannot silently start retiring anchors.
- **The submitted fact returns the caller's own `source_id`.** That parameter already exists on `team.submit` and Core already deduplicates by it; it simply never came back out. The Channel recognizes its own submission on the submitted fact — which precedes the body in the same block — remembers the turn, and hides that turn's user body exactly once. Another producer's body still displays.
- **Recipient identity loses a redundant dimension.** The Channel knows a team name, or that it is submitting to the Dispatcher, and a Team has one leader, so the leader and agent names were doing no work. The session's live leader-name map goes with them.

Body-text matching was considered and rejected: the projected body is the submitted text modulo sanitization, so it would have failed exactly on messages containing a path or a secret.

## Design record

Requirement items 2 and 9, the rejected-designs list, and the operator adjudications behind each turn of this are in `.agents/tasks/channel/feishu-cot-conversation-cards/`. This knowingly supersedes the 2026-09-01 admission-gated-anchor rule; the reasoning is recorded there rather than left in a review thread.

## Validation

Rush build, lint, and the full test suite with the real Codex live suite skipped; ten package `tsc --noEmit` checks; `rush change --verify`; knowledge-base check; `git diff --check`; the release manifest gate's own scan over built `dist`.

The delivery-seam tests now emit the submitted fact and the user body **synchronously inside** the `team.submit` call, which is the ordering production actually has and which no test previously exercised — that gap is why this class of bug survived. They cover the receipt with no echo, another producer's body still visible, exactly one successor card for a second inbound, deduplication keeping the anchor, and each retiring outcome.


packages/channel/feishu-channel/src/feishu-cot-adapter.ts
```diff
@@ -142,36 +152,61 @@ export class FeishuCotAdapter {
...
+        release();
+        if (
+          this.closed ||
+          this.states.get(key) !== state ||
+          state.generation !== generation
+        ) {
+          return;
+        }
+        this.advanceAnchor(key, state, null);
+      },
```
ryanxiang7, line 197: **Answering your question directly: I agree with anchorless, and I would not change it.**

Returning to the previous anchor would be worse:

1. By the time this runs, the predecessor card was already closed as `interrupted` by the optimistic swap at `advanceAnchor`. Restoring the anchor would not restore that card — it would make the *next* activity open a fresh card under an *older* message, which is exactly the "small card under the previous message" symptom this PR removes.
2. All three rollback statuses mean the recipient provably is not taking the work (Team gone/closed, submission failed, runtime stopped), so there is no live output left without a home.
3. Restoring needs a predecessor-anchor stack with its own staleness questions — what if that target was unbound, or the fence retired it, in between? More machinery guarding a case with nothing to protect.
4. Anchorless is the honest state: the newest message produced no turn and the older one is stale, so there is no message the Channel can truthfully claim. It self-heals on the next inbound, and "a fact with no anchor produces no card" is already a recorded accepted loss.

The generation guard here is the right shape too — it is what stops a slow rejection from wiping an anchor a newer inbound already took. I also checked the `retire()`-then-`finally` double `release()`: it is idempotent, because `removeTurn` nulls `turnKey` and the `pending` branch bails on a `Set.delete` miss, so `size` cannot double-decrement.

packages/channel/feishu-channel/src/feishu-channel.ts
```diff
@@ -684,6 +680,12 @@ export class FeishuChannelSession {
   }
 }
 
+function submissionProvesNoAdmission(outcome: FeishuSubmitOutcome): boolean {
+  return outcome.status === 'rejected' ||
+    outcome.status === 'failed' ||
+    outcome.status === 'stopped';
+}
+
```
ryanxiang7, line 688: Confirmed this is a real whitelist, not a blacklist wearing one: only `rejected | failed | stopped` return `true`, and everything else — `submitted`, `duplicate`, `error`, and any status added later — falls through to `false` and is non-destructive **by construction**. That is the right default direction for a judgment whose wrong answer destroys an anchor.

The `duplicate` case is the one that matters most given at-least-once Feishu delivery, and the delivery test encodes it properly: after `duplicate`, activity from the still-alive original turn lands on the live card, so the receiver is never left anchorless mid-turn.

One structural cost worth **recording** (not fixing): a proven non-admission still leaves a visible stub — the optimistic receipt card, opened then closed as `interrupted`, under a message that produced no turn. On the `TEAM_NOT_FOUND` → Dispatcher fallback the operator sees that stub alongside the real Dispatcher card. It is inherent to taking the anchor before Core is invoked, which is the point of the change — but it is not in the accepted-loss list and belongs there beside the others.

ryanxiang7 (approved):
# Review — PR #364 "the Channel takes its own anchor, and hides its own body"

*Review by the **Devbox review seat**, against head `fa68061c`. Every claim in the change
report was re-derived from the code; nothing was accepted on description.*

## Verdict: APPROVE

The root cause is real, the fix addresses it at the right layer, and the two failure modes I
most expected — a mis-hidden body from another producer, and a correlation-token leak — are
both closed and covered by tests I could not break. CI is 8/8 green on the head, the base is
current `next` (with #363 in), and the red-line sweep is clean on both the diff and the branch
tree. My position on the anchorless question is **agree**, reasoned below.

## What I verified

**1 — The root cause is real, not a plausible story.** `EntityTurnCoordinator.attachSubmission`
(`turn-coordinator.ts:228`) calls `projectSubmitted` synchronously, and that projection
publishes the submitted fact and then the user body back to back
(`conversation-projection.ts:112-125`). `attachSubmission` runs inside the awaited admission
continuation (`turn-coordinator.ts:188`), so both events are on the bus **before** `team.submit`
resolves with a `turn_id`. Anything the Channel did with that `turn_id` was structurally too
late. Establishing the anchor before invoking Core removes the ordering problem instead of
racing it — that is the right call.

**2 — The rollback really is a whitelist.** `submissionProvesNoAdmission`
(`feishu-channel.ts:683`) returns true only for `rejected | failed | stopped`. Every other
value — `submitted`, `duplicate`, `error`, and any status added later — falls through to
`false` and is non-destructive by construction. `retire()` is additionally generation-guarded
(`feishu-cot-adapter.ts:187-197`), so a slow rejection cannot wipe an anchor a newer inbound
already took. Coverage is real: `it.each(['rejected','failed','stopped'])` in the delivery
suite exercises all three rollback paths, and the duplicate test asserts the important half —
after `duplicate`, activity from the still-alive original turn lands on the live card, so the
receiver is not left anchorless. That is the regression you self-caught, and the test encodes
it rather than merely naming it.

**3 — `source_id` end to end, and I could not construct a false hide.** The parameter
pre-existed on `team.submit` (`team-collection/commands.ts:226,244`) and Core already keyed
dedup on it (`admission-ledger.ts`); this PR only echoes it back. Timing holds: the submitted
fact is published immediately before the body in the same synchronous block, so recognition
genuinely precedes what it must hide. Suppression is exactly-once — `consumeTurn` removes the
token, so a repeated body for the same turn displays. Cross-producer safety is structural
rather than incidental: `submitted()` only binds a token if the echoed id is in **this
session's own** `pending` map, so a task, cron, system, or other-Channel turn can never acquire
one. The only way to mis-hide would be a deliberate id collision with this session's own
message ids, which is not a realistic producer.

**4 — The delivery test uses production timing, not a mock of it.** The `team.submit` handler
emits the submitted fact and the user body **synchronously, before returning** the admission
result — the exact shape that broke this in production. This is genuinely new coverage; the
previous suite could not have caught the bug.

**5 — Correlation lifecycle: no leak found.** I worked the five paths against
`feishu-inbound-anchor.ts` rather than reading the summary. Every token is removed by exactly
one of `consumeTurn` or `release`, and double-release is idempotent — `removeTurn` nulls
`turnKey` and the `pending` branch bails on a `Set.delete` miss, so the `retire()`-then-`finally`
double call cannot double-decrement `size`. Cap exhaustion fails open in the safe direction:
past 256 in flight, `begin` adds nothing and returns a no-op releaser, so suppression degrades
(bodies show) rather than leaking. A `submitted` for an already-released id is a no-op, and
duplicate `begin` on one id keeps independent tokens. The dedicated test suite covers each of
these, including the two I would have written myself — release when the fact never arrives, and
release when the body never arrives.

**6 — Redundancy removal is clean, including one detail worth calling out.** Collapsing the
receiver key to team-name / dispatcher is sound (one leader per Team, one Dispatcher per
session), and it incidentally makes a leader replacement keep its card lineage instead of
orphaning the anchor. The non-obvious part is handled correctly: `onRouteClaimed` switched from
a `startsWith` prefix match to an equality match, which it had to, because the new key has no
trailing separator — a prefix match would have silently stopped clearing route fences. The only
surviving `leaderName` reference is a test helper constructing a `team.state` event, which
legitimately still carries `leader_name`.

**7 — Text matching is genuinely absent.** It exists only as a rejected design in the
requirement, with the correct reason recorded (the projected body is post-redaction, so it
would fail exactly on the messages that matter). No residual comparison code anywhere in the
Channel source.

## Position on the anchorless-after-rollback semantic: agree

Returning to the previous anchor would be worse, for four reasons.

1. **It would resurrect the bug being fixed.** By rollback time the predecessor card has already
   been closed as `interrupted` by the optimistic swap. Restoring the anchor does not restore
   that card — it makes the *next* activity open a fresh card under an *older* message, which is
   precisely the "small card under the previous message" symptom this PR exists to remove.
2. **Nothing live is lost.** All three rollback statuses mean the recipient provably is not
   taking the work: the Team is gone or closed, the submission failed, or the runtime stopped.
   There is no in-flight output that needs a home.
3. **Restoring costs new state for no benefit** — a predecessor-anchor stack with its own
   staleness questions (what if that target was unbound, or the fence retired it, in between?).
   That is more machinery guarding a case with nothing to protect.
4. **Anchorless is the honest state.** The newest operator message produced no turn and the
   older one is stale, so there is no message the Channel can truthfully claim as the current
   anchor. It is self-healing — the next inbound re-establishes one — and "a fact with no anchor
   produces no card" is already a recorded, accepted loss.

One cost worth recording rather than fixing: a proven non-admission still leaves a visible stub
— the optimistic receipt card, opened and then closed as `interrupted` under a message that
produced no turn. On the `TEAM_NOT_FOUND` → Dispatcher fallback the operator sees that stub plus
the real conversation on the Dispatcher card. It is structural, not a defect: the card must exist
before Core is invoked, which is the whole point of the change. It is not in the accepted-loss
list, and it belongs there next to the others.

## Sweep

- **CI** — `gh pr checks 364`: **8/8 pass**, every run bound to head `fa68061c`. Base
  `770350b9` is the current tip of `next`, so #363 is in.
- **Test integrity — no weakening.** 14 `it(` blocks removed against 18 added, for a net −47
  lines and *more* tests. Six removals are the deleted `FeishuSubmittedTurns` class's own suite;
  the rest are replaced by contract-updated equivalents where the contract genuinely inverted
  (an ambiguous outcome now *keeps* the anchor; the Channel's own body is now hidden — and the
  companion assertion that another producer's body still displays was added alongside).
- **Red line — clean** on the diff and on the whole branch tree (internal mounts, account names,
  private-relay and internal mail domains, internal registry/hosts, real Feishu id shapes). Given
  this follows a leak incident I swept the tree, not just the diff.
- **Layering — holds.** `source_id` is an opaque caller-owned string Core already accepted and
  never parses; echoing it on the submitted fact teaches Core nothing about Feishu, and the type
  states it is not a routing, persistence, or presentation key.
- **Invariants — clean.** No synchronous IO added under `packages/*/src/**`, no `tsx`, no legacy
  rollback.
- **Changelog — compliant.** All three new declarations lead `BREAKING: Review:`; the 0.x
  packages take `minor` and `feishu-channel` (5.0.0) takes a real `major`.

**Approved.**

PR #365, opened 2026-09-02
Open a COT card with a flavour label, not a promise

YourWildDad:
## Why

The card's opening line said processing had started. It is written *before* Core is called, so it asserted something the Channel cannot know at that moment: an ambiguous submission may never reach the runtime, and the line claimed otherwise.

## What changed

The fixed line becomes one of five labels, chosen independently per card:

```
✳ Vibing...
✳ Shipping...
✳ Reticulating...
✳ Manifesting...
✳ Baking...
```

Cards are append-only, so this cannot animate — the label is picked once when the card opens. Selection is random rather than round-robin: a counter would be state, and a fixed sequence would invite reading meaning into the order.

The labels state nothing about admission or progress, which is the point: the line has to remain correct when nothing follows it.

## Scope

Presentation only. No anchor, terminal, suppression, or admission behaviour changes — an ambiguous outcome still keeps the anchor and adds nothing, because a runtime-native end already closes a genuinely broken turn's card, and otherwise the next inbound replaces the anchor.

`.agents/product/README.md` calls the Feishu COT card a deliberately tuned surface whose changes are their own requirement rather than a refactor side effect, so this is recorded as an intentional operator-visible presentation decision, taken by the operator.

## Validation

Rush build, lint, and the full test suite with the real Codex live suite skipped; ten package `tsc --noEmit` checks; `rush change --verify` against `origin/next`; knowledge-base check; `git diff --check`; the release manifest gate's own scan over built `dist`.

Every existing assertion on the opening line was kept and changed from equality to set membership — they pin that a freshly opened card starts with the opening line, and that contract is unchanged.

### DIFF
--- a/common/changes/@excitedjs/feishu-channel/fix-cot-receipt-does-not-promise_2026-09-02-04-43.json
+++ b/common/changes/@excitedjs/feishu-channel/fix-cot-receipt-does-not-promise_2026-09-02-04-43.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "comment": "Open a Feishu COT card with one of five flavour labels chosen independently per card, replacing the fixed line that said processing had started. That line was written before Core was called, so it asserted an admission the Channel could not yet know; an ambiguous submission may never reach the runtime. The opening line now states nothing about admission or progress. Cards are append-only, so the label is picked once when the card opens and does not animate. This is an intentional operator-visible presentation change to a deliberately tuned surface.",
+      "type": "minor",
+      "packageName": "@excitedjs/feishu-channel"
+    }
+  ],
+  "packageName": "@excitedjs/feishu-channel"
+}
--- a/packages/channel/feishu-channel/src/feishu-cot-adapter.ts
+++ b/packages/channel/feishu-channel/src/feishu-cot-adapter.ts
@@ -74,7 +74,13 @@ import {
 
 const FEISHU_COT_OPEN_TOOL_CALLS_MAX = 512;
 const FEISHU_COT_CLOSE_DRAIN_MS = 5_000;
-const FEISHU_COT_RECEIVED_TEXT = '已收到消息，开始处理。';
+export const FEISHU_COT_OPENING_LABELS = [
+  '✳ Vibing...',
+  '✳ Shipping...',
+  '✳ Reticulating...',
+  '✳ Manifesting...',
+  '✳ Baking...',
+] as const;
 
 export interface FeishuCotAdapterOptions {
   readonly dispatcherId: string;
@@ -209,15 +215,18 @@ export class FeishuCotAdapter {
     );
   }
 
-  /** The one line a card opens with, so the operator sees it was received. */
+  /** Pick one independent flavour label when this append-only card opens. */
   private openReceipt(key: string, state: CotState): void {
+    const label = FEISHU_COT_OPENING_LABELS[
+      Math.floor(Math.random() * FEISHU_COT_OPENING_LABELS.length)
+    ]!;
     this.acceptOpeningActivityForState(
       key,
       state,
       textMessageEvents({
         sourceId: `receipt:${randomUUID()}`,
         role: 'assistant',
-        content: FEISHU_COT_RECEIVED_TEXT,
+        content: label,
       }),
     );
   }
--- a/packages/channel/feishu-channel/tests/feishu-cot-delivery.test.ts
+++ b/packages/channel/feishu-channel/tests/feishu-cot-delivery.test.ts
@@ -32,16 +32,25 @@ import type {
 } from '@excitedjs/dreamux-types';
 
 import { FeishuChannelSession } from '../src/feishu-channel.js';
+import { FEISHU_COT_OPENING_LABELS } from '../src/feishu-cot-adapter.js';
 import { chatTarget } from '../src/routing/target.js';
 import { createFakeFeishuBot, type FakeFeishuBot } from './helpers/fake-feishu-bot.js';
 import {
   cotRunStatus,
   cotTexts,
   createFakeCotClient,
+  type FakeCotCard,
   type FakeCotClient,
 } from './helpers/fake-feishu-cot.js';
 
-const RECEIPT = '已收到消息，开始处理。';
+function expectOpeningTexts(
+  card: FakeCotCard,
+  expectedAfterOpening: readonly string[],
+): void {
+  const texts = cotTexts(card);
+  expect(FEISHU_COT_OPENING_LABELS).toContain(texts[0]);
+  expect(texts.slice(1)).toEqual(expectedAfterOpening);
+}
 
 let dir: string;
 let attachDir: string;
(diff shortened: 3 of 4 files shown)


packages/channel/feishu-channel/src/feishu-cot-adapter.ts
```diff
@@ -209,15 +215,18 @@ export class FeishuCotAdapter {
     );
   }
 
-  /** The one line a card opens with, so the operator sees it was received. */
+  /** Pick one independent flavour label when this append-only card opens. */
   private openReceipt(key: string, state: CotState): void {
+    const label = FEISHU_COT_OPENING_LABELS[
+      Math.floor(Math.random() * FEISHU_COT_OPENING_LABELS.length)
+    ]!;
```
ryanxiang7, line 222: **On random vs round-robin: I agree with random.**

Round-robin needs a counter — per-adapter mutable state that has to live somewhere and stay correct across card churn and adapter teardown — and a fixed cycle invites operators to read meaning into position ("Baking means it's slow"). Random is stateless, and because the tests assert set membership rather than a specific value they cannot flake on it. `Math.random()` is the right tool: display flavour, no uniqueness or unpredictability requirement.

Verified the properties that actually matter here: the pick happens once at card creation, nothing is stored on `CotState` (so there is no per-card state to leak or desynchronize), and since cards are append-only the label cannot re-roll on a later append.

One consequence worth naming out loud: independent per-card choice means two consecutive cards can show the same label. That is inherent to the design and fine — your change note already says "chosen independently per card", so it reads as stated rather than accidental.

common/changes/@excitedjs/feishu-channel/fix-cot-receipt-does-not-promise_2026-09-02-04-43.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "comment": "Open a Feishu COT card with one of five flavour labels chosen independently per card, replacing the fixed line that said processing had started. That line was written before Core was called, so it asserted an admission the Channel could not yet know; an ambiguous submission may never reach the runtime. The opening line now states nothing about admission or progress. Cards are append-only, so the label is picked once when the card opens and does not animate. This is an intentional operator-visible presentation change to a deliberately tuned surface.",
+      "type": "minor",
+      "packageName": "@excitedjs/feishu-channel"
+    }
+  ],
+  "packageName": "@excitedjs/feishu-channel"
+}
```
ryanxiang7, line 10: Verified at the blob level: the already-merged #364 declaration has blob SHA `980ce2ea…` on **both** the base and this head, so it was restored byte-identical rather than edited, and this PR contains exactly one new file and zero change-file modifications.

Your reasoning about the CI gap is correct and I can confirm it — the declaration check counts only *added* files (`git diff --diff-filter=A`), so editing an already-merged change file would have passed CI while silently altering the changelog for a change that was already declared. Restore-plus-new-file is the right handling.

This declaration itself is well formed: right package, no `email` field, no internal identifiers, and `type: minor` is correct — the package is past 1.0 so `minor` is legal, and since nothing breaks the `BREAKING: Review:` form correctly does not apply.

Worth flagging separately: the CI gap remains open as a repo-level hazard. Not a blocker here, but a good candidate for the guardrail backlog.

ryanxiang7 (approved):
# Review — PR #365 "the COT card opening line promises nothing"

*Review by the **Devbox review seat**, against head `6648be2f`. Small PR, same standard: every
claim re-derived from the diff and the tree.*

## Verdict: APPROVE

It is exactly what it claims — display copy, its tests, and one new change declaration. Nothing
else moved. CI is 8/8 green on the head, base is the current tip of `next`, and the red-line
sweep is clean on both the diff and the branch tree.

## What I verified

**Display-only — confirmed.** Four files. The entire source change is one constant replaced by a
five-element label list and one `content:` argument in `openReceipt`
(`feishu-cot-adapter.ts:77-83, 218-222`). Nothing in the anchor, terminal, suppression,
admission, routing, or projection paths is touched, and **zero files under `packages/dreamux`
are in the diff** — so the layering question does not even arise this round. No stale reference
to the old constant survives anywhere.

**Flavour implementation — matches the description.** Exactly the five labels listed, chosen by
one `Math.floor(Math.random() * length)` evaluated inside `openReceipt` at card creation and
passed straight through as content. Nothing is stored on `CotState`, so there is no per-card
state to leak or desynchronize, and because cards are append-only the label cannot re-roll on a
later append. `Math.random()` is the right tool here: this is display flavour with no uniqueness
or unpredictability requirement.

**Test contract — not weakened; slightly strengthened.** This was the thing worth checking, and
it holds. `expectOpeningTexts(card, tail)` is equivalent-or-stronger than the literal it
replaces: it still pins the exact tail *and* the total length via `texts.slice(1)`, and
constrains the first line to the five-element set instead of one string. So
`toEqual([RECEIPT, 'x'])` → `expectOpeningTexts(card, ['x'])` loses only the one thing that
genuinely changed, and still pins the real contract — a new card's first line **is** an opening
line.

I counted rather than trusted: in the delivery suite, **7 assertion lines removed against 8
added** — a net gain. The extra one is a genuinely new test covering an ambiguous
`ADMISSION_UNKNOWN` outcome, asserting the card opens with an opening line, stays open, and
keeps its anchor. That is the right test to add alongside this rationale: it proves the new copy
reads correctly in exactly the case the old line lied about. The `feishu-cot.test.ts` count is
12 usages before and 12 call sites after. **No assertion was deleted.**

**Change-file handling — verified at the blob level.** The already-merged #364 declaration has
blob SHA `980ce2ea…` on both the base and this head, so it was restored **byte-identical**, not
edited. The PR contains exactly one new file and zero change-file modifications. The new
declaration names the right package, carries no `email` field and no internal identifiers, and
`type: minor` is correct — the package is past 1.0 so `minor` is legal, and nothing breaks, so
the `BREAKING: Review:` form correctly does not apply here.

Your reasoning about the CI gap is right, and I can confirm it: the declaration check counts
only added files, so editing an already-merged change file would have passed CI while silently
altering the changelog for a change that was already declared. Restoring it and adding a new
file is the correct handling. Worth noting the gap itself is still open as a repo-level hazard —
not a blocker for this PR, but a good follow-up for the guardrail backlog.

## On the two things you flagged

**Random over round-robin — agree.** Round-robin needs a counter, which is per-adapter mutable
state that has to live somewhere and stay correct across card churn, and a fixed cycle invites
operators to read meaning into position ("Baking means it is slow"). Random is stateless, and
because the tests assert set membership rather than a specific value they cannot flake on it.
The one consequence worth naming out loud: independent per-card choice means two consecutive
cards can show the same label. That is inherent and fine — the change note already says
"chosen independently per card", so it is stated rather than accidental.

One observation, not a finding: this also shifts an operator-visible line from Chinese to
English. The change note frames it as an intentional change to a deliberately tuned surface, so
I am reading it as a product decision rather than an oversight — flagging it only so the choice
is a conscious one.

## Sweep

- **CI** — `gh pr checks 365`: **8/8 pass**, every run bound to head `6648be2f`. Base
  `56a7e5a5` is the current tip of `next`.
- **Red line — clean** on the diff and on the whole branch tree (internal mounts, account names,
  private-relay and internal mail domains, internal registry/hosts, real Feishu id shapes).
  Swept the tree rather than just the diff, as I have since the leak incident.
- **Invariants — clean.** No synchronous IO, no `tsx`, no legacy rollback.
- **Public surface — unchanged.** `FEISHU_COT_OPENING_LABELS` is exported from the adapter
  module for the tests but is not re-exported from the package index, so the pinned
  package-boundary surface is untouched.

**Approved.**

PR #366, opened 2026-09-02
Use a COT label glyph Feishu renders as text

YourWildDad:
## Why

The card's opening label shipped with U+2733 EIGHT SPOKED ASTERISK, which is in the Unicode emoji set. A terminal renders it as text; Feishu upgrades it to a coloured emoji tile, which is not what a terminal-styled progress line should look like beside its word.

## What changed

The five labels keep their words and their per-card random selection; only the leading glyph changes to U+274B HEAVY EIGHT TEARDROP-SPOKED PROPELLER ASTERISK, a dingbat with no emoji presentation.

```
❋ Vibing...     ❋ Shipping...     ❋ Reticulating...
❋ Manifesting...     ❋ Baking...
```

Chosen by the operator from candidates rendered in the client itself, which is the only reliable test for this — the character's Unicode category does not tell you what a given chat client will do with it.

## Validation

The focused Feishu COT suites pass unchanged (56 tests): they assert the opening line is a member of the exported label set rather than a fixed string, so the glyph is covered without pinning it. Full Rush build, lint, test, `tsc --noEmit`, `rush change --verify`, knowledge-base check and the release manifest scan run in CI.

### DIFF
--- a/common/changes/@excitedjs/feishu-channel/fix-cot-label-glyph_2026-09-02-05-40.json
+++ b/common/changes/@excitedjs/feishu-channel/fix-cot-label-glyph_2026-09-02-05-40.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "comment": "Render the COT card's opening label with a glyph Feishu shows as text. The previous character is in the Unicode emoji set, so Feishu replaced it with a coloured emoji tile; the replacement is a dingbat with no emoji presentation. Labels and behaviour are otherwise unchanged.",
+      "type": "minor",
+      "packageName": "@excitedjs/feishu-channel"
+    }
+  ],
+  "packageName": "@excitedjs/feishu-channel"
+}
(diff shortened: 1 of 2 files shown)


ryanxiang7 (approved):
Devbox seat review (operator-requested, no teammate round).

Verified the diff is exactly what the title says: the five COT opening labels swap `✳` (U+2733, default emoji presentation — Feishu renders it as a coloured emoji tile) for `❋` (U+274B, a dingbat with text presentation). Nothing else in the diff: 2 files, +15/−5, no anchor/terminal/suppression/admission behavior touched.

- Change file is well-formed: `@excitedjs/feishu-channel` (5.x line), `type: minor` is correct for a non-breaking display fix, no `email` field, comment accurately describes the glyph rationale.
- CI 8/8 green on the head (`b0cde0a`), including rush build/typecheck/lint/test on both OSes — so the label tests are in sync with the swap.
- Red line clean: no internal identifiers, paths, or tokens in the diff.

Approving.

PR #367, opened 2026-09-02
refactor(runtime): split the display line from push-back; make the anti-leak gate mandatory

YourWildDad:
This branch carries two changes. They are unrelated; say the word and I will
split the guardrail half onto its own branch.

---

# 1. Split the streaming display line from the push-back line

## Why

Two operator motives drove this task:

1. the push-back line is full of gates;
2. the COT display must reflect what the agent provider is actually doing right now.

Those two facts had one producer. The streaming display was keyed on
`RuntimeSubmission`, and a native turn's end was emitted from push-back
bookkeeping — so the card showed what push-back had *made of* the turn, several
gates after the provider had already said the turn was over.

## What changed

**The display fact is the Agent's, not the submission's.** `ChannelCoreEvent`
loses `teammate.turn.submitted` / `settled` / `message` / `tool_call` and
`teammate.native_turn.ended`; `teammate.activity` carries
`assistant.message | tool.call | turn.ended` for the Agent, and
`teammate.input` announces a submission at the moment it is submitted.

**A native turn's end now comes from the provider's own terminal, where the
provider reports it.**

- codex ends in `observeTerminal`, as its first statement, from the terminal
  alone: `completed` for a `CollectedTurn`, `failed` with the `Error`'s message.
- Claude Code ends in `handleProtocolEvent`'s `result` branch, ahead of
  attribution: claude's own `isError` and its `errors`/`subtype` text.
- The synthesized ends (codex `stop()`/`failProtocol`, claude
  `stopUnsettled`/`markTurnFailed`) end every native turn still open, whether or
  not a submission was left to settle.

Of the seven gates that stood between "codex says the turn ended" and
"`turn.ended` is emitted", two remain in front of the end and both belong to the
display line: the collector's line admission (`rememberTerminal`'s
duplicate/conflict verdict — codex contradicting itself must not paint two ends)
and the at-most-once flag, which moved off `NativeTurnRecord` to
`displayTurns: Map<turnId, 'open' | 'ended'>`. The display end reads no
`NativeTurnRecord` at all any more. Claude Code's four gates go the same way.

**Core still ends a card no runtime will ever speak for.** An input that reached
no runtime (stopped, skipped, ambiguous, failed) creates no `EntityTurn` and
guarantees no native end, so Core ends the open card as failed and says why.
That is the other half of the invariant "every published input ends", not a
leftover coupling.

## Behaviour changed deliberately

A codex turn whose encoded result cannot be restored, and a claude `result` no
started command can be attributed to, used to paint the card as a failed turn.
They now show the turn the provider reported; their submissions still settle as
failed.

## One trap the tests now pin

claude's legal lifecycle order is `started → result → completed`, so only
`started` and stream lines mean claude is working. Treating `completed` as news
re-opens a turn that already ended and makes teardown paint a spurious
`interrupted` end on every ordinary turn. The existing fakes do not emit
`completed`, so the whole suite stayed green through that regression.

## Gates

`rush build` / `lint` / `typecheck:tests` / `test` all green (test 1m56s,
including the live codex 0.151.0 integration), `.agents/scripts/check.sh` green,
change files present for every affected package.

## Records updated in the same change

`technical-design/final.md` departure 7 (rewritten around what was built, with
the gate table), the operator's rulings verbatim in the task `requirement.md`,
`.agents/domains/channel.md`, `.agents/domains/provider-runtime.md`, COT
requirement rules 12 and 14, and `.agents/product/README.md`.

---

# 2. Make the anti-leak gate mandatory, and scan the tree not just the tarball

## Why

The public-repo red line was enforced in exactly one place that could reject anything: the release job's audit of packed tarball contents. That gate only ever sees what reaches `dist`, so an internal mount path in a **test file** reached the default branch on 2026-09-02 and was found by hand, not by a gate.

Two holes, both closed here.

## The hook could stand aside

`common/git-hooks/pre-commit` warned and passed when gitleaks was absent, which on a machine without it is not a guardrail at all. It now refuses the commit and says installation is required and skipping is not permitted.

`common/scripts/install-gitleaks.sh` makes installing it one pinned, checksum-verifying command. The hook also looks in the directory that script installs into (`~/.local/bin`, or `GITLEAKS_INSTALL_DIR`), because it is not on every shell's `PATH` — a gate that fails right after a successful install is how people learn to reach for `--no-verify`.

## Nothing scanned the source tree

`.gitleaks.toml` cannot carry the internal path patterns: it is shared verbatim with the internal sibling repository, where those paths are legitimate. So `common/scripts/check-internal-content.sh` owns them instead, running over staged content in the hook and over every tracked file in a new CI job.

Its allowlist is by reviewed placeholder user name and by named pattern-definition file, **never by directory** — the leak that motivated this was in `tests/`, which a directory exclusion would have missed. The tree is clean today after rewording one prose false positive (`thread/home/bin/socket/stream` in `packages/dreamux/CLAUDE.md`) rather than allow-listing it.

The gate proved itself during development: the new test file, which necessarily feeds the scanner the tokens it must reject, was caught at commit time. It was fixed by assembling those tokens in code, not by opening a hole in the allowlist.

## Two things that fell out

- `gitleaks protect` and `gitleaks detect` are gone from 8.30.1's command list, surviving only as undocumented aliases. The hook and CI both move to `gitleaks git`, and CI installs through the same pinned script, so the version now lives in one place.
- The release `version` job commits the version bump through this hook **without** `--no-verify` on purpose, so it installs gitleaks first. Release automation is not exempt from the red line. It is the only job in any workflow that commits at all.

## A silent skip of its own, closed

A tracked path holding any non-ASCII byte comes back from git C-quoted by default (`"\346\265\213.md"`), so the content lookup found nothing and the file was passed over without a word. `core.quotePath=false` in both modes, verified against a staged `测试.md` carrying an internal path: caught, where before it was skipped. Both modes now read the indexed blob, which also collapses two read paths into one that cannot trip over a tracked symlink pointing at a directory.

## What is verified, and what is not

Verified locally: `rush build`, `rush lint`, `rush test` all green; `shellcheck` clean over both new scripts; the hook exercised in all three states (gitleaks present with a leak staged → blocked; gitleaks absent → hard failure with install instructions; clean tree → passes).

**Not verifiable before this PR:** the `internal-content` job and the rewritten `gitleaks` job have never run — feature-branch pushes do not trigger `ci.yml`. Those two jobs are what this draft exists to exercise.

**Not verifiable until the next real release:** the `release.yml` change. What to watch then is the "Install gitleaks for the mandatory pre-commit gate" step, and the version-bump commit passing the hook rather than being rejected by it.


common/changes/@excitedjs/agent-runtime-codex/refactor-split-display-from-pushback_2026-09-02-15-00.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "packageName": "@excitedjs/agent-runtime-codex",
+      "comment": "Report live activity through one agent-keyed sink: the native-turn sink is gone and a turn end is emitted as a `turn.ended` activity carrying codex's own error text. The per-turn pending-activity buffer is deleted, so thread items observed before a submission binds are reported instead of buffered and possibly dropped, and such a turn's end is reported too rather than dropped, so the card its items opened can close; orphan turns are still released from the collector. No config, state, or path shape changed and no rebuild is required. The end is now reported the moment codex reports the native turn's terminal, independent of the submission line: a turn codex only ever streamed items for is ended by a stop or a protocol failure instead of disappearing, and a turn whose encoded result cannot be restored is shown as the completed turn codex reported (the submission still settles as failed). The provider keeps no display state: a stop or a protocol failure reports one end without asking whether a turn was open, and a consumer with nothing open ignores it.",
+      "type": "minor"
```
ryanxiang7, line 6: **Minor — this note should lead with `BREAKING: Review:`.**

Its own text describes a breaking provider-contract change: "the native-turn sink is gone" and "the per-turn pending-activity buffer is deleted". Two things make the omission look like an oversight rather than a judgement:

- the symmetric `@excitedjs/agent-runtime-claude-code` declaration in this same PR **does** lead with `BREAKING: Review:` for the equivalent change;
- the `@excitedjs/dreamux-types` declaration states that *every* Agent Runtime provider must be rewritten against the single activity sink — which is exactly what this package implements.

`type: minor` is right (0.x line, so `major` is forbidden) and the content is thorough; only the leading token is missing. Everything else in the six declarations checks out.

.github/workflows/ci.yml
```diff
@@ -264,29 +268,45 @@ jobs:
...
+    # same scan the pre-commit hook runs over staged changes and the release
+    # workflow runs over packed tarballs.
+    #
+    # Tree-wide rather than diff-wide on purpose: the release gate only ever
+    # saw what reached `dist`, which is why an internal path in a test file
+    # reached the default branch on 2026-09-02 without a gate firing.
+    steps:
+      - uses: actions/checkout@v4
+      - name: check-internal-content (every tracked file)
+        run: common/scripts/check-internal-content.sh --tree
```
ryanxiang7, line 312: The gate itself checks out, and I want to be precise about what "mandatory" buys, because two separate things are involved.

**Mandatory within CI — yes.** No `continue-on-error`, no `if:`, no fork skip, no path filter, and `--tree` walks `git ls-files`, so tests and fixtures are covered. That is exactly the surface the tarball-only release gate could never see, which is how an internal path in a test reached the default branch. Two details worth calling out as good: `/data00/` requires the trailing slash so prose naming the mount does not false-positive, and `is_pattern_definition` allowlists three files **by name, never by directory** — the right lesson from a leak that lived in `tests/`.

**Merge-blocking — no, and not because of anything in this PR.** `next` currently has **no required status checks at all**: branch protection defines none, and the `release-branch-pr-gate` ruleset carries only the one-approval `pull_request` rule. So this job always runs and fails the workflow, but a red CI would still not stop a merge — for this job or the other eight. A workflow file cannot fix that; adding the checks to the ruleset is a repo-settings action, and now that the scan exists it is worth doing.

**One claim outruns the patterns.** The comment above describes the guardrail as covering "private-mirror URLs", but neither this scanner's `FORBIDDEN_RE` nor `.gitleaks.toml` matches a hostname or registry URL. I swept the tree and it is clean of them today — the gap is that nothing would catch one tomorrow. Either add the pattern or narrow the sentence.

ryanxiang7 (approved):
# Review — PR #367 "split the display line from push-back; make the anti-leak gate mandatory"

*Review by the **Devbox review seat**, against head `5429a864`. 80 files; every substantive
claim re-derived from the code rather than the PR body.*

## Verdict: APPROVE

The refactor is the right fix for the right reason, and it is the cleanest of this series: the
display fact becomes the Agent's, the catalog shrinks from seven kinds to four, and `turn.ended`
folds into the activity vocabulary so a runtime that learns a new fact changes no catalog, no
seal, and no subscription. That is "prefer a capability over a special case" applied properly.
The end now comes from the provider's own terminal ahead of attribution, which is exactly the
layering the previous rounds kept working around. The guardrail half is real and closes the gap
the 2026-09-02 incident went through. CI is 9/9 green. My findings are all minor or
observational; none is a correctness defect, and none blocks the beta.

## What I verified

**Merge safety against the newer `next`.** The base is one commit behind `next` (#362, the
ask-user card), and CI ran on the PR head rather than a merge commit — so I checked the
interaction by hand instead of trusting `mergeable: true`. It is clean on both axes: #362's new
code references none of the events this PR deletes, and the two `channel.md` edits are in
disjoint regions (#362 at the tools section, this PR from the event catalog down). This PR does
not touch `feishu-channel.ts` at all, which is where #362 did its work.

**Event contract — provider-neutral, no leakage.** `TeammateActivity`'s members mirror
`RuntimeActivity`'s deliberately, so there is one vocabulary rather than two, and Core learns
nothing about any channel or runtime. The `source_id` doc is now honest in a way it was not
before — "its mere presence proves nothing, because cron fires, task push-backs, and restart
notices carry one too" — which closes the misuse the previous shape invited. Nothing consumes
the removed `teammate.turn.*` kinds; the only remaining references are in
`core-event-catalog.test.ts`, where they are asserted to be **rejected** by the seal. That is
the right way to retire a catalog entry.

**Push-back survived the split.** `turn-coordinator.ts` loses 263 lines and its docstring now
says it "holds the push-back line and nothing else"; `deliverCompletion` / `ensureDelivery` are
intact. The deletion is the display bookkeeping leaving, which is the whole point.

**Provider terminals — never missing; extras are harmless by design.** This is a deliberate
shift from "exactly once" to "never missing, and a duplicate is absorbed", and it is coherent
end to end. Codex ends in `observeTerminal` as its first statement; claude ends in the `result`
branch ahead of `completeStartedGroup`. Synthesized ends are guarded by *runtime* conditions
rather than the old settle-guard: claude reports a stop or fence end only when a live child
exists (`session !== null`) and skips `markTurnFailed`'s end after `stop()` already reported;
codex's `failProtocol` reports only on the first failure. I walked the termination paths —
success, result-with-error, window failure, stop, fence, protocol loss, and a start that failed
leaving no child — and found none without an end and none that can lose one. Duplicates are
absorbed at the consumer: `finishCard` opens with `if (state.active === null) return`.

The completeness hole is closed from the other side too: Core publishes its own `turn.ended`
for an input no runtime ever accepted, so a surface that opened always has something to close
it. `reason` is new displayable text carrying provider error strings, and I checked it is
sanitized through the same `sanitizeText` path as everything else — no new leak surface.

**Anti-leak gate — genuinely mandatory within CI, and it closes the real gap.** The new job has
no `continue-on-error`, no `if:`, no fork skip and no path filter, and `--tree` walks
`git ls-files`, so test files and fixtures are covered — which is precisely what the tarball-only
release gate could never see. Two details I liked: `/data00/` requires the trailing slash, so
prose that names the mount does not false-positive, and `is_pattern_definition` allowlists three
files **by name, never by directory**, with the reason stated ("the leak this gate exists for was
in `tests/`"). The `core.quotePath=false` handling closes a silent-skip the scanner would
otherwise have had. The pre-commit hook now fails hard when gitleaks is absent instead of
warn-and-passing, so local and CI agree.

**Test integrity — a contract swap, not a weakening.** 37 `it(` blocks removed against 51
added. The removals are all things the redesign makes false — representative attribution, the
"drops live activity with no started command" behaviour that was the bug, and the two tests that
pinned the old exactly-once end. Each has a named replacement pinning the new contract, including
`reports the teardown end after a turn already completed: the runtime holds no display state`,
which asserts the new behaviour head-on rather than quietly dropping the old assertion. New
coverage reaches the cases that motivated the change: activity no started command could have
owned, a turn codex only ever streamed items for, an end while an admission is still in flight,
and Core announcing then ending a submission whose runtime never started.

**Parked gap — resolved, not left.** The claude-side silent drop is gone: `emitStreamActivity`
no longer looks up a submission at all, and its docstring names the old failure exactly
("anything claude emitted before a command's `started` lifecycle arrived, the fact was dropped
entirely"). Codex's orphan turns now report ends rather than disappearing. The remaining
`if (active === null) return` in `onProtocolEvent` is a different and inherent guard — no window
means no turn to report on.

**Sweep.** Red line clean on the diff and the tree, including the classes the scanner does not
itself cover. `mkdirSync` in the diff is in `tests/codex-live.test.ts`, outside the
`packages/*/src/**` no-sync-IO rule. No `tsx`, no legacy rollback, path contracts intact.
Change files: all six present, 0.x packages on `minor`, `feishu-channel` (5.x) on a real `major`,
the guardrail half separately declared `type: none`.

## Findings

**Minor — the codex change note does not lead with `BREAKING:`.** It describes a breaking
provider-contract change ("the native-turn sink is gone", "the per-turn pending-activity buffer
is deleted"), its symmetric `agent-runtime-claude-code` sibling leads with `BREAKING: Review:`,
and the `dreamux-types` note says every Agent Runtime provider must be rewritten against the
single activity sink. Under-marked relative to both. Inline.

**Minor — `open.larkoffice.com` is newly introduced** (six places under `.agents/`; zero
occurrences on `next` today). It is a vendor documentation host rather than an internal
hostname, and the new scanner does not treat it as forbidden — so I am raising it as a conscious
decision rather than a violation. It does signal which Lark tenant the team reads docs on, and
it lands in the same PR that tightens the public-repo red line, so it is worth an explicit
yes/no rather than arriving by default.

**Observation — the gate is mandatory in CI but nothing is merge-blocking.** `next` currently
has **no required status checks at all**, in neither branch protection nor the
`release-branch-pr-gate` ruleset; that ruleset carries only the one-approval rule. So the new job
always runs and fails the workflow, but a red CI would still not stop a merge — for this job or
any of the other eight. A workflow file cannot fix that; adding the checks to the ruleset is a
repo-settings action, and worth doing now that the scan exists.

**Observation — one guardrail claim outruns the patterns.** The `ci.yml` comment describes the
anti-leak guardrail as covering "private-mirror URLs", but no pattern in either the scanner or
`.gitleaks.toml` matches a hostname or registry URL. The tree is clean of them today; the gap is
that nothing would catch one tomorrow. Either add the pattern or narrow the sentence.

**Nit.** `packages/agent-runtime/claude-code/tests/runtime.test.ts:531` still describes the
fact as the one "Core turns into `teammate.native_turn.ended`" — an event this PR deletes.

## On splitting the two halves

Land them together this time, and split by default in future. The reason it is tolerable here is
specific and checkable rather than a shrug: the two halves share **zero files**. The guardrail is
two scripts, two workflows, contributor docs and its own `type: none` declaration; the refactor
is `packages/**` plus the knowledge base. So the usual cost of bundling — that reverting one
drags the other — is nearly absent; either half can be reverted by path. The residual cost is
real but small: a post-beta revert of the refactor would take the guardrail with it unless
someone cherry-picks, and the guardrail is the half you would least want to lose. Given it is
already reviewed and green here, re-running the gate to separate them buys less than it costs.

## CI status

`gh pr checks 367` — **9/9 pass**, every run bound to head `5429a864`, including the new
`internal-content — public-repo path scan (tree)`. No `[skip ci]` in any commit. Base is one
commit behind `next`; I verified that gap is semantically clean above. The real Codex live suite
is unchanged, so the issue #63 gate is intact.

**Approved.**

PR #370, opened 2026-09-02
Make the managed-daemon self-upgrade SOP runnable

YisenFE:
## Why

The bundled self-upgrade reference added by #311 was 363 lines and 18 steps, and its main path could not complete.

- Step 10 required transferring an inventory of staging directories, rehearsal prefixes, backups, and rollback commands to *"an independently executing operator/controller"* over a private path, and required an acknowledgement before the first live mutation. Without one, its own instruction was to clean up, refuse the upgrade, and stop.
- Step 8 handed every action that stops or quiesces the daemon to that same actor.

A Dispatcher is the only actor present when an upgrade arrives through a Channel, and that independent controller is not a role this product defines — so the ordinary path terminated in refusal.

The staging and rehearsal machinery (two `npm pack` operations plus four global installs into private prefixes, purely to prove a rollback closure) was defense with no named failure scenario behind it, which this repository deletes rather than corrects.

## What

**Four steps over surfaces the CLI already has**, replacing 363 lines with 67:

1. install the requested version (default `latest`), recording `oldVersion` / `targetVersion`;
2. `dreamux changelog --json`, review `(oldVersion, targetVersion]` oldest-first, apply each applicable migration through its routed owner;
3. `dreamux doctor --json`, repair until it passes, never restart before it does;
4. `dreamux daemon restart --notify-resumed --dispatcher <id>`, then report through the originating Channel.

**Knowledge owners realigned**: the `dispatcher-skill` and `model-facing-writing` domain pages, and the maintenance-synchronization clause in `CLAUDE.md`, which described the exception as reading a *staged* target's changelog.

**Root-skill residue removed**: the mutation gate and Secret Safety list still stopped on a `private recovery` condition that no reference defines once the staged SOP is gone. Both now state only what a routed owner actually requires.

**Deliberately not done**: the archived progressive-disclosure proposal is left intact — it is the decision trail for what #311 chose, and this change is accountable to that history. The `independent operator` wording in the Feishu access and service lifecycle references also stays; each states a fact its own owner still needs.

## Alongside

- **Development workflow fast path** restated as a minimal-change test rather than a literal one-file test. The risk-bearing criteria are unchanged; only the file count stops being the gate, and a high-risk change still takes a reviewed path however small its diff.
- **`.agents/scripts/check.sh` fixed.** It could not run on stock macOS bash: bash 3.2 re-scans a process substitution body and reads a `case` pattern's closing bracket as the end of the substitution, so the script failed to parse and the knowledge gate this change had to pass was unrunnable locally. That one `case` becomes a parameter-expansion prefix test. The defect predates this change; CI runs bash 5 and never saw it.

## Validation

| Gate | Result |
|---|---|
| `rush build` | SUCCESS, 8 operations |
| `rush lint` | SUCCESS, 7 operations |
| `rush test` | SUCCESS, no failure |
| `.agents/scripts/check.sh` | `KB OK (147 files reachable from root.md)` |
| `shellcheck .agents/scripts/check.sh` | clean |
| `init_task.py check` | Task record OK |
| `git diff --check` | clean |

Targeted: `feishu-allow-chats-release-contract` and `bundled-skill-sources` pass (13 tests) — the former locks the `references/self-upgrade.md` and `current-state-only` strings in `CLAUDE.md`, both preserved.

## Limitations

- Documentation and bundled skill text only. No CLI, daemon, restart-marker, or config/state behavior changes; no `packages/*/src` and no test file is touched.
- The lean SOP has no rollback step. That is an accepted consequence, not an oversight: a failed install is diagnosed before any restart, and doctor must pass before the daemon is restarted at all.
- Change file is `patch` for `@excitedjs/dreamux`; not breaking, so no `Rebuild:` note.

Task record: `.agents/tasks/builtin-skills/adopt-lean-self-upgrade-sop`


packages/dreamux/skills/dispatcher/dreamux-maintenance/references/self-upgrade.md
```diff
@@ -1,363 +1,67 @@
...
-downgrade requires a separate independently reviewed recovery plan; explicit
-version selection does not opt into the forward-only algorithm.
-
-### 6. Stage and validate exact old and target artifacts
-
-Before overwriting the live service prefix, create a private temporary staging
-directory and capture the npm executable, global prefix, and an explicit
-isolated cache path. In the commands below, `npm` is that captured executable.
-Run these two operand-complete commands:
+Use the same Node/npm environment and global prefix as the managed service. When
```
YourWildDad, line 14: [P1] Keep all four steps on the managed install

This has a concrete supported trigger: onboarding may pin `service.execStart[0]` to an absolute local/npx launcher while the runtime shell's `npm` and `dreamux` resolve a different global copy. Step 1 then installs and reads B, but Step 4 restarts service A; A can resume and emit the notice, so the Channel reports B's version while the daemon is still A. The minimal fix is one guard, not the old preflight: resolve the managed launcher/environment once before mutation, stop if that launcher is not in the selected global prefix, and run the existing four steps under that authority. No staging, rollback, or new recovery mechanism is needed.

packages/dreamux/skills/dispatcher/dreamux-maintenance/references/self-upgrade.md
```diff
@@ -1,363 +1,67 @@
...
-targetVersion]` and order entries by semver from oldest to newest. Do not
-inspect only the newest entry or execute storage order. The `--json` flag used
-later changes only the output format and is not a range selector; range
-selection belongs to the reader.
-
-### 8. Classify live-safe and independent-quiesced work
-
-Classify all applicable `BREAKING:`, `Review:`, and `Rebuild:` instructions
-before touching the live prefix, config, or state.
+Require the install to succeed, then run `dreamux --version` again and record the
```
YourWildDad, line 21: [P1] Reject a lower target before this install

A named version—or `latest` when the current install is a newer prerelease—can be lower than `oldVersion`. This command then downgrades the live prefix before the later forward-only `(oldVersion, targetVersion]` range can be evaluated. Keep this inside Step 1: resolve the exact selector and stop unless it is greater than `oldVersion` (equal can be a no-op). No additional phase is needed.

packages/dreamux/skills/dispatcher/dreamux-maintenance/references/self-upgrade.md
```diff
@@ -1,363 +1,67 @@
...
-  pre-acknowledged independent owner. Never expose them through a broad Channel.
-- If the restart notice does not arrive, the stopped or reaped Dispatcher
-  cannot self-diagnose. The independent operator must run `dreamux status`,
-  inspect daemon logs, and contact the user.
-- Foreground `dreamux serve` is not silently treated as a managed daemon. It
-  needs an external stop/start and recovery path.
-- On any rollback that cannot be proven, use the final confirmed-stop handoff
-  in step 12. The independent operator owns stop confirmation, exact-marker
-  cleanup with `ENOENT`-only absence proof, and recovery restart authority.
+Use the root skill's Service lifecycle route if restart fails or the notice does
```
YourWildDad, line 66: [P2] Remove the unreachable no-notice instruction

If Dispatcher resume/startup fails, notice injection never creates a model turn, so this Dispatcher cannot observe the absence and load another route. The lean fix is deletion: keep Service lifecycle routing only for a synchronous restart failure while the caller survives, and remove `or the notice does not arrive`. Do not add a controller or recovery protocol.

.agents/tasks/builtin-skills/adopt-lean-self-upgrade-sop/README.md
```diff
@@ -0,0 +1,48 @@
...
+
+- Pull request / CI / merge: Not started.
+- Knowledge closeout: Complete. Owners touched and results:
+  - `.agents/tasks/**` — this task record is the decision record: requirement,
+    final solution, operator rulings, and verification.
+  - `.agents/domains/dispatcher-skill.md`, `.agents/domains/model-facing-writing.md`
+    — updated to the four-step contract.
+  - `CLAUDE.md` — maintenance-synchronization clause updated to the installed
+    target's changelog.
+  - `.agents/product/README.md` — `N/A`: the catalog's upgrade entry owns
```
YourWildDad, line 36: [P2] Record the changed Channel outcome outside the runtime skill

The old shipped guide refused this request without an independent controller; the new guide executes install/migration/restart and reports back. That is a user-visible outcome, and the repository catalog says missing behavior is added in the change that touches it. This does not require more SOP machinery: one catalog entry with the deciding-record pointer is sufficient.

YourWildDad (dismissed):
The lean four-step direction is good, but the current procedure can restart the wrong installed copy, accept a downgrade as an upgrade, and has no live actor capable of executing its no-notice fallback. These are reachable correctness failures in the shipped model-facing SOP, so I am requesting changes. I also left one knowledge-closeout comment because this changes a Channel-visible product outcome.

YourWildDad (approved):

PR #372, opened 2026-09-02
feat(channel): support dispatcher-scoped commands

YourWildDad:
## Summary

- allow Channel providers to declare immutable, dispatcher-scoped command catalogs
- register those commands in the canonical Core command registry with shared input/output validation, canonicalization, and adapter behavior
- add per-channel admission, draining, rollback, and revocation across Channel lifecycle transitions
- expose non-sensitive Channel command descriptors through `dispatcher.status` without materializing dormant dispatchers

## External provider contract

External Channel providers depend only on `@excitedjs/dreamux-types`. A Channel-authored business refusal is represented by the command's declared output schema and returned as an ordinary result. A thrown error is an implementation fault and is normalized to `INTERNAL`; Core does not define or import Channel-specific business error vocabularies.

Command names use `channel.<encoded channel_id>.<local_name>`, where the host applies the documented injective channel-id encoding. Registration is atomic per Channel catalog, and stopping or rolling back a Channel closes admission before draining in-flight executions and unregistering definitions.

## Validation

- `rush build`
- `rush lint`
- `rush typecheck`
- `rush typecheck:tests`
- `DREAMUX_SKIP_LIVE_CODEX=1 rush test`
- Dreamux package tests: 68 files passed, 1045 passed, 1 skipped
- `rush smoke-built-cli`
- `.agents/scripts/check.sh`
- `git diff --check`

The live Codex compatibility test was intentionally skipped by `DREAMUX_SKIP_LIVE_CODEX=1`; the non-live suite and both admin-socket/in-process Channel command adapters passed.

### DIFF
--- a/.agents/domains/channel.md
+++ b/.agents/domains/channel.md
@@ -11,11 +11,13 @@ selection, or the Core event source a Channel subscribes to.
 ## Ownership
 
 A Channel provider owns platform I/O, inbound normalization, target resolution,
-provider-specific tools, message ownership facts, and **routing**: which
-conversation reaches which Team, and what a Collaboration Space is. Dreamux core
-owns Channel session lifetime, the Command port a Channel invokes, the scoped
-Core event source it subscribes to, and generic MCP forwarding. Core holds no
-binding table, no target model, and no Collaboration Space container.
+provider-specific tools, its own Command definitions, message ownership facts,
+and **routing**: which conversation reaches which Team, and what a Collaboration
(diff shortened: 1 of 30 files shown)


YourWildDad:
Closing as obsolete after #350. The dynamic `channel.<id>.<name>` registration/admission/drain layer conflicts with the current fixed, domain-owned Core command registry. The concrete Feishu slash-command work in #379 uses the existing Core command and Channel MCP boundaries instead of reviving dynamic registration.

PR #375, opened 2026-09-03
Start the TeamLeader lazily and surface its start failure on the card

YourWildDad:
## Problem

After #367, team `flow-codex-gbr211nj`'s TeamLeader could not start codex
(`no rollout found`), yet the Feishu COT card stayed on "Vibing..." with no
error. Two gaps:

1. The leader submit paths pre-started the runtime in `ensureRouteReady()`,
   outside the entity's admitted-input span, so a start failure had no
   `turn.ended` to ride and the card never learned of it.
2. `TeamService.createNew` called `leader.activate()` at creation. That opened
   a codex thread with no turn; codex writes a rollout only at the first turn,
   so the thread could never be resumed after a restart.

## Change

- **Start the TeamLeader inside the submission that needs it.** Delete
  `ensureRouteReady()`; Channel inbound, the Team MCP `send`, and cron all go
  through `submitToLeader`, and the runtime starts inside
  `TeammateService.submitAdmitted`. A start failure now projects
  `turn.ended { status: 'failed', reason: <provider message> }`, which the
  Feishu adapter prints on the card before ending it as a failure.
- **Start no TeamLeader runtime at Team creation.** Delete the creation-time
  `activate()`. A promptless `team.create` writes the record and nothing
  else; the leader's runtime starts with its first real turn (bound-channel
  inbound or a later Team MCP send). Laziness already lived at the
  entity→provider boundary, so no new mechanism is added.

Operator-visible changes (recorded in `.agents/product/README.md`):

- `team.create` without a prompt no longer proves the runtime can start.
- A promptless Team's `leader_state` stays `starting` until its first turn,
  across restarts.

Residual window, deliberately undefended: codex killed between
`thread/start` and its first rollout, or a cleared `~/.codex/sessions`, still
fails resume. The card now shows the error; recovery is dissolve and
recreate.

## Tests

- New `team-leader-start-failure.test.ts`: the provider's start error reaches
  `turn.ended` with the provider's text; the record stays `starting`. Fails
  on the old code.
- New `team-leader-lazy-start.test.ts`: promptless create makes zero
  `createRuntime` calls; the first submission orders
  input → createRuntime → start → submit; a prompt at creation follows the
  same order; a start error rejects with the provider message.
- Harness: `mockLeaderActivationResolved` removed; `mockLeaderSubmission` /
  `mockLeaderSubmissionRejected` added. Idempotency, read-path, and
  dissolve-recovery tests updated to the new contract.

`rush build`, `lint`, `test`, `typecheck:tests`, and `.agents/scripts/check.sh`
all pass. Two patch change files included.

Not done: live verification (daemon restart, one message to a fresh
promptless Team, restart, second message to prove resume).

https://claude.ai/code/session_01WxU1pdZzxGWNh1gGYiXQ1B


ryanxiang7 (approved):
Approving. I verified the parts that the deletion of `ensureRouteReady()` could have quietly dropped, and each one is covered elsewhere.

**What I checked**

- **Closed/closing guard.** `ensureRouteReady()` threw on a `closed` record. `submitToLeader` already runs inside `admit()`, which refuses both `dissolveTask !== null` and `status === 'closed'` with `TeamClosedError`, so the guard is not lost — it is now stricter for the cron path, which previously bypassed `admit()` entirely.
- **`workflowService.start()`.** Also dropped from the submit path. Still called in `createNew` (for every Team created in-process) and at boot via `DispatcherWorkflows.start() → teams.startWorkflows()`, which lists every non-closed record and materializes it. No Team can reach a leader submission without having had its workflow admission opened.
- **"Every leader path goes through `submitToLeader`".** Holds. `git grep leaderService()` leaves exactly one submit call site (inside `submitToLeader` itself); the others are `prepareCompletion`, `status()`, and the dissolve close. Entry points are `DispatcherService.submitToTeamLeader` (Channel + Team MCP `send`) and `submitScheduled` (cron), both through `submitToLeader`; `createNew`'s prompt path calls `submitInput` directly but sets the record status itself. The one remaining `.activate()` in `src` is `input-source-lifecycle.ts:355`, which is the Dispatcher's own agent, not a TeamLeader — correctly out of scope.
- **The fix itself.** `submitAdmitted` announces the input via `projectInput`, then `admitToRuntime` → `runtimeOwner.ensureStarted()`, with the `catch` projecting `turn.ended { status: 'failed', reason }` before rethrowing. A start failure now lands inside the admitted-input span, which is exactly the gap the stuck card came from. `team-leader-start-failure.test.ts` asserts the ordering and the provider's own text, not just that an error surfaced — good test.
- **Base drift.** Diffed against the PR's real base (`4561f52`), not a stale local `next`. CI is green on all 9 checks.

**Two non-blocking observations**

1. `submitToLeader` flips `starting → running` only on `admission.status === 'submitted'`. A `duplicate` admission (durable ledger key replayed after a crash between ledger persist and record update) leaves the record on `starting`. It self-heals on the next distinct `sourceId`, and arguably a deduplicated repeat proves nothing new about the leader — so this reads as intended rather than a defect. Worth a line in the comment if you agree it is deliberate.
2. The cron path picks up `admit()` for the first time. A fire during dissolve now rejects with `TeamClosedError` instead of proceeding; the scheduler already had to tolerate a throwing `submitScheduled` (the old `ensureRouteReady()` threw on a closed record), so only the timing and error type changed. Reads as a fix, noting it since it is a behavior change the PR description does not call out.

The residual window (codex killed between `thread/start` and its first rollout) is stated in the description and left undefended deliberately — with the card now showing the error, that is the right trade.

PR #376, opened 2026-09-03
Hide Claude Code user envelopes and title COT tool rows from the runtime's own label

YourWildDad:
Two display-line changes to the Feishu COT card, both from operator rulings on 2026-09-03.

## 1. Hide Claude Code `user` envelopes

Skill bodies (the whole SKILL.md) showed up on the card as assistant text. The CLI emits them as `type:"user"` text blocks after a `Skill` call, with no field marking them as injected. Ruling: 「所有的 user 消息都隐藏即可」.

The runtime now maps by envelope role: `assistant` → text + tool calls, `user` → tool results only. Research (SDK 0.3.259 source catalog, official docs audit, two live probes) is frozen in `.agents/research/claude-code-stream-json-protocol.md`.

## 2. Tool rows say what the call was

Rows rendered as `Bash ({"command":"ls -la && echo ...`. Ruling: 「优化一下 COT 的展示效果 … 你需要同时兼顾 claude code 和 codex 这两边 … 在 dreamux 的 activity 的 interface 里，扩展一些字段，最后 feishu-cot-adapter 这里对齐最终上传到飞书的消息格式」.

- **Contract** (`dreamux-types`): `tool.call` gains `summary` (the one line the runtime's own UI labels the call with) and `invocation` (the call in the caller's notation). Both nullable; `TeammateActivity` adds the matching `*_truncated` flags. This fulfils COT requirement rule 15.
- **Claude Code**: derived from the built-in tool input schemas (`sdk-tools.d.ts`) in the new `src/tool-display.ts`, which also absorbs the old `namedToolAction`. Bash → description (else first line of the command) + command line; file tools → path; Grep/Glob/WebSearch/ToolSearch → pattern/query; Agent → description + prompt; Skill → name; WebFetch → URL. MCP tools → `null`.
- **Codex**: derived in `src/tool-display.ts` in the TUI's own wording (exec cell, patch cell, web-search cell). A uniformly parsed command is labelled by the files read / paths listed / `query in path`, else by its first line; a patch by its paths with the prepared diffs as invocation. `webSearch` items, previously dropped (no tool name on the wire), now appear as `web_search` rows.
- **Core**: both facts go through the same path renaming and secret redaction as arguments/results (and, since the ruling below, no bounding at all).
- **Feishu channel**: the row title comes from the summary (verb-prefixed for read/list/search/edit, tool-name-prefixed for rows without an action), the row carries the documented built-in icon for its action, a titled row sends no raw `TOOL_CALL_ARGS`, and the expanded result shows the invocation as a `code` segment before the output. Code segments now spell their body in the documented `code` field (the probe showed the client renders the previous `content` too, so this follows the docs rather than fixing a blank block). Wire shapes follow the enterprise-only COT Message Brief, fetched verbatim this time.

- **Teammate-tool overrides removed** (ruling 「那你先给 dreamux 内置mcp 工具的覆盖都删掉吧。我想想这里怎么做。」): the Channel used to title `spawn` / `send` / `close` / `workflow_run` rows by matching the `mcp__teammate__` / `teammate.` name and re-parsing the arguments against Core's field names — a copy of Core's tool schema inside the Channel, undocumented since #347. Deleted, together with the result knobs only it used (`resultLanguage`, `forceResultCode`); those rows are now unlabelled MCP rows. How Core labels its own tools is left open for the operator's design.
- **Channel-owned tool titles removed** (ruling 「这些全部回退吧」, after the operator noticed the `react` row's built-in `default` icon renders no glyph): the Channel's own `reply` / `react` / `list_chat_bots` rows lose their hand-made titles, icons and fixed Completed/Failed result line. Deleted with them: the `OwnedTool` type and table, the MCP server-name matcher, the `channelId` threaded through the activity sink and the event builders for it, and the two-shape `ToolPresentation`. A row is now presented from what the runtime said about the call and nothing else; the Channel's tools render exactly like a foreign MCP tool.
- **MCP rows** (ruling 「现在 mcp 工具效果比较差，mcp 工具隐藏掉参数吧，icon 选 app-default_outlined」): a call with neither an action nor a runtime label — an MCP tool, today — shows its display name behind the icon-library token `app-default_outlined` and no longer sends `TOOL_CALL_ARGS`; no row sends that event any more. Its output still expands.
- **Compaction as one line** (rulings 「我不要正文，正文太长了，只显示压缩发生了即可。claude code 的网页上只显示了 Compacted session，我只需要这一行字即可。」 and 「没必要给他单独加一个新的 activity 类型，你直接在provider 里，多推一个 assistant message，内容就这一行。」): both runtimes publish an `assistant.message` reading `Compacted session` — Claude Code on its `system`/`compact_boundary` envelope (live-probed on CLI 2.1.259 with `/compact`: the boundary is followed by a synthetic `user` envelope carrying the summary as one string, which stays hidden), Codex on the completion of its `contextCompaction` item (id only on the wire; the summary never leaves codex-core, verified against the schema generated from the installed 0.151.0 binary). No new activity kind.
- **Card text in English** (ruling 「全都翻译成英文吧」): the COT adapter's own words — row verbs (Read/List/Search/Edit), the Completed/Failed status lines, and the truncation marker — are English. Other feishu-channel cards (pairing, ask-user, binding notification, session ops, introduce) still carry their Chinese text; not touched here.
- **Files as pills** (follow-up ruling 「两边自己去按照自己的协议，把能拆分出来的 file item 拆出来，拆不出来的就不管了」): `tool.call` also gains `items`, what the call is about one entry each (named for the list, not for files, per 「那我推荐你给改成 items，不要固定叫 files，限制了其他用法」; today the files a read or edit touches), filled by each provider only from structured members of its own protocol (Claude Code `file_path`/`notebook_path`; Codex `fileChange.changes[].path` and the `path` of a parsed `read` command action), empty otherwise, never parsed out of text. Feishu shows them as the pills of a `list` result segment with the action's icon,, bounded by `TOOL_ITEMS_SOFT_MAX_BYTES` with one `+N` pill for the rest. A read or edit that succeeded expands to its pills and nothing else (ruling after probe #2: 「有了胶囊的，可以忽略底下这个编辑的代码段。只看编辑了哪些文件就行了」, since the client cannot fold a code segment away); a failed one keeps the diff and the output, which is where the failure's reason appears.

What stays raw: MCP tool rows the Channel does not own — no runtime can label them. Recorded in the requirement page together with the Codex-side gaps found on the way (declined → completed, exit code unread, MCP error stringified).

Live probe (2026-09-03, a card built with the transport's own `createFeishuCotClient` and sent to the operator's DM): a titled row shows the `title` alone with its icon and no tool name in front; a `TOOL_CALL_ARGS` delta sent beside a title is displayed nowhere, which is what makes omitting it on titled rows correct and putting the invocation in the result necessary; a row with no `icon` renders without one; `code` and `content` both render; a `list` result renders as chips.

## Verification

- `rush build|test|lint|typecheck:tests --from @excitedjs/dreamux-types`: all green.
- New unit tests: `tool-display.test.ts` in both runtimes, `feishu-cot-tool-rows.test.ts` in the channel, two projection privacy cases in core.
- `.agents/scripts/check.sh`: KB OK. Leak grep clean.
- The running daemon is still 0.23.0-beta.195; the card changes only after upgrade.

- **Review follow-ups** (Devbox approved with seven inline notes; the operator ruled on each): a first item longer than the whole pill budget is truncated into one pill instead of folding into `+N` (「按照单条去截断即可」); `normalizedDetail` loses its dead alias; `ClaudeActivityLine` and the KB say three envelopes (`assistant`, `user`, `system`/`compact_boundary`); tests pin the uniform `listFiles` command on both sides and the `openPage` label. Declined: bounding `items` in Core, propagating `summary_truncated` into titles, and a `uuid` id for the compaction line (not reachable: the sequence is monotonic within a window and a card ends with the window).

- **A JSON output is pretty-printed as JSON; a text output is text.** A result that parses as a JSON object or array is shown in a `json` code segment via `JSON.stringify(value, null, 2)`; anything else is a plain text segment (「文本的输出，就按文本输出。能解析成JSON的再放进代码段」, 2026-09-04). This replaced both #347's inline exception for a one-line output under 120 bytes and the all-code rule that lasted one commit.
- **Core redacts and never truncates; the Channel cuts last, at Feishu's limit.** The `CONVERSATION_*` character caps, the tool-name cut and every `*_truncated` flag on the ChannelCoreEvent display facts are gone (「core那边只做脱敏，不做截断」); the provider seam drops the always-false `truncated` on assistant messages and, per 「顺手删掉 RuntimeCompletion.truncated」 (commit 265a3baa), on completions: both bundled providers only ever set it to false and nothing in Core read it (the `truncated` on `teammate.last` is the activity page's live flag and stays). The Channel parses and prints the whole value first and cuts only against the 4,096-byte per-event content limit with the marker (「截断长度以飞书平台上给出的最长长度为准，现在有点短」); the 512-byte invocation/title and 1,024-byte output soft caps are gone. `RuntimeCompletion.truncated` (persisted turn record, `last` tool output) is untouched pending a ruling.
- **A redacted structured result still parses.** Follow-up to the two bullets above (commit 5e098ee7): secret redaction used to turn `{"token":"abc"}` into `{"token":<redacted>}`, which no longer parsed and so fell silently to a text segment, and a bare `token: xyz` phrase inside a JSON string swallowed the quote and brackets closing it. A quoted secret is now the quoted string `<redacted>` and a bare one stops at a quote or closing bracket. Visible change in plain text: `password: "<redacted>"` keeps its quotes.

https://claude.ai/code/session_01WxU1pdZzxGWNh1gGYiXQ1B
https://claude.ai/code/session_01THnULV9wt1Ew4TNg6bJAz6


packages/dreamux/src/channel/conversation-projection.ts
```diff
@@ -198,6 +199,13 @@ function projectedActivity(
       };
     }
     case 'tool.call': {
+      const summary = activity.summary === null
+        ? null
+        : sanitizeText(activity.summary, cwd, homePathPrefixes, CONVERSATION_TOOL_SUMMARY_MAX);
+      const invocation = activity.invocation === null
+        ? null
+        : sanitizeText(activity.invocation, cwd, homePathPrefixes, CONVERSATION_TOOL_ARGUMENTS_MAX);
+      const items = activity.items.map((item) => redactText(item, cwd, homePathPrefixes));
```
ryanxiang7, line 213: **minor (contract consistency):** `items` is the only `tool.call` field Core redacts but never bounds — `summary` (512), `invocation`/arguments (60k) and result (120k) all have caps, and the contract carries `*_truncated` flags for the texts. A runtime filling `items` with long strings (a `Read` whose `file_path` is a huge value — the `started` activity ships before the tool ever fails — or a Codex patch over hundreds of paths) publishes an unbounded `teammate.activity` payload. `TOOL_ITEMS_SOFT_MAX_BYTES` is a Feishu presentation budget; it protects the card but not the Core event or a second subscriber, and `TeammateActivity` has no `items_truncated` to tell anyone the list was cut. Consider a per-item/count cap here (plus the flag), or an explicit contract note that `items` is best-effort and channel-bounded.

packages/channel/feishu-channel/src/feishu-cot-presentation.ts
```diff
@@ -55,203 +81,85 @@ const ACTION_TOOL_NAMES: Readonly<Record<RuntimeToolAction, string>> = {
...
+function runtimeToolTitle(event: CotToolCallActivity, actionName: string): string | null {
+  if (event.summary === null) return null;
+  const summary = boundedTitleText(event.summary);
+  if (summary === '') return null;
+  return event.tool_action === null
+    ? `${actionName}: ${summary}`
+    : `${ACTION_VERBS[event.tool_action]}${summary}`;
 }
 
 function boundedTitleText(value: string): string {
```
ryanxiang7: **minor:** `boundedTitleText` hardcodes `sourceTruncated: false`, so a summary Core clipped at `CONVERSATION_TOOL_SUMMARY_MAX` renders verbatim with no `… (truncated)` marker — 512 ASCII chars are exactly 512 escaped bytes, so the budget check passes and the title can end mid-path-name. The invocation path right above (line 120) already propagates `event.invocation_truncated`; pass `event.summary_truncated` here the same way.

packages/agent-runtime/claude-code/src/runtime-submissions.ts
```diff
@@ -212,32 +215,67 @@ function failUnattributedResult(
...
+ * 单独加一个新的 activity 类型，你直接在provider 里，多推一个 assistant message，
+ * 内容就这一行。」
+ */
+const COMPACTED_SESSION_MESSAGE = 'Compacted session';
+
+function compactedActivity(active: ActiveTurn): RuntimeActivity {
+  return {
+    kind: 'assistant.message',
+    occurredAt: Date.now(),
+    id: `stream-${active.activitySequence++}:compacted`,
```
ryanxiang7, line 270: **minor (id collision across windows):** `activitySequence` is per-`ActiveTurn` and starts at 0 (`runtime.ts:284`), and assistant envelopes carry a native `message.id`, so in practice the sequence is still 0 when a compaction arrives — every execution window's compaction publishes `event_id = "stream-0:compacted"`. A COT card explicitly outlives a turn, and the channel derives the wire `messageId` straight from the event id (`feishu-cot-activity.ts:121` → `opaqueDisplayId('message', …)`), so two compactions on one card send the same `messageId`. The research doc notes `uuid` sits on every `SDKMessage` member, `compact_boundary` included — `stringValue(line.raw['uuid']) ?? \`stream-${active.activitySequence++}\`` + `:compacted` would make it unique.

packages/agent-runtime/codex/src/tool-display.ts
```diff
@@ -0,0 +1,169 @@
...
+  for (const entry of value) {
+    const record = recordValue(entry);
+    if (record === null) continue;
+    switch (record['type']) {
+      case 'read': {
+        const name = stringField(record, 'name');
+        actions.push({ action: 'read', detail: name, file: stringField(record, 'path') ?? name });
+        break;
+      }
+      case 'listFiles':
```
ryanxiang7, line 78: **minor (test gap):** the uniform-`listFiles` path — the subtlest branch of this module and a common case (`ls`) — has no covering test in either `tool-display.test.ts` or `feishu-cot-tool-rows.test.ts`; a regression collapsing `list_files` to `run` would go uncaught. `openPage` (line 135) is likewise untested. One test per side would pin it: codex asserts `{ action: 'list_files', summary: '<path>' }`, the channel asserts the `List <path>` title and the `search` icon.

packages/channel/feishu-channel/src/feishu-cot-presentation.ts
```diff
@@ -55,203 +81,85 @@ const ACTION_TOOL_NAMES: Readonly<Record<RuntimeToolAction, string>> = {
...
+ * for its diff: the pills that fit, then one `+N` pill for the rest.
+ */
+function itemList(event: CotToolCallActivity): CotItemList | null {
+  if (event.items.length === 0) return null;
+  const icon = event.tool_action === null ? undefined : ACTION_ICONS[event.tool_action];
+  const items: CotListItem[] = [];
+  let bytes = 0;
+  for (const [index, item] of event.items.entries()) {
+    bytes += escapedBytes(item);
+    if (bytes > TOOL_ITEMS_SOFT_MAX_BYTES) {
```
ryanxiang7, line 141: **nit:** the item's bytes are added *before* the budget test, so a first item that alone exceeds `TOOL_ITEMS_SOFT_MAX_BYTES` is never pushed — the result is `{ items: [], more: '+1' }`, and a succeeded read/edit (items-only branch in `feishu-cot-events.ts`) then expands to a lone `+1` pill. A path over 512 escaped bytes is legal (`PATH_MAX` is 4096). Consider truncating the over-budget item into one pill (`truncateEscaped`) instead of folding it entirely into `more`.

packages/channel/feishu-channel/src/feishu-cot-presentation.ts
```diff
@@ -55,203 +81,85 @@ const ACTION_TOOL_NAMES: Readonly<Record<RuntimeToolAction, string>> = {
...
 }
 
-function workflowMetaName(script: string): string | null {
-  const match = /\bexport\s+const\s+meta\s*=\s*\{[^}]*?\bname\s*:\s*(['"])([^'"\\\r\n]+)\1/u
-    .exec(script);
-  const name = match?.[2]?.trim();
-  return name === undefined || name === '' ? null : name;
-}
-
 export function normalizedDetail(
```
ryanxiang7: **nit (dead code):** inside this function, `const normalized = value` is a pure alias and `if (normalized === '') return null` is unreachable — the guard at the top already returned for `''`. Leftover from when this function actually normalized (the comment two lines up gives it away). Delete the alias and the check, use `value` directly.

packages/agent-runtime/claude-code/src/types.ts
```diff
@@ -147,13 +167,24 @@ export interface ClaudeCodeSessionSpec {
   onProtocolEvent?: (event: ClaudeProtocolEvent) => void;
 }
 
+/**
+ * The two envelopes the display line reads: what claude said and did
```
ryanxiang7: **nit (docs):** the docstring says "two envelopes" but `ClaudeActivityLine` now covers three — `compact_boundary` was added in the `Extract` below. The KB repeats it (`.agents/domains/provider-runtime.md`: "reads two stdout envelopes") and lists `system` notices among what stays inside the RPC, which is no longer true of `compact_boundary` (`rpc.ts:403-405` forwards it). Say three in both places and move `compact_boundary` out of the stays-inside list.

ryanxiang7 (approved):
## Review

Three-model pass (codex / seed / claude teammates) plus an independent lead pass. The architecture direction is sound: the Channel no longer copies Core's tool schema, display facts flow through the neutral contract, the deletions are clean (no dangling `OwnedTool` / teammate-tool / `channelId` references), every new field is redacted on the Core projection, no row sends raw `TOOL_CALL_ARGS` any more, tests are additive with no weakened assertions, and the change files follow the 0.x rules (`dreamux-types` correctly records the compile-time break as `minor` + `BREAKING:` with no `Rebuild:`).

Findings: 4 minor + 3 nit, all non-blocking — approving. Take them here or in a follow-up, author's call.

**minor**
- `items` is the only `tool.call` field Core redacts but never bounds — contract consistency, not a live leak
- the card title drops Core's `summary_truncated` marker while the invocation right beside it keeps its own
- the compaction activity id collides across execution windows (`stream-0:compacted`), and a COT card outlives a turn
- the uniform-`list_files` path — the subtlest branch of the codex module — has no covering test on either side

**nit**
- a first item over the pill budget yields zero pills and a lone `+1`
- dead alias + unreachable check left in `normalizedDetail`
- "two envelopes" doc is now three (`compact_boundary`); the KB repeats it

PR #378, opened 2026-09-04
Keep the spacing of a plain-text tool output on a COT card

YourWildDad:
## Why

A Bash row's plain-text output lost its indentation on the COT card: the Feishu client collapses a run of ordinary spaces in a `text` segment to one and drops the run that begins a line (operator, 2026-09-04: 「这里头的空格都没了」).

A probe card sent the same day settled the client facts:

- a raw run of spaces collapses;
- U+00A0, U+2002 and the `&nbsp;` entity all keep the indentation;
- a `<pre>` wrapper is shown as literal text with its spaces still collapsed;
- an entity inside a row `title` is decoded too — tags are escaped, entities are decoded.

## What

- `preserveSpacing` in `feishu-cot-presentation.ts`: each space that begins a line, or sits in a run of two or more, becomes a no-break space (U+00A0). A single space between words stays ordinary so a long line still wraps.
- Applied on the `text` branch of `toolResultOutput`, before the event budget is measured, so the cut counts the converted bytes. JSON code segments are untouched.
- The character rather than the entity: two bytes of the 4,096-byte event budget instead of six; the probe showed both render the same.
- Tests: indentation and column alignment kept, single spaces untouched, JSON output unaffected, the cut measured after conversion.
- Product README, `channel.md` (probe facts), and a `patch` change file for `@excitedjs/feishu-channel`.

## Gates

`rush build`, `lint`, `test`, `typecheck:tests` from `@excitedjs/feishu-channel`: green. `.agents/scripts/check.sh`: KB OK.

https://claude.ai/code/session_01THnULV9wt1Ew4TNg6bJAz6

### DIFF
--- a/.agents/domains/channel.md
+++ b/.agents/domains/channel.md
@@ -781,7 +781,18 @@ truncated into one pill rather than folded into that count, per 「按照单条�
 than by how long it is: a value that parses as a JSON object or array is
 pretty-printed (`JSON.stringify(value, null, 2)`) in a `json` code segment,
 anything else is a plain `text` segment (「文本的输出，就按文本输出。能解析成JSON
-的再放进代码段」, 2026-09-04). That rule replaced, within one day, both #347's
+的再放进代码段」, 2026-09-04). The text segment travels with each space that
+begins a line, or sits in a run of two or more, turned into a no-break space
+(U+00A0, `preserveSpacing` in `feishu-cot-presentation.ts`): the client
+collapses a run of ordinary spaces and drops a leading one, so a Bash
+output's indentation and column alignment were lost (「这里头的空格都没了」,
+2026-09-04). A probe card the same day settled the client facts: a raw run
+collapses; U+00A0, U+2002 and the `&nbsp;` entity all keep it; a `<pre>`
+wrapper is shown as literal text with its spaces still collapsed; and an
+entity inside a row `title` is decoded too — tags are escaped, entities are
+decoded. The character was chosen over the entity for costing two bytes of
+the event budget rather than six, and a single inner space stays ordinary so
+a long line still wraps. A code segment keeps its own spaces. That rule replaced, within one day, both #347's
 inline exception for a one-line output under 120 bytes and the all-code rule
 that lasted one commit (「全都给他们包到代码块里面」), once MCP rows made the
 mix of boxed and unboxed outputs visible. A code segment spells its body in
--- a/packages/channel/feishu-channel/src/feishu-cot-presentation.ts
+++ b/packages/channel/feishu-channel/src/feishu-cot-presentation.ts
@@ -165,6 +165,23 @@ export function nonEmpty(value: string | null): string | null {
   return value === null || value === '' ? null : value;
 }
 
+const NO_BREAK_SPACE = '\u00a0';
+
+/**
+ * A `text` segment as the Feishu client keeps its spacing. The client
+ * collapses a run of ordinary spaces to one and drops the run that begins a
+ * line, so a text output lost its indentation and its column alignment; a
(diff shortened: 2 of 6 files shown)


packages/channel/feishu-channel/src/feishu-cot-presentation.ts
```diff
@@ -165,6 +165,23 @@ export function nonEmpty(value: string | null): string | null {
...
+ * collapses a run of ordinary spaces to one and drops the run that begins a
+ * line, so a text output lost its indentation and its column alignment; a
+ * no-break space stays where it was (probe card, 2026-09-04: raw spaces
+ * collapsed, U+00A0 and the `&nbsp;` entity both kept the indentation, a
+ * `<pre>` wrapper was shown as literal text). Each space that begins a line,
+ * or sits in a run of two or more, becomes U+00A0; a single space between
+ * words stays a space, so a long line still wraps there. The character, not
+ * the entity: two bytes of the event budget instead of six.
+ */
+export function preserveSpacing(text: string): string {
```
ryanxiang7, line 181: **nit (follow-up probe, not a defect in this PR):** the pattern matches U+0020 only, so a tab-indented output (`cat Makefile`, Go sources, `git diff` hunks) reaches the card with its tabs untouched — if the client folds a tab the way it folds a space run, that output still shows flush-left, the same symptom this PR fixes for spaces. The 2026-09-04 probe only exercised spaces / U+2002 / `&nbsp;` / `<pre>`, so the client's tab behavior is unverified. Worth one more probe card before deciding whether to extend the conversion — and recording the result in `channel.md` either way, so the gap is closed knowledge rather than an open question.

.agents/domains/channel.md
```diff
@@ -781,7 +781,18 @@ truncated into one pill rather than folded into that count, per 「按照单条
...
+(U+00A0, `preserveSpacing` in `feishu-cot-presentation.ts`): the client
+collapses a run of ordinary spaces and drops a leading one, so a Bash
+output's indentation and column alignment were lost (「这里头的空格都没了」,
+2026-09-04). A probe card the same day settled the client facts: a raw run
+collapses; U+00A0, U+2002 and the `&nbsp;` entity all keep it; a `<pre>`
+wrapper is shown as literal text with its spaces still collapsed; and an
+entity inside a row `title` is decoded too — tags are escaped, entities are
+decoded. The character was chosen over the entity for costing two bytes of
+the event budget rather than six, and a single inner space stays ordinary so
+a long line still wraps. A code segment keeps its own spaces. That rule replaced, within one day, both #347's
```
ryanxiang7, line 795: **nit:** the inserted spacing paragraph splits the pre-existing `That rule replaced, within one day, both #347's inline exception…` sentence from its antecedent (the parses-as-JSON-vs-text rule, now eleven lines up), so `That rule` reads as if it referred to `A code segment keeps its own spaces`; the joined line is also 109 columns against this file's ~76. Moving the spacing paragraph to after the `That rule replaced…` sentence and re-wrapping would keep both rules' follow-ons attached.

ryanxiang7 (approved):
## Review

Three-model pass (codex / seed / claude teammates) plus an independent lead pass — all four approve. The change is minimal and correct: the regex does what its doc comment says, the U+00A0 conversion happens before the event budget is measured (the cut-after-conversion test pins this, and `checkedEvent` hard-caps at 4,096 bytes), the redaction ordering is untouched (Core redacts first; the substitution cannot corrupt a `<redacted>` marker or a renamed path), JSON code segments are unaffected, and the modified assertion is a correct adaptation, not a weakening. Change file is `patch` on `@excitedjs/feishu-channel` (past 1.0, real semver) — right type.

Two nits below, both non-blocking — take or defer.

PR #379, opened 2026-09-05
Add Feishu slash commands with a Core turn-interrupt capability

YourWildDad:
## What this adds

Three Feishu slash commands the Channel executes itself — `/stop`, `/teams`,
`/dissolve` — plus the Core capability `/stop` needed.

A human message whose leading text (after mentions) is a known `/command` token
is carried out by the Channel and answered as a receipt: no agent runtime sees
it, and no agent is asked to render the reply. Only the leading token counts —
trailing words after it are ignored, and the same token mid-message is ordinary
text. In a group the bot must be @-mentioned. Nothing gates a command beyond the
ordinary authorization to deliver a message there; there is no separate command
permission. When there is no object to act on, the reply is one line saying why.

The commands live in one table (`feishu-slash-commands.ts`) behind a single
dispatch site, so a fourth command is a table entry rather than a new branch in
the inbound path.

### `/stop` — a new `AgentRuntime` capability

`AgentRuntime` gains a fourth method next to `start`, `submit`, and `stop`:
`interrupt`, reported through the new `team.interrupt` Command. It ends the
current turn of the agent this conversation talks to directly — the bound Team's
TeamLeader, or the Dispatcher Agent in a DM — and never its TeamMates. A runtime
with no turn in flight answers `idle` and is never started in order to be
interrupted.

- Claude Code maps it to a `control_request` interrupt and settles the
  interrupted turn as its own outcome, instead of as a hard failure or a hang
  until the idle reaper.
- Codex maps it to `turn/interrupt`, present since the 0.137 minimum this repo
  already requires. No version floor moves.

### `/teams`

Renders the running Teams as a card built here in TypeScript rather than by an
agent, so `team.list` now projects `leader_agent_runtime` and the transport can
resolve a chat's real name. Teams are grouped by repository; repository and
runtime colors come from a stable hash, so a given repo or runtime always reads
the same.

### `/dissolve`

Dissolves the Team bound to this conversation with a generated note, never
`force`. Making its receipt honest required a Core fix: `TeamService.dissolve`
now assesses the managed worktree **before** it publishes the dissolve task, so a
refusal (dirty or unmerged) reaches the caller instead of being logged in the
background behind an `accepted` receipt. The same reordering closes a race where
two concurrent non-forced dissolves both dismantled the same Team.

### Text

All Channel-authored text for these commands, and the existing `/introduce`
acknowledgement, is English. The pairing-approval card keeps its existing
Simplified Chinese default copy.

## Validation

- `rush build`, `rush lint`, `rush test`, `rush typecheck:tests` — green.
- `rush change --verify` and `git diff --check` — clean. Six `minor` change
  files; none needs a rebuild.
- `.agents/scripts/check.sh` — KB OK.
- New `feishu-slash-commands.test.ts` covers recognition (leading token only,
  trailing text ignored, mid-message token is text, casing, the group
  @-mention requirement, bot senders rejected) and each command's receipt shape.
  Runtime `interrupt` and the dissolve ordering and concurrency contract are
  covered in the runtime and dissolve suites.
- An independent `xhigh` review ran over the working diff and found two real
  regressions — an interrupted Claude Code turn settling as a failure, and two
  concurrent non-forced dissolves both dismantling the Team. Both are fixed here
  and re-reviewed at source.

## Review round 2

Rebased onto `next` at `2575e056`. Nine inline findings, each answered on its
comment. The one that mattered:

**An interrupt never actually settled as an interrupt on a real CLI.** Review
asked whether a late artifact could reap the resident session. Probing that
against `claude` 2.1.263 — twice, interrupting a stream and interrupting a 60s
tool call — showed the ordering it feared never happens, and showed something
worse: the artifact carries the interrupted command's own `user_message_uuid`,
while the recognition condition required that field to be absent. Every real
interrupt fell through to the ordinary result path and reported the turn
`failed`, the outcome the capability exists to prevent, and the unit test passed
on a fixture fabricated from the same wrong belief. The condition is now our own
accepted request plus the subtype; the fixture is the captured envelope; and the
reviewer's hazard is answered by deleting the settlement on
`command_lifecycle: cancelled`, so the artifact is the only place an interrupt
settles and the chain has no entry point.

Per operator ruling — 「你只要给[Request interrupted by user] 输出到 cot 上，最后卡片
是什么状态其实没那么重要。一切以对当前架构侵入性最小，为最主要的决策路径」 — an
interrupted turn now shows `[Request interrupted by user]` on its card, pushed as
an assistant message the way `Compacted session` is. The probe table is in the
task record's `verification.md`.

Also: the supervisor's `interruptTurn` gains the liveness guard its siblings
have; `unavailableTeamReason` owns a mapping that was written twice;
`TeamDissolveBlockedError` names a next step now that it reaches a caller; two
stale comments and one wrong doc table row are corrected. `/teams` staying
unscoped is an operator ruling, recorded verbatim in `requirement.md`.

## Round 3 — operator rulings

Two rulings answered the questions round 2 left with the operator.

**「codex 也加一个打断text ，对齐claude code」** — a codex interrupt now writes the
same line. codex reports an accepted interrupt only as a terminal status, so the
provider had never read it. Measured against a real `codex-cli 0.153.4`: an
interrupted turn's `turn/completed` carries `turn.status: "interrupted"`,
`error: null`, no items; a turn codex answers carries `"completed"`. So
`TurnCompletedNotification.turn` gains `status`, `CollectedTurn` carries it, and
`observeTerminal` pushes `[Request interrupted by user]` ahead of the end it
already reports. The fixtures now carry `status` too — a fixture that does not
model the wire is exactly what let round 1 ship a Claude Code recognition
condition no real interrupt could match.

Reading the status rather than remembering that we asked is deliberate:
`interrupt()` picks a turn with no terminal yet and only then sends the request,
so codex can finish that turn in the window between the two.

Unchanged, and deliberately: an interrupted codex turn still ends `completed` and
settles its submissions with codex's completion. The ruling names the text, and
the operator had already ruled 「最后卡片是什么状态其实没那么重要」. The remaining
divergence is recorded in `provider-runtime.md`, not closed silently.

**「可以先不动。」** — review finding #5, no idempotency key on the command path,
is deferred by the operator. Recorded as a deferral in `requirement.md`, not as a
claim that redelivery cannot happen. Both redelivery windows of all three
commands were traced and answered on that thread.

Still standing and labelled: the inherited `#342` ignore-branch, whose premise
the probes did not reproduce. It is marked unreproduced in both the source and
the fixture rather than deleted on one negative observation.

## Round 4 — rebased onto the resident-session model

`next` moved twice under this branch while it waited for review, and both
changes landed in files it rewrites.

**#384 replaced Claude Code's per-turn execution window with a resident request
table.** `PendingTurn`, `submitTurn`, `steerTurn`, and `failPending` are gone;
`submit()` returns a `RuntimeAdmission`, and one session holds many outstanding
requests at once. The interrupt was ported rather than replayed:
`interruptTurn` is now `interrupt`, and the mark recording that we asked for one
moved from the turn record to the session, because there is no longer a single
turn to hang it on. `interrupt()` answers `false` when nothing is outstanding,
and the mark is spent on the first `result` either way — an accepted interrupt
that found nothing to stop must not let a later genuine `error_during_execution`
read as somebody's interrupt. The control channel is its own object now
(`control-rpc.ts`): it correlates one control request with its
`control_response` and knows nothing about turns.

**#386 widened route removal from two reasons to three** —
`team_closed | route_ended | stale_route`, with `announceRoutesRemoved` deciding
what the conversation hears. That behavior wins over this branch's two-reason
form, and it now lives inside this branch's extracted
`FeishuRouteReconciliation` rather than beside it. The visible consequence: a
command whose Team rejects with `TEAM_CLOSED` shows the route-ended card, not
the dissolved card — a pending dissolve can still be refused, so only the final
`team.state` proves the Team closed.

The branch was squashed to one commit before the rebase. `runtime-submissions.ts`
is deleted on `next` and `rpc.ts` was rewritten there, so replaying nine commits
meant resolving the same conflict against five intermediate states that no
longer exist; `next` squash-merges on the way in regardless. The pre-rebase
history is kept at `backup/pre-rebase-94e31828`.

Three follow-ups sit on top of the ported commit:

- Operator decisions that had been quoted in code comments moved to the task
  record. The rationale stays in the comment, in English, without attribution.
- Session teardown answering an outstanding interrupt now has a test. It is the
  third settlement path, and the one that keeps `/stop` from waiting on a
  session that has already gone away.
- `channel.md` still described route removal in the two-reason form. Corrected.

Re-verified on the new base: `rush build`, `rush lint`, `rush test`,
`rush typecheck:tests` and `.agents/scripts/check.sh` are all green.

## Limitations

- Only the Feishu channel has a slash-command surface. `interrupt` is a neutral
  `AgentRuntime` capability, so every runtime provider must now implement it.
- `/dissolve` never forces. A Team whose worktree is dirty or unmerged is refused
  with the reason; clearing it stays a manual decision.

Task record: `.agents/tasks/channel/add-feishu-slash-commands/`. No
solution-review Issue exists — the operator waived solution consultation and the
public Issue for this task.


packages/agent-runtime/claude-code/src/rpc.ts
```diff
@@ -466,6 +486,10 @@ export class ClaudeCodeStreamRpc {
           line.outcome.userMessageUuid === null &&
           line.outcome.text === null
         ) {
+          if (pending.interruptRequested) {
```
ryanxiang7: **Major — a late interrupt artifact reaps the resident session.**

When the lifecycle terminal arrives first (`command_lifecycle: cancelled` for the interrupted command), `settleIfReady` reaches this branch with `interruptRequested` set and calls `settleInterrupted()`, which clears `this.pending`. The CLI then still emits the interrupt artifact (the `error_during_execution` result with no uuid/text — the file's own `interruptArtifactLine()` docstring says it is "what an interrupt leaves behind", and `provider-runtime.md` says the marked turn "resolves as interrupted from whichever terminal arrives first", i.e. both arrive). That late artifact lands in the `case 'result'` `pending === null` branch at line 465, which logs an error and calls `this.options.reapOnTimeout()` → `LiveClaudeCodeSession.stop()` — killing the resident child this capability exists to keep alive.

The lifecycle-supported test below never sends the artifact, so this ordering is untested; the no-lifecycle test covers only the artifact-first ordering, which is safe.

Suggested fix: recognize the artifact shape (`subtype === 'error_during_execution' && userMessageUuid === null && text === null`) in the `pending === null` branch and ignore it silently (debug log at most), or have `settleInterrupted` leave a short-lived absorption window for one late artifact.

packages/channel/feishu-channel/src/feishu-slash-commands.ts
```diff
@@ -0,0 +1,129 @@
...
+      : {});
+    const result = raw as { status: 'interrupted' | 'idle' };
+    return {
+      kind: 'text',
+      text: result.status === 'interrupted'
+        ? 'Current turn interrupted.'
+        : 'No turn is running.',
+    };
+  }),
+  teams: defineCommand('teams', 'Command /teams failed', async (context) => {
```
ryanxiang7: **Design question for the operator — `/teams` has no conversation scope.**

This command ignores `context.plan` entirely and renders *every* running Team on the dispatcher — team name, intent text, repo basename, leader runtime, and each Team's bound chats with their current names and clickable links — to any chat that passes the ordinary inbound gate, including a group bound to a different Team. For comparison, the MCP surface does not expose `team.list` to TeamLeaders at all (`mcp-delegate.ts` gives them only `dissolve`), so this is a new enumeration capability, not a relocation of an existing one.

If the intended audience is flat (everyone who can reach the bot may see all Teams), this is fine as designed and the PR description is transparent about it. If chats should be scoped, the card should filter to Teams bound to this conversation (or `/teams` should be DM-only). Flagging because the PR's "no separate command permission" ruling answers *who may invoke*, not *what data comes back*.

packages/agent-runtime/claude-code/src/supervisor.ts
```diff
@@ -159,6 +159,13 @@ class LiveClaudeCodeSession implements ClaudeCodeSession {
     return this.rpc.steerTurn(prompt, options, commandUuid);
   }
 
+  interruptTurn(reason: string): Promise<boolean> {
```
ryanxiang7: **Minor — missing the liveness guard the sibling write paths have.**

`submitTurn`/`steerTurn` on this class reject when `this.child === null || this.exited`, and `ClaudeCodeStreamRpc.submitTurn`/`steerTurn` reject on `!this.stdin.writable`. `interruptTurn` only checks `stopRequested || stopped || rpc === null`. If the child has died but its `exit` event has not yet run `failPending`, this writes to a broken pipe; the write callback rejects with `EPIPE`, and the user sees `Command /stop failed: write EPIPE` where the semantic answer is `No turn is running.` (the turn is gone with the child). The window is narrow, but this is the only new write path that didn't align with the existing guards.

packages/agent-runtime/codex/src/turn-manager.ts
```diff
@@ -81,6 +83,19 @@ export class TurnManager {
     );
   }
 
+  interrupt(): Promise<AgentRuntimeInterruptOutcome> {
```
ryanxiang7, line 86: **Minor — how an interrupted codex turn settles is undefined and untested.**

`interrupt()` returns `{status:'interrupted'}` on the RPC ack, but the turn's own terminal still flows through `observeTerminal`: an Error terminal maps to `endNativeTurn('failed', ...)` and settles the bound submissions as failed. Both new tests emit a *clean* `turn/completed` after the interrupt, so the error-terminal path is uncovered. If codex reports an interrupted turn as an error terminal (plausible — an aborted turn is not a clean completion), the COT card ends as **failed** with codex's abort text and the submission settles `{kind:'failed'}`, diverging from the Claude path, which settles as `interrupted` / `{kind:'stopped'}`.

Worth either pinning codex's actual terminal shape after `turn/interrupt` with a test, or mapping the aborted terminal to the same interrupted/stopped settlement the Claude runtime uses, so the neutral outcome means the same thing across providers.

packages/channel/feishu-channel/src/feishu-session-inbound.ts
```diff
@@ -362,6 +375,31 @@ async function deliverAcceptedMessage(
       () => h.targetRouter.projectInbound(acceptedEvent, work.signal),
     );
     work.assertSessionActive();
+    if (command !== null) {
+      const reply = await h.delivery.command({
```
ryanxiang7, line 379: **Minor — the command path has no idempotency key.**

Ordinary delivery dedupes redelivered events through Core on `sourceId: event.messageId`; a recognized command bypasses submission entirely (by design), so nothing dedupes it. If the `im.message.receive_v1` event is redelivered (process crash after the command ran but before the WebSocket ack, then reconnect), a `/stop` replays: the first delivery interrupted turn A, and if another user started turn B before the redelivery lands, the stale command interrupts B. `/dissolve` replays are benign (the `dissolveTask` fence joins them) and `/teams` just duplicates a card, so this is specifically about `/stop`.

A small channel-side seen-set of command message ids (bounded, TTL'd) would close it; the alternative is documenting that command delivery is at-most-once-intended.

common/changes/@excitedjs/feishu-channel/dreamux-team-dreamux-claude-ebd3zw_2026-09-04-23-49.json
```diff
@@ -0,0 +1,11 @@
+{
```
ryanxiang7, line 1: **Minor — change type should be `major`.**

`@excitedjs/feishu-channel` is at 5.0.0, so per the repo's semver rules it uses real majors. This PR adds the **required** `command()` member to `FeishuInboundDelivery` (re-exported from `src/index.ts:31`), which is source-breaking for any external implementer of that interface. A new required interface member is a breaking change, so this file should be type `major` (the other five change files are correctly typed: the 0.x packages use `minor`, with the `BREAKING:` prefix on `dreamux-types`).

.agents/tasks/channel/add-feishu-slash-commands/technical-design/final.md
```diff
@@ -0,0 +1,254 @@
...
+  collaboration-space topic must not create a Team. A `provision` plan means
+  "no bound Team" for `/stop` and `/dissolve`.
+- An unmentioned command in a group is not a command. It falls through to
+  ordinary delivery unchanged.
+
+### The three commands
+
+| Command | Object | Action |
+| --- | --- | --- |
+| `/stop` | the agent this conversation talks to | `team.interrupt` with the bound Team name, or without one when the plan is `dispatcher` or `provision` |
```
ryanxiang7: **Minor — this table row contradicts the code and the doc's own prose.**

The row says `/stop` omits `team_name` when the plan is `dispatcher` **or `provision`**. The code (`feishu-slash-commands.ts`) answers `This conversation has no bound Team.` for a `provision` plan and never invokes Core; only `dispatcher` omits the name. The bullet list 10 lines above this table says the same thing ("A `provision` plan means 'no bound Team' for `/stop` and `/dissolve`"), and `.agents/domains/channel.md` matches the code. A maintainer implementing from this authoritative design table would interrupt the Dispatcher Agent's unrelated turn from a fresh collaboration-space topic.

packages/dreamux/src/service/team-service/index.ts
```diff
@@ -481,7 +487,7 @@ export class TeamService {
       await this.closing.dissolve(input);
```
ryanxiang7, line 483: **Nit — stale doc comment.** "The fence goes up before the first await" describes the old shape: the fence now rises in `dissolve()` (`this.dissolveTask = task` at line 468), and `runDissolve` is deferred behind `setImmediate` precisely so the receipt is observable before any of this runs. The comment at line 461 already explains the new ordering; this one should say the fence was raised by the caller before this method was queued.

packages/channel/feishu-channel/src/feishu-channel.ts
```diff
@@ -407,6 +347,42 @@ export class FeishuChannelSession {
...
+          return await this.invoke(command, payload);
+        } catch (error) {
+          const code = commandErrorCode(error);
+          if (
+            plan.kind === 'bound' &&
+            (code === 'TEAM_NOT_FOUND' || code === 'TEAM_CLOSED')
+          ) {
+            await this.routeReconciliation.forgetTeamRoutes(
+              plan.teamName,
+              code === 'TEAM_CLOSED' ? 'team_closed' : 'stale_route',
```
ryanxiang7: **Nit — the unavailable-team reason mapping is now written twice.** `code === 'TEAM_CLOSED' ? 'team_closed' : 'stale_route'` exists here and again in `deliver()` (line ~427, with the same intent behind a slightly different predicate). Since `FeishuRouteReconciliation` is now the single capability that removes these rows, it could own this derivation too (e.g. a static `reasonForUnavailableCode`) so the announce-vs-stay-silent rule is stated once.

ryanxiang7 (changes requested):
## Review: Feishu slash commands + Core turn-interrupt

Reviewed with three teammate lenses (architecture, simplicity, correctness) plus an independent lead pass over the full 67-file diff. The architecture is clean — `AgentRuntime.interrupt` is a proper neutral seam, the command table/dispatch is exactly the right shape, `FeishuRouteReconciliation` is a genuine capability extraction, and the dissolve reordering (assess → publish fence, with a second `dissolveTask` check after the await) correctly closes the concurrent-dissolve race. The locked contract tests (`removed-surfaces`, `core-command-registry`, `package-boundary-guards`) are updated deliberately, not weakened.

One confirmed major regression and one design question block approval; the rest are minors/nits.

### Major

1. **A successful `/stop` kills the resident Claude Code session it promised to preserve** — inline at `rpc.ts:489`. In the lifecycle-supported ordering (`command_lifecycle: cancelled` arrives before the interrupt artifact), `settleInterrupted()` clears `this.pending`; the artifact then lands in the `pending === null` branch, which logs an error and calls `reapOnTimeout()` → `LiveClaudeCodeSession.stop()`. This is the exact outcome the capability exists to avoid, and the PR's own docs (`provider-runtime.md`: "a marked turn resolves as interrupted from whichever terminal arrives first") acknowledge both terminals arrive — the late one has no handler. The lifecycle-supported test never sends the artifact, so this is untested.
2. **`/teams` enumerates every running Team to any chat** — inline at `feishu-slash-commands.ts:71`. This needs an operator ruling, not necessarily a code change: any human in any chat with the bot (including a group bound to Team A) gets team names, intent text, repo basenames, runtime names, and the current names + clickable links of every Team's bound chats. The MCP surface does not even expose `team.list` to TeamLeaders, so this is a new enumeration capability. If the dispatcher's audience is intended to be flat, this is fine as designed; if chats are meant to be scoped, the card should filter by binding.

### Also worth fixing (details inline)

- `supervisor.ts:162` — `interruptTurn` lacks the child-exited/writable guard its sibling write paths have; a dead-but-not-yet-reaped child surfaces `write EPIPE` as `Command /stop failed` instead of `idle`.
- `turn-manager.ts:86` — how a codex turn settles after `turn/interrupt` is undefined and untested: both tests emit a *clean* `turn/completed`. If codex reports the aborted turn as an error terminal, the COT card ends as **failed** and the submission settles `{kind:'failed'}`, diverging from the Claude path's `interrupted`/`{kind:'stopped'}`.
- `feishu-session-inbound.ts:379` — the command path carries no idempotency key (ordinary delivery dedupes on `messageId` through Core). A redelivered `/stop` can interrupt a *newer* turn that started after the original was processed.
- `common/changes/@excitedjs/feishu-channel/...json` — `feishu-channel` is at 5.0.0 (past 1.0, real semver). Adding the **required** `command()` member to the exported `FeishuInboundDelivery` interface is source-breaking for implementers; this change file should be `major`.
- `.agents/tasks/.../technical-design/final.md:197` — the design table says `/stop` omits `team_name` for `dispatcher` **or `provision`** plans; the code (and `channel.md`, and the doc's own prose 10 lines up) treats `provision` as "no bound Team" and never calls Core.
- `team-service/index.ts:487` — the `runDissolve` doc comment is stale: "The fence goes up before the first await" — the fence now goes up in `dissolve()` and `runDissolve` is deferred behind `setImmediate`.
- `feishu-channel.ts:372` — the `TEAM_CLOSED → team_closed / else stale_route` mapping is now written twice (command and deliver paths); `FeishuRouteReconciliation` could own it as one derivation.

### Non-inline note

`team-collection/errors.ts` — `TeamDissolveBlockedError`'s class comment still says a blocked dissolve "never reach[es] a caller" and "neither states a next step, because neither is answering anybody." This PR makes it a caller-facing refusal (the whole point of the receipt-honesty fix), so the comment is stale and the Feishu reply (`Team dissolve refused: ... is dirty`) states no next step. Per the repo's `StatedFailure` discipline, a business failure answered to a caller should say what to do (commit/stash the work, or pass `force`). `errors.ts` is outside this diff so it could not be anchored inline.

### Verified clean (lead pass)

- `pendingInterrupt` cannot leak: `this.pending` is only cleared in `settlePending`, which always calls `control.settleTurn`; concurrent interrupt callers dedupe onto one promise with correct outcome mapping.
- Interrupt never starts a dormant runtime at any layer (rpc → runtime → runtime-owner → team-service), locked by the new `runtimeCreates() === 0` test.
- Dissolve concurrency: two concurrent non-forced dissolves publish exactly one dismantle (test-pinned); a blocked assessment throws before the fence rises; the post-stop final assessment remains the sole authority for destructive reclaim.
- Card rendering escapes every interpolated value and degrades name lookups to the chat id; no repo paths leak (basename only).
- Change files for the 0.x packages are correctly typed `minor` with the `BREAKING:` prefix on `dreamux-types`; no persisted state/config shape changed, so no `Rebuild:` is needed.

packages/agent-runtime/claude-code/src/rpc.ts
YourWildDad: Rejected as written, and the code it pointed at is gone — but the question found a bigger bug, so thank you for it.

Probed against a real `claude` 2.1.263, twice (interrupt during streaming, and interrupt during a 60s tool call). Both runs, same shape and same order: `[Request interrupted by user]` on a `user` envelope → the artifact → `command_lifecycle: cancelled`. The artifact never arrives late.

The artifact itself was the real defect. It is `subtype: "error_during_execution"`, `is_error: true`, no `result` text, and **`user_message_uuid` = the interrupted command**. The recognition condition required that uuid to be absent, so it never matched: every real interrupt fell through to the ordinary result path and reported the turn `failed`, the exact outcome this capability exists to prevent. The passing test used a fixture fabricated from the same wrong belief. The condition is now our own accepted interrupt request plus the subtype, and the fixture is the captured envelope.

Your chain is answered by deletion rather than by an absorber: the settlement on `command_lifecycle: cancelled` is removed, so the artifact is the only place an interrupt can settle and there is no path that leaves it to land on an empty `pending`. Adding a reap-absorber for an ordering neither probe produced would have been defensive code with no named scenario.

packages/agent-runtime/claude-code/src/supervisor.ts
YourWildDad: Fixed. `interruptTurn` now carries the same `this.child === null || this.exited` check its siblings do, with the scenario named in the comment: a child that exited between the last stream line and the call still has an `rpc`, and writing to it surfaces `Command /stop failed: write EPIPE` instead of the honest `No turn is running.`

.agents/tasks/channel/add-feishu-slash-commands/technical-design/final.md
YourWildDad: Fixed. The row now says a `provision` plan is answered without a Command at all, which is what the code does.

packages/dreamux/src/service/team-service/index.ts
YourWildDad: Fixed. The comment now describes the current shape: `dissolve` publishes `dissolveTask` before handing back the receipt, and every refusal that can still be answered has already happened by the time `runDissolve` starts.

packages/channel/feishu-channel/src/feishu-channel.ts
YourWildDad: Fixed. `unavailableTeamReason` in `feishu-route-reconciliation.ts` owns the reading, and both sites call it; the duplicated expression is deleted from both, not wrapped.

packages/channel/feishu-channel/src/feishu-slash-commands.ts
YourWildDad: Ruled by the operator, and recorded verbatim in `requirement.md` so it is not refiled as a gap.

He decided it by code size: 「这个按照代码量来决策吧：哪个代码量最少，就选哪个方案。因为这个功能完全不投递给Agent，我理解只要艾特了就返回，是不是最简单？」 — "decide this by code size: whichever option is least code, take it. Since this feature is never delivered to an Agent, my understanding is it just returns as soon as you @-mention — isn't that the simplest?"

Rendering the Command's own answer is the least code, so `/teams` lists every running Team on the dispatcher to any conversation that may reach the bot. The product catalog and `channel.md` now say so explicitly, including that it is a ruling rather than an oversight.

common/changes/@excitedjs/feishu-channel/dreamux-team-dreamux-claude-ebd3zw_2026-09-04-23-49.json
YourWildDad: Rejected on both halves.

`@excitedjs/feishu-channel` is not at 5.0.0 — it is 6.0.0 on this branch and 6.1.0 on `next`. And the type was already put to the operator in review round 1, who ruled the package stays on `minor`: the `command()` method added to `FeishuInboundDelivery` has exactly one implementer, `FeishuChannelSession`, and one consumer, both inside this package. The external implementer the finding assumes does not exist.

packages/agent-runtime/codex/src/turn-manager.ts
YourWildDad: The premise is wrong on the codex side; the gap you point at is real but different, and it is with the operator.

Read at `rust-v0.137.0`, an interrupted codex turn is **not** an error terminal. `turn/interrupt` leads to `turn/completed` carrying `TurnStatus::Interrupted` with `error: None`. `events.ts` discriminates failure on `params.turn.error != null` only, so the turn settles as a clean `completed` — no `failed` card, and no invented failure text.

What is true is the divergence: a Claude Code interrupt now puts `[Request interrupted by user]` on the card, and a codex interrupt puts nothing there, because `turn.status` is never read. That is a product question, not a bug in this diff, and it is asked of the operator. `provider-runtime.md` records the divergence as known and unresolved rather than leaving it implicit.

packages/channel/feishu-channel/src/feishu-session-inbound.ts
YourWildDad: Real, and deliberately not fixed here — it is with the operator.

A fix is not a one-liner: commands answer with a receipt rather than a Core submission, so there is no Core-side dedupe to reuse, and adding one means a new bounded, TTL-expiring seen-set in the Channel. That is a new persisted-ish mechanism for a failure nobody has reported, which this repo's standing rule says not to add without a named scenario. If he wants it, it is its own change with its own record.

Worth stating what the exposure actually is. Each command has two redelivery windows, and I traced both rather than assuming the late one:

- **`/stop`, redelivered while the first interrupt is still in flight** — `ClaudeCodeControlRpc.interruptTurn` returns the existing `pendingInterrupt.promise` and writes nothing to stdin, so the second delivery joins the first request rather than issuing another `control_request`. **Redelivered after the turn settles** — `pendingInterrupt` is cleared and `pending` is `null`, so it answers `false` and the card says `No turn is running.`
- **`/dissolve`, redelivered while the dissolve runs** — `TeamService.dissolve` sees `dissolveTask !== null` and hands back the same `{accepted, status: 'submitted'}` receipt; no second dissolve is started. **Redelivered after the durable close** — it never reaches `TeamService`: `TeamCollection.open` refuses a `closed` record with `TeamClosedError`, so that one is answered as a public failure, not as a receipt.
- **`/teams`** — a pure read plus a card render, so a redelivery renders the card twice and touches nothing.

An earlier revision of this reply stated only the late window for `/stop` and only the in-flight window for `/dissolve`. Both conclusions held, but the mechanism was half-stated; the list above is what the code does.

packages/agent-runtime/codex/src/turn-manager.ts
YourWildDad: Resolved on `b5192737`, and the divergence you pointed at is now closed for the line itself.

The operator ruled: 「codex 也加一个打断text ，对齐claude code」 — add an interrupt text on the codex side too, aligned with Claude Code.

Rather than repeat this task's earlier mistake, I measured before writing. Two probes against a real `codex-cli 0.153.4` app-server:

| Turn | `turn/completed` payload |
| --- | --- |
| interrupted mid-tool-call | `turn.status: "interrupted"`, `error: null`, `items: []`, `itemsView: "notLoaded"` |
| answered normally | `turn.status: "completed"`, `error: null`, items present |

So the source read in my earlier reply was right, and the wire also settles the two things the source could not: the camelCase spelling reaches the client, and a normal turn says `"completed"` — the status alone is a sufficient discriminator.

What changed: `TurnCompletedNotification.turn` gains `status`, `CollectedTurn` carries it, and `observeTerminal` pushes `[Request interrupted by user]` ahead of the end it already reports. The fixtures now carry `status` too; a fixture that does not model the wire is exactly what let the Claude Code half of this task ship a recognition condition no real interrupt could match.

Reading the status rather than remembering that we asked is deliberate. `interrupt()` picks a turn with no terminal yet and only then sends the request, so codex can finish that turn in the window between the two; a request-flag discriminator would mark it interrupted, and codex's own terminal does not.

What did not change: an interrupted codex turn still ends `completed` and settles its submissions with codex's completion. The ruling names the text, and the operator had already ruled 「最后卡片是什么状态其实没那么重要」. That remainder is recorded in `provider-runtime.md` as a deliberate divergence rather than left implicit.

ryanxiang7 (approved):
## Re-review (round 3): APPROVE

All nine round-2 comments are resolved on `b519273` — six by code changes, three by operator rulings recorded verbatim in `requirement.md`. CI is 9/9 green on this head. I re-read every fix at source, not just the replies.

### F1 (the blocking one) — fixed, and fixed better than I suggested

My round-1 comment assumed the `error_during_execution` artifact had no uuid and argued for absorbing a late artifact after a `cancelled`-first settlement. The author probed a real `claude` 2.1.263 (interrupted stream and interrupted 60s tool call) and overturned both halves of that: the real artifact **carries the interrupted command's own uuid** with `is_error: true`, and it **arrives before** the command's `cancelled` in both probes. The round-1 recognition condition (`uuid === null && text === null`) could never have matched a real interrupt — the unit test passed only because its fixture was fabricated from the same wrong belief.

The fix (`037cd96`) deletes the settlement on `command_lifecycle: cancelled` entirely, so the artifact is the only place an interrupt settles and the late-artifact-reaps-the-child chain I flagged has no entry point — in either terminal ordering. The discriminator is now the runtime's own accepted request (`interruptRequested && subtype === 'error_during_execution'`), which is sound: the flag is rolled back on write failure and on a rejected control response, and a genuine failure nobody asked for stays a failure (new test pins this). The lifecycle-supported test now sends the artifact *and then* the `cancelled`, asserting no reap — exactly the coverage I asked for. The one unprobed case (interrupt of a still-queued command) is recorded in `verification.md` as falling to the pre-existing failure path, not papered over.

I owe the author a correction here: my round-1 fix direction targeted an artifact shape that does not occur. The probe-driven rewrite is the right resolution.

### Fixed as requested

- **F2** `supervisor.ts` — `interruptTurn` now has the `child === null || exited` liveness guard its siblings have; a dead child answers `idle` instead of surfacing EPIPE.
- **F3** `errors.ts` — `TeamDissolveBlockedError` now names a next step ("Commit or move that work, or dissolve with force") and its comment reflects that it reaches a caller; `TeamDissolveFailedError` keeps the "answers nobody" comment, correctly.
- **F7** `team-service/index.ts` — the `runDissolve` comment now describes the fence the caller raised.
- **Design table** `final.md` — the `/stop` row no longer claims a Command for a `provision` plan.
- **Reason mapping** — `unavailableTeamReason` owns the `TEAM_CLOSED → team_closed` reading in one place; both call sites use it with identical semantics.

### F5 (codex) — premise corrected, gap closed

My round-1 comment worried codex might report an interrupted turn as an *error* terminal. The author measured `codex-cli 0.153.4` over the real app-server surface: an accepted `turn/interrupt` is answered with an ordinary `turn/completed` carrying `turn.status: "interrupted"`, `error: null`, no items — so my error-terminal premise was wrong, but the underlying divergence (no card marker) was real. `b519273` reads that status (modelled as required on the wire since 0.137.0) and pushes the same `[Request interrupted by user]` line ahead of the end, per the operator's ruling. Reading codex's own terminal rather than remembering we asked is the right call — it handles the turn-that-finished-between-select-and-interrupt window. Tests cover both the interrupted and the answered terminal.

### Resolved by operator ruling (acknowledged, not blocking)

- **F4** (`/teams` unscoped): ruled by code size — least code wins, recorded verbatim in `requirement.md` and `channel.md`.
- **F6** (no command-path idempotency key): operator deferred — 「可以先不动。」, recorded as a deferral, not as a claim redelivery cannot happen.
- **F8** (change file type): I was wrong on the facts — the package is at **6.1.0** on this head, not 5.0.0 (I read a stale worktree base). The operator ruled `minor` stays in round 1; `FeishuInboundDelivery` is a structural interface with its one in-repo implementation in this same diff. Withdrawing the comment.

### Recorded divergences (acceptable, documented)

- An interrupted codex turn keeps the `completed` end and its completion settlement, while Claude Code ends `interrupted`/`stopped` — the operator ruled the card's end state does not matter once the line is there; `provider-runtime.md` records the divergence instead of closing it silently.
- The inherited #342 no-uuid/no-text branch stays, relabelled as unreproduced rather than asserted.

No new issues found in the fix commits. Approving.

ryanxiang7 (approved):
Re-reviewed the 5 commits added after my b519273 approval (head now 94e31828). APPROVE.

- **codex interrupted end (5e5d9ff):** `endNativeTurn('interrupted')` was already a supported status (teardown path) and the `turn.ended` union already carries it; the COT mapping renders 任务中断. Both card facts now read the one `turn.status` codex reports, and the tests pin both the marker and the end. The remaining divergence (submission settlement) is correctly left as a documented non-card fact.
- **/dissolve silent (91feafc):** `silent` as a third reply shape (not a nullable return) with an exhaustive switch at the render site is the right call. The no-receipt path is backed by the verified `team.state` closed → route removal → `announceTeamClosed` chain; refusal and no-bound still answer in words, as the ruling requires.
- **/teams grid (ea3dc09, 7e4afe):** layout per operator pick; the waterfall premise is the operator's own statement, and reverting both truncations follows it. One-pass binding grouping and `plural()` are clean.
- **one-owner cleanup (94e3182):** the best kind of diff. The interrupt flag now lives only on the turn record `ClaudeRpc` owns — I traced the races (sync write failure, async rejection, turn-ends-before-response, response-first, refused interrupt) and each settles exactly once. `resolveChatName` is optional in exactly one place; verified the transport already normalizes `''` → `undefined` (feishu.ts:512), so dropping the card's re-normalization is safe. The event demultiplexing moving to the session that owns both consumers leaves the reconciliation class with one job; session tests cover the moved path.

Knowledge delta is complete (channel.md, provider-runtime.md, product README, requirement/verification). CI 9/9 green on this head.

packages/dreamux/src/service/dispatcher-service/index.ts
```diff
@@ -514,6 +514,16 @@ export class DispatcherService {
...
 
+  interruptAgent() {
+    return this.admitOperation(() => this.mustAgent().interrupt());
+  }
+
+  interruptTeamLeader(teamId: string) {
+    return this.admitOperation(async () =>
+      (await this.teams.open(teamId)).interruptLeader(),
+    );
+  }
```
YourWildDad: 这里怎么有一个单独打断teamleader的函数？

packages/dreamux/src/service/team-service/closing.ts
```diff
@@ -186,13 +186,25 @@ export class TeamClosing {
...
+  reportDissolveFailure(message: string, error: unknown): void {
+    this.deps.log.error(
+      {
+        dispatcher_id: this.deps.dispatcherId,
+        team_id: this.deps.teamId,
+        err: teamErrorInfo(error),
+      },
+      message,
+    );
+  }
```
YourWildDad: 打印日志还需要单独加一个函数？

ryanxiang7 (approved):
Re-reviewed after the rebase onto current next (head now 99dc642c46, 5 commits). The rebase was not trivial — next absorbed #384 (resident-session settlement), #386 (COT/route cards), #391 (completion fence) since my 94e31828 approval — so I reviewed the port, not just the conflict markers. APPROVE.

**Interrupt ported to the resident-session model:** the flag is now session-scoped and spent on the first result (an accepted interrupt that found nothing to stop ends in an ordinary result; a later failure is nobody's interrupt). Three settlement points — control response, the error_during_execution artifact, session teardown — each settle the ask exactly once; the new test covers the teardown window that previously left /stop waiting on a gone session. The neutral `AgentRuntime.interrupt()` seam is a proper capability addition with its BREAKING change file.

**Dissolve admission redesigned, and the redesign is earned:** /dissolve's refusal must be answerable (operator ruling), which the old receipt-first design could not do — the refusal happened behind the receipt. `team.dissolve` is now async with a pre-admission worktree assessment that refuses while the Team is untouched, while the final assessment after every runtime stops remains the authority that permits close (a writer still cannot invalidate the admission answer). `setImmediate` over `Promise.resolve().then` makes the receipt observable before the background stop begins, which the self-dissolve case actually needs. Three assessments, three distinct roles; change file documents the semantic change.

**Route reconciliation merged with the widened taxonomy:** team_closed / route_ended / stale_route, with the announcement boundary in one place (`unavailableTeamReason`). A TEAM_CLOSED rejection announces only that the route ended — correct, since a pending dissolve can still be refused and only the final team.state proves the close.

Verified no file silently reverts upstream (branch == merge-base check is empty), the /teams grid survived intact, the .agents delta is complete and accurate, and all four Rush gates are green per the verification record. CI 9/9 on this head.

packages/dreamux/src/service/team-service/index.ts
```diff
@@ -442,45 +440,54 @@ export class TeamService {
...
+    // raises the admission fence before anything else can enter the Team.
+    const task = new Promise<void>((resolve, reject) => {
+      setImmediate(() => {
+        void this.runDissolve({ ...input, note }).then(resolve, reject);
+      });
+    });
+    this.dissolveTask = task;
+    void task.catch(() => {});
+    return this.dissolveReceipt();
+  }
```
YourWildDad, line 466: 这个地方为什么要改异步？

packages/dreamux/src/service/team-service/index.ts
```diff
@@ -561,6 +560,11 @@ export class TeamService {
     });
   }
 
+  /** Interrupt the TeamLeader's owned runtime without starting a dormant one. */
+  interruptLeader(): Promise<AgentRuntimeInterruptOutcome> {
+    return this.admit(async () => (await this.leaderService()).interrupt());
+  }
```
YourWildDad, line 565: 这个为什么不能直接外层查询到TeammateService，直接调用到 teammateService 的 interrupt？

ryanxiang7 (approved):
Re-reviewed on 105d7cf (six commits since the 99dc642 approval). The new commits are net-positive: 0c63120 removes the slash-command path as a third route-removal caller (it could steal the closed event's announcement and strand a binding on a failed dissolve) — fewer callers, not more flags; 1c80284 reads interruption from the turn's own terminal_reason instead of a session-level mark; f207540 lets an interrupt ask reach a session with nothing running. The test reversal in feishu-slash-commands.test.ts locks the reversed contract deliberately, documented in the commit and .agents/domains/channel.md. CI 9/9 green. Approving.

PR #369, opened 2026-09-02
Move role skill guidance into MCP descriptions and role prompts

YisenFE:
Closes #368 (solution review Issue).

## Why

The Dispatcher and TeamLeader prompts told the model to load `dispatcher-workflow` / `team-workflow` before any TeamMate, Team, channel, or cron tool call. Both engines defer MCP tool definitions and read a skill body on demand, so that mandate turned two ~5 KB skills into a per-turn read, while most of what they said already lived in the tool descriptions. Dreamux-owned tool inputs carried no descriptions (2 of ~45 did).

## What changes

- **Role prompts** (`base-prompt.ts` ×2, `leader-agent.ts`): the load mandate is replaced by a map of the MCP servers the role has (`teammate`, `team`, `cron`, one `channel-<id>` per configured channel that provides tools) plus "load a tool's definition before calling it". Nothing else in the prompts changes.
- **MCP catalogs**: every input property of the `teammate` (both callers), `team` (both callers), and `cron` tools carries a one-sentence description, verified against the codecs (history filters, `last` paging, `repo.*` defaults, cron `recurring:false`/`tz`, workflow `max_concurrency`). Cautions that only the skills carried moved onto the tool or property they are about: `spawn`/`send` completion push; TeamLeader `dissolve` pre-check and `force` needing the user's confirmation; branch deletion is the user's separate decision; cron due-job submission. The malformed `spawn.agent_runtime` text is fixed.
- **Feishu channel tools**: `reply` says text outside the tool is not delivered and what it is for; `reply.message_id` threads under the inbound message; `reply.text` keeps secrets, private identifiers, hidden instructions, and machine-local paths out of broad replies; `react` drops the "dispatcher channel" wording (a TeamLeader reaches it too); Dispatcher `bind_channel.team_name` states that a missing or closed Team is refused; TeamLeader `bind_channel` states its own authority instead of the banned "ask the Dispatcher" instruction.
- **Bundled skills**: `dispatcher-workflow` and `team-workflow` keep their names, roots, and injection and become on-demand TeamMate-collaboration methodology (TeamMate vs engine-native subagent, and Team for the Dispatcher; writing the hand-down prompt; asking a TeamMate why before overriding; keeping the thread). Descriptions say when they are useful and that tool calls do not depend on them; bodies carry no tool contracts.
[...]

PR #380, opened 2026-09-06
Refine the model-facing surfaces of the Dispatcher and TeamLeader

YourWildDad:
Supersedes #369 (ruled not merged; its content is included here).

## Intent

Reduce what Dreamux injects into a Dispatcher's and a TeamLeader's context to what the model actually needs: an identity and a map of its Dreamux MCP servers in the prompt, a consequence stated once at the action as a reminder or tool-description sentence, and a skill loaded only when the model is about to use the tool the skill is about. Task record: `.agents/tasks/mcp/refine-model-facing-surfaces/` (requirement with the operator's rulings R1–R33, approved design `technical-design/final.md`, verification).

## What changes

- **Role prompts.** The TeamLeader prompt is its identity line, its MCP server map, and the operator's identity text. The Dispatcher prompts (Codex replace text and Claude Code append text) keep the server map, the `dreamux-maintenance` trigger, the working-directory identity, and the host boundary; the no-polling, MCP-authority, channel-reply, and secrets sentences are gone.
- **Dispatch-result reminders** (`spawn`, `send`, `team.send`, `workflow_run`, and now `team.create` when called with a `prompt`) state the consequence the caller cannot see: the result arrives later as a pushed message. The hand-off tool descriptions end with the same fact, which on Claude Code is the only place the model can read it. The `repo` descriptions of `spawn` and `team.create` now state the workspace policy (a fresh directory, or the dispatcher's own directory when workspace isolation is disabled).
- **Skills.** `team-workflow` → `teamwork`, shared `workflow` → `dynamic-workflow`; the three bundled skills carry narrow load triggers in their descriptions and are pointed at from the tool descriptions. `teamwork` teaches the hand-down as context, not control, and treats a TeamMate's "cannot" as a finding. The two new names are reserved like the old ones (see the `@excitedjs/dreamux` change note).
- **Channel MCP server named by provider.** `channel-<provider>` (`channel-feishu`, identity `dreamux-channel-feishu`) instead of `channel-<configured id>`. The fully qualified tool name a model calls changes on both engines; local tool names, schemas, results, `channels[].id`, and routing state do not. `ChannelProviderCatalog.resolve` returns `{ id, implementation }`, the shape of its agent-runtime twin (five callers adapted, type-only).
- **Feishu.** Both the inbound envelope and the ask-card settlement envelope carry a `source="feishu"` attribute (its position is not a contract); the per-message reminder states the consequence the channel owns ("The user in this chat sees only what you send through the reply tool; your assistant text is not shown to them.").
- **Team-scoped TeamMates** are told which TeamMate of which Team they are; an ordinary TeamMate is also told that its TeamLeader receives what it outputs when the turn ends, a workflow agent is not (the workflow consumes its output).
- **TeamLeader workspace.** The TeamLeader prompt states whether Dreamux removes the Team's workspace on dissolve (managed worktree, `delete-on-close` or `keep`, or a reused directory), and the `dissolve` description asks it to check a workspace that is removed for uncommitted, untracked, or unmerged work before dissolving. A managed worktree requested without `cleanup` now defaults to `delete-on-close`; pass `cleanup: keep` to retain one.
- **Completion push-backs** say once that they are an automated Dreamux notification, not a message from the user.
- **Claude Code skill adapter.** The adapter manifest (version 2) now records each source root's child skill inventory and is compared on every start, so an in-place package upgrade that renames bundled skills refreshes the view instead of serving the stale one. The root stays keyed by the set of source roots, so the same roots map to the same directory and a refresh replaces it in place (a root left by the previous release is refreshed, not orphaned). A custom root that cannot be read fails the start with an error naming the source and its path.
- Knowledge base, task records, and three Rush change files (`@excitedjs/dreamux` minor, `@excitedjs/feishu-channel` minor, `@excitedjs/agent-runtime-claude-code` patch; #369's two files are replaced).

## Validation

- `rush build`, `rush lint`, `rush test` (live Codex tests ran against codex 0.153.4 and observed the `channel-feishu` server with its `reply` tool), `rush typecheck:tests`, `.agents/scripts/check.sh`, `rush change --verify --target-branch origin/next --no-fetch`, `git diff --check`: green.
- Built-artifact probes: both Dispatcher prompts, the bundled skill names, and the reminder texts from `dist/`; the Claude skill view materialized against a copy of a live adapter cache lands on a new root with `teamwork` and `dynamic-workflow` and leaves the old roots untouched.
- Live probe on Claude Code (2.1.260): a real `Server` from this build with a fake Feishu backend, the adapter cache pre-seeded with pre-rename roots; the Dispatcher replied through `channel-feishu` listing the eleven `mcp__channel-feishu__*` tools, the skills `dispatcher-workflow`, `dreamux-maintenance`, `dynamic-workflow`, and the servers `teammate, team, cron, channel-feishu`; a fourth adapter root with a version-2 manifest appeared next to the untouched old ones.
- Independent implementation review through Dreamux's own workflow (Codex and Seed nodes): seven finders, one verifier per source location, sweep, synthesis; six confirmed findings, all accepted and fixed or recorded; a follow-up cleanup review added one accepted finding. Details and adjudication in `verification.md`.
- Alpha acceptance (`0.24.0-alpha.gffd8b4b7feab`): a Codex Dispatcher, a Claude Code TeamLeader, and a Codex TeamLeader on the installed package each answered the acceptance probes of the design (`verification.md`, "Alpha acceptance").
- External review round: eight findings, ruled by the operator (R28–R33) and answered in one commit; each thread on this PR carries its disposition (`verification.md`, "External review of #380").

## Limitations

- Not exercised live after the review round: the in-place refresh of a real adapter root on an installed package (covered by the materializer tests against a version-1 root) and the TeamLeader workspace sentence on a managed worktree (covered by the prompt tests; the alpha probes ran Teams without a repository).
- Description-driven skill loading is the engine's heuristic.
- An external (`npm:`) channel provider's id as a server-name segment is unverified on both engines; none exists yet, and the constraint is recorded where the first author will look.


common/changes/@excitedjs/dreamux/refine-model-facing-surfaces_2026-09-06-03-05.json
```diff
@@ -0,0 +1,10 @@
...
...d `teamwork` or `dynamic-workflow` is rejected at creation, and a Team created before this release with such a child fails loudly at its TeamLeader's next start on Claude Code (`duplicate Claude skill name`); rename the custom skill. The channel MCP server is now named after its provider instead of its configured channel id, so the fully qualified tool name a model calls changes on both engines — `mcp__channel-<id>__reply` becomes `mcp__channel-feishu__reply` on Claude Code, and `channel_<id>__…` becomes `channel_feishu__…` on Codex — while the tool's own local name, its input/output schema, its result shape, `channels[].id`, and routing state are all unchanged. A Team-scoped TeamMate's system prompt now opens with an identity sentence naming its Team and stating that its TeamLeader receives its output when the turn ends. A pushed completion notification now states once that it is an automated Dreamux notification, not a message from the user. Every Dreamux MCP tool input property carries a description, and tool descriptions carry the operational cautions role prompts used to state — this continues PR #369's relocation work, which was not merged and ships through this PR instead.",
```
ryanxiang7: **Missing `BREAKING:` / `Rebuild:` on a note that describes an upgrade blocker.**

The note itself states the blocker: *"a Team created before this release with such a child fails loudly at its TeamLeader's next start on Claude Code (`duplicate Claude skill name`); rename the custom skill"*, plus a model-facing tool-name change (`mcp__channel-<id>__reply` -> `mcp__channel-feishu__reply` on both engines).

CLAUDE.md "Changelog Responsibility" lists **bundled skills** as a typical upgrade blocker and requires that breaking notes *"lead with `BREAKING:` and include `Rebuild:` with the exact manual action"*. There is direct precedent for this exact class of change at `packages/dreamux/CHANGELOG.md:42`:

> BREAKING: Add Dynamic Workflow orchestration ... and a required bundled workflow skill. Rebuild: remove or rename any custom workflow skill before restarting existing Dispatcher or TeamLeader identities.

As written this ships as a plain `minor` entry, so an operator upgrading in place gets no warning, hits a TeamLeader that will not start, and has no `Rebuild:` line naming the fix. `type: minor` is correct for the 0.x line — it is only the `BREAKING:` / `Rebuild:` framing that is missing.

packages/dreamux/src/service/teammate-collection/system-prompt.ts
```diff
@@ -0,0 +1,29 @@
...
+ * TeamMate has no TeamLeader, so it is told none of it.
+ */
+export function teammateSystemPromptOptions(
+  identity: AgentEntityIdentity,
+  operationAppend: readonly string[] | undefined,
+): { systemPrompt: AgentRuntimeSystemPrompt } | undefined {
+  const append = [
+    ...(identity.team_id !== null
+      ? [
+          `You are TeamMate ${JSON.stringify(identity.name)} of Dreamux Team ${JSON.stringify(identity.team_id)}. Your TeamLeader receives what you output when your turn ends.`,
```
ryanxiang7: **A Team-scoped workflow agent is given two contradictory output contracts, and this one is the false one.**

`workflow-service/run.ts:347` creates workflow agents through `TeammateCollection.createLocked` with `systemPromptAppend: [WORKFLOW_AGENT_SYSTEM_PROMPT]`. Inside a Team `identity.team_id !== null`, so the composed append is:

1. `You are TeamMate "..." of Dreamux Team "...". Your TeamLeader receives what you output when your turn ends.`
2. `Your final response is the return value consumed by the workflow, not a human-facing progress message. Return only the requested value...`

Sentence 1 is not true for a workflow agent: `run.ts` submits it with no completion initiator, so the turn's output is consumed by the workflow script — only the run's *terminal* completion is delivered to the caller. A model that reads sentence 1 first writes a leader-addressed progress report instead of the schema-conforming value the node consumes, which is precisely the failure `WORKFLOW_AGENT_SYSTEM_PROMPT` exists to prevent. `tests/teammate-system-prompt.test.ts` currently pins this ordering, so the test locks the contradiction in.

Suggestion: suppress the TeamLeader sentence when an `operationAppend` already owns the output contract, or let the workflow path opt out of the identity sentence.

packages/dreamux/src/service/team-collection/mcp-delegate.ts
```diff
@@ -242,10 +251,25 @@ function teamToolDescriptors(
...
+        'Call this only when the Team\'s work is complete. First check the workspace for uncommitted, untracked, or unmerged work; if there is any, or you cannot tell, do not dissolve: report it and ask the user. Submit a dissolve of this descriptor-bound Team. It returns a receipt as soon as the request is accepted ({ accepted, team_name, status: submitted }) and never reports how the dissolve went: the Team\'s Workflow, TeamMates, and this TeamLeader are stopped behind that receipt, so expect this call to lose its response. note is required and records why the Team stopped. Uncommitted, untracked, or unmerged work in a managed delete-on-close worktree leaves the Team open and running instead of closing it. force: true only overrides a delete-on-close removal blocked by uncommitted, untracked, or unmerged work, by discarding that work; under cleanup: keep the checkout and its changes are retained; never the branch, its commits, a reused directory, or the source repository; deleting them is a separate decision that is the user\'s.',
```
ryanxiang7: **The new opening rule contradicts the narrower rule later in the same string, and only the later one matches Core.**

- Opening (new here): *"First check the workspace for uncommitted, untracked, or unmerged work; if there is any, or you cannot tell, do not dissolve"*.
- Later in the same description: *"Uncommitted, untracked, or unmerged work **in a managed delete-on-close worktree** leaves the Team open"*.

Core agrees with the second one: `assessCleanup` (`src/service/worktree/manager.ts`) returns `terminal` for anything that is not `managed` + `delete-on-close`, so outside that one mode the work is retained no matter what and the dissolve is never blocked. A Team created with `repo` omitted goes through `prepareDefaultWorkspace` -> `reuseCwdWorktree(path)`, i.e. mode `reuse-cwd` — the common case — and such a Team almost always leaves untracked files behind (build output, scratch files, logs).

Net effect: a TeamLeader that follows the opening sentence literally never dissolves a default Team, and its Workflow, TeamMates and scheduler stay alive indefinitely. Note also that the dispatcher-facing `dissolve` below did *not* get this sentence, so the two callers now read different rules for the same operation.

Scoping the opening sentence to the delete-on-close case would make it agree with Core, with the rest of this paragraph, and with the dispatcher-facing twin.

ryanxiang7 (commented):
Reviewed the whole diff: the ~2.2k-line source change read in full, the ~4k lines under `.agents/` skimmed for consistency. CI is green on both runners.

The design holds up. Two things I checked specifically and found correct:

- `ChannelProviderCatalog.resolve` -> `{ id, implementation }` is threaded through all five callers (`mcp-delegates`, `channel-service/index`, `onboard/wizard`, and both `provider-diagnostics` sites); type-only, nothing behavioural.
- The uniqueness claim behind naming the server by provider holds: `config.ts` rejects a duplicate provider per dispatcher on the *normalized descriptor* ref (`formatProviderRef(descriptor.ref)`), not on the raw config string, so two aliases of the same provider cannot slip past and produce two `channel-feishu` servers.

Seven inline comments. Three I would want addressed before merge — the change-file framing, the workflow-agent prompt contradiction, and the TeamLeader `dissolve` description — one design question on the adapter cache, and three nits.

One more finding with no diff line to anchor to: **`.agents/proposals/admin-control-plane-surface.md:286`** still reads *"Custom roots cannot shadow the bundled `team-workflow` skill name"*. It is the only live (non-archived) proposal, and that skill name no longer exists after this PR. The archived proposals that mention `team-workflow` / `workflow` are historical record and are fine as they are.

For the record, one thing I deliberately did **not** raise: the disappearance of the public-artifact secrets rule from every model-facing surface. `technical-design/final.md` 3.12 states it outright and R4 / R2 / R19 settle it, so it is a disclosed operator decision rather than an oversight.

packages/agent-runtime/claude-code/src/skill-materializer.ts
```diff
@@ -11,27 +11,37 @@ import { dirname, join } from 'node:path';
...
+  cacheDir: string,
   sources: readonly AgentRuntimeSkillSource[],
   testHooks: { beforePublish?: () => void | Promise<void> } = {},
-): Promise<void> {
-  if (sources.length === 0) return;
-  if (await validateSkillAdapter(skillAddDirRoot, sources)) return;
-  const tmpRoot = `${skillAddDirRoot}.${randomUUID()}.tmp`;
+): Promise<string | null> {
+  if (sources.length === 0) return null;
+  const manifest = await readSkillAdapterManifest(sources);
```
ryanxiang7, line 37: Two consequences of moving the adapter key onto the on-disk inventory. The fix this PR makes is right; these are the two edges it opens.

**1. Adapter roots now accumulate without bound.** The key hashes every child skill's name and path, so *any* add / remove / rename in *any* source root produces a fresh `<cacheDir>/claude-code/skills/<key>/` tree — and nothing ever removes the previous one. The doc comment says earlier roots "are never revalidated again", and `skill-materializer.test.ts` asserts the old root stays byte-identical. With a user root like `~/.claude/skills`, a set that used to be stable becomes one that grows on every skill edit, and there is no sweep in this package nor in `dreamux-maintenance`. Sweeping sibling roots whose manifest `key` is not the current one, after a successful publish, would bound it cheaply.

**2. A source root that disappeared now fails the start rather than the validation.** `readSkillAdapterManifest` readdirs every root *before* `validateSkillAdapter` runs; previously validation was computed from `sources` alone and the cached-manifest fast path never touched the roots. So a custom `skill_sources` root deleted or moved after creation now raises a bare `ENOENT ... scandir` out of `materializeClaudeSkillAddDir` -> `startRuntime`, and the agent cannot be resumed at all. Fail-loud may well be the intent here, but it currently surfaces as an unnamed errno with no actionable text, and the behavior change is not in any change note.

packages/dreamux/src/service/mcp/tool-metadata.ts
```diff
@@ -150,12 +150,53 @@ export function repoInputSchema(): Record<string, unknown> {
...
+      slug: {
+        type: 'string',
+        minLength: 1,
+        maxLength: 64,
+        description:
+          'managed: label for the worktree directory and the default branch ' +
+          'name; defaults to the TeamMate\'s name, or team-<team_name> for a ' +
+          'Team.',
+      },
+      cleanup: {
```
ryanxiang7, line 193: Nit, in the spirit of this PR: every sibling property here states its default (`base_ref` "default HEAD", `branch` "default dreamux/<slug>", `slug` "defaults to the TeamMate's name"), but `cleanup` does not. The default is `keep` (`src/service/worktree/manager.ts`, `input.request?.cleanup ?? 'keep'`).

It matters more than usual after this PR, because both `dissolve` descriptions now branch on exactly that distinction (`managed delete-on-close` vs `cleanup: keep`) for the blocking rule *and* for what `force` means. A model that created a worktree without passing `cleanup` has no way to tell which branch applies to it, and the safe assumption is the destructive one.

packages/dreamux/src/service/team-collection/mcp-delegate.ts
```diff
@@ -259,14 +283,50 @@ function teamToolDescriptors(
...
+        },
+        identity: {
+          type: 'string',
+          minLength: 1,
+          maxLength: 4000,
+          description:
+            'Standing role and boundaries appended to the TeamLeader\'s system ' +
+            'prompt for every turn.',
+        },
+        prompt: {
```
ryanxiang7, line 323: Nit: `team.send.prompt` (below) and `teammate.spawn.prompt` both carry `minLength: 1`; this one does not.

`optionalString` returns `""` rather than `null` for an empty string, so `create({ ..., prompt: "" })` passes schema validation, submits an empty first TeamLeader turn, and — with the new reminder above — also gets `TEAM_DISPATCH_SUCCESS_REMINDER` attached, telling the caller "The Team is working; Dreamux will notify you automatically when its TeamLeader finishes" for a turn with no content. `minLength: 1` would match the two siblings.

packages/channel/feishu-channel/src/feishu-message.ts
```diff
@@ -127,6 +127,9 @@ export async function formatFeishuMessageForRuntime(
   }
   const attachments = resolution.attachments;
   const attrs: Array<[string, string]> = [];
+  // First, and constant: the provider name is what lets the model tie this
+  // envelope to the `channel-feishu` MCP server whose tools answer it.
+  attrs.push(['source', 'feishu']);
```
ryanxiang7, line 132: Nit / contract drift: attribute *position* is now load-bearing — `channel-input-format.test.ts` asserts `startsWith('<channel source="feishu" ')` and `feishu-settlement-envelope.test.ts` pins the exact key order — but `TeammateSubmitInput.attrs` is still documented as *"Unordered display attributes"* (`src/service/teammate-service/submission.ts`, outside this diff), and `renderSubmission` just iterates `Object.entries(input.attrs)`.

So the guarantee rests on insertion order while the written contract says there is none. A later change that honours the doc comment — sorting attrs for stable rendering, or a channel building the record via `Object.fromEntries` — would silently move `source` out of first position and break nothing in `submission.ts`'s own tests. Either the doc comment should state that insertion order is preserved and meaningful, or "source first" should not be load-bearing.

common/changes/@excitedjs/dreamux/refine-model-facing-surfaces_2026-09-06-03-05.json
YourWildDad: Fixed in f009a0d4. The note now leads with `BREAKING:` (the two reserved names, the provider-named server and the changed fully qualified tool name, the prompt changes) and carries a `Rebuild:` line: rename a custom skill directory named `teamwork` or `dynamic-workflow` before restarting; creation rejects the name and a Claude Code TeamLeader fails with `duplicate Claude skill name`. Same shape as the workflow skill's own note. It also records the new `cleanup` default (see the `tool-metadata.ts` thread).

packages/dreamux/src/service/teammate-collection/system-prompt.ts
YourWildDad: Agreed, and the operator ruled the same way (R28 in the task record): a workflow agent already struggles to produce stable structured output, so nothing may compete with the workflow contract. In f009a0d4 a workflow agent receives only `You are TeamMate "<name>" of Dreamux Team "<team_id>".`; the sentence about the TeamLeader receiving the output is appended only when there is no operation append, i.e. for an ordinary TeamMate, where it is true. `teammate-system-prompt.test.ts` pins the workflow case.

packages/dreamux/src/service/team-collection/mcp-delegate.ts
YourWildDad: Confirmed. My first fix only scoped the wording to a managed `delete-on-close` worktree; the operator pointed out that the leader cannot know which case it is in and has to be told at start (R29). In f009a0d4 the TeamLeader prompt carries one sentence built from the Team record's worktree identity, e.g. `Your Team's workspace <path> is a managed git worktree that Dreamux removes when the Team dissolves; uncommitted, untracked, or unmerged work there blocks the dissolve.` (or `... is kept after the Team dissolves.` for a kept worktree / reused directory), and the description now opens with `Your system prompt says whether Dreamux removes the Team's workspace on dissolve. If it does, first check ... A kept workspace never blocks the dissolve.` `team-leader-prompt.test.ts` covers the three cases.

packages/agent-runtime/claude-code/src/skill-materializer.ts
YourWildDad: Both edges are real, and the growth one changed the shape (R30: the operator asked for one stable directory). In f009a0d4 the root is keyed by the set of source roots again (the version-1 key), so one set of roots always maps to one directory; the version-2 manifest records the child inventory, and a mismatch (renamed child under an unchanged root, a version-1 manifest, a malformed target) rebuilds the view in `<root>.<uuid>.tmp`, renames the old root aside, swaps the new one in, and removes the stale tree. A root left by the previous release is refreshed in place rather than orphaned, and no earlier view accumulates. A root that cannot be read fails the start with `skill source "<name>" root <path> cannot be read: <reason>`. `skill-materializer.test.ts` covers the rename, the version-1 root, the concurrent malformed target, and the unreadable root.

packages/dreamux/src/service/mcp/tool-metadata.ts
YourWildDad: The operator went further than the nit (R31): a managed worktree requested without `cleanup` now defaults to `delete-on-close`, because kept worktrees were piling up. `worktree/manager.ts` applies the default, `reuse-cwd` stays `keep` and is never removed, the property description states the default, and the change note records it (identities created before this release keep the `cleanup` they recorded).

packages/dreamux/src/service/team-collection/mcp-delegate.ts
YourWildDad: Fixed in f009a0d4: `team.create.prompt` carries `minLength: 1` like `team.send.prompt` and `teammate.spawn.prompt`. An empty string used to submit an empty first turn with the Team reminder attached.

packages/channel/feishu-channel/src/feishu-message.ts
YourWildDad: Agreed; the operator ruled the position is not a contract (R33). In f009a0d4 `channel-input-format.test.ts` and `feishu-settlement-envelope.test.ts` assert that `source="feishu"` is present among the attributes, not that it comes first, and the design, `channel.md`, and the feishu-channel change note no longer say "first". The channel still adds it first and Core still renders attrs in insertion order, but nothing may rely on that.

YourWildDad:
Thanks for the read. All eight points are answered in f009a0d4; each inline thread carries its disposition. The unanchored one: `.agents/proposals/admin-control-plane-surface.md` now names `teamwork`; the archived proposals keep the historical names, as you said.

Two of the rulings went further than the comments asked, both from the operator: the TeamLeader prompt now states whether Dreamux removes the Team's workspace on dissolve (the `dissolve` description points at that sentence instead of asking every leader to inspect a workspace whose fate it cannot know), and a managed worktree requested without `cleanup` defaults to `delete-on-close`. The adapter cache is back on one directory per set of source roots, refreshed in place when the recorded inventory no longer matches. The task record (`requirement.md` R28–R33, `verification.md` "External review of #380") lists what each finding became.

packages/dreamux/src/service/mcp/tool-metadata.ts
YourWildDad: Follow-up in 26b19ad8: the Team runtime registry still built a fallback request with `cleanup: keep` for a `repoCwd` that arrived without a worktree request. Both callers derive the pair from `repoWorktree()`, which owns the managed-by-default policy, so the fallback was unreachable and would have pinned the old default; it is removed and the manager holds the only default. The maintenance skill's service-lifecycle reference now records what the Team record's `worktree.cleanup` is written from.

packages/agent-runtime/claude-code/src/skill-adapter.ts
```diff
@@ -24,62 +50,76 @@ export async function validateSkillAdapter(
...
+    value !== null &&
+    typeof value === 'object' &&
+    !Array.isArray(value) &&
+    JSON.stringify(value) === JSON.stringify(expected)
+  );
+}
+
+/**
+ * Read the adapter's identity from disk — the source roots and the skill
+ * directories under each. Every adapter root is keyed by this manifest, so a
```
ryanxiang7, line 78: Non-blocking, and only because it is the exact class of defect this PR exists to remove: this doc comment is left over from the previous revision and now contradicts the code and its two siblings.

`skillAdapterKey` hashes `inventory.map(({ name, path }) => ...)` — roots only, children dropped — and the comments on `SkillAdapterSource` and on `skillAdapterKey` both say so ("keyed by the roots alone", "names the directory for one set of roots and nothing else"). But this one still says *"Every adapter root is keyed by this manifest, so a renamed child lands on a new key instead of reusing the stale view"*, which is precisely the behaviour 26b19ad/f009a0d replaced: a renamed child now keeps the same key and refreshes the view in place.

packages/dreamux/src/service/teammate-collection/mcp-tool-descriptors.ts
```diff
@@ -293,8 +437,18 @@ export function teammateToolDescriptors(
       'close',
       'Close a named TeamMate agent and retain its history; send reopens it later. note is required: it records why a recoverable session was stopped.',
```
ryanxiang7, line 438: Non-blocking, but worth a follow-up given R31: now that a managed worktree requested without `cleanup` defaults to `delete-on-close`, `close` is the action that removes it, and this description does not say so.

`teammate-service/index.ts` cleans up on `transitionToClosed` when `ownsWorktreeOnClose && mode === 'managed' && cleanup === 'delete-on-close'` — and dispatcher-scoped TeamMates are exactly the ones with `ownsWorktreeOnClose: true` and the only caller offered `spawn.repo`. So a Dispatcher that spawns with `repo: { mode: 'managed' }` and later closes the TeamMate removes the checkout, having read that consequence only in the nested `repo.cleanup` property description, possibly many turns earlier.

By this PR's own principle the consequence belongs at the action: both `dissolve` descriptions state it for Teams, and R29 even puts it in the TeamLeader's system prompt. `close` is the TeamMate-side twin and got nothing. (The branch and its commits survive either way, and a dirty tree still blocks the removal, so this is about what the model is told, not about data loss.)

I checked the rest of the default flip and it holds up: `reuse-cwd` still records `keep`; pre-existing identities keep the `cleanup` they recorded; `assessCleanup` is unchanged; and no shutdown path closes a dispatcher-scoped TeamMate, so a daemon restart cannot remove a worktree.

ryanxiang7 (approved):
Approving. Re-reviewed the three new commits against every thread.

All seven are fixed, and four went further than the comment asked:

- **Change note** — now leads with `BREAKING:` and carries a `Rebuild:` naming the exact rename. Matches the `packages/dreamux/CHANGELOG.md:42` precedent.
- **Workflow agent prompt** — a workflow agent now gets the membership sentence without the "your TeamLeader receives your output" clause, which was the false half. The discriminator is "an operation append exists"; slightly indirect, but the doc comment says why and today `WORKFLOW_AGENT_SYSTEM_PROMPT` is the only one.
- **`dissolve`** — better than what I suggested. Rather than only narrowing the wording, R29 puts the fact in the TeamLeader's system prompt (built from the Team record's own worktree identity) and has the description defer to it, so the leader is told which case it is in instead of being asked to infer it. Verified the three prompt branches, and that `mustRecord()` is safe at both `createLeader` call sites (the record is published at index.ts:255 and :359, before :266 and :374).
- **Skill adapter** — the right fix for both edges. The key is the root set again, so one set of roots maps to one directory and nothing accumulates; the v2 manifest carries the inventory and a mismatch refreshes the same path. `publish()`'s aside-and-swap is sound for the interleavings I worked through — a loser on the final rename validates the winner's view, and a v1 manifest or half-written tree is treated as stale rather than orphaned. Unreadable roots now fail with a named error.
- **`repo.cleanup`** — R31 went past the nit and flipped the default to `delete-on-close`. I traced the blast radius and it is bounded: `reuse-cwd` still records `keep`, `assessCleanup` is unchanged so a dirty tree still blocks, the branch and its commits are never touched, pre-existing identities keep their recorded `cleanup`, and no shutdown path closes a dispatcher-scoped TeamMate — so a daemon restart cannot remove anything. `.agents/domains/dispatcher-orchestration.md` and `.agents/product/README.md` both record the new default.
- **`team.create.prompt` `minLength: 1`**, and the **attrs-order tests** relaxed to membership with the "first" claim dropped from the design, `channel.md`, and the change note — the ordering is no longer a contract, which is the cleaner of the two resolutions.
- The `.agents/proposals/admin-control-plane-surface.md` stale name from my summary is fixed too.

I also checked the follow-up in 26b19ad independently: `repoWorktree()` returns `cwd` and `worktree` as a pair or `null`, and both `create` call sites (`mcp-delegate.ts` and `commands.ts`) spread them together, so `repoCwd` set with `worktree` unset was indeed unreachable and the removed `cleanup: 'keep'` fallback would only have pinned the old default. CI is green on both runners.

Two non-blocking notes inline — a stale doc comment and one missing consequence sentence on `close`. Neither needs to hold the merge.

PR #382, opened 2026-09-06
Name only the workspace and its cleanup mode in the TeamLeader prompt

YourWildDad:
Follow-up to #380, before the beta.

## Why

Reading the report of the #380 review round, the operator ruled (R34 in `.agents/tasks/mcp/refine-model-facing-surfaces/requirement.md`) that the TeamLeader prompt's workspace sentence must name only the worktree and its cleanup mode, and that everything about dissolving belongs in the `dissolve` tool description alone: a leader that starts work should not meet the word "dissolve" in its prompt, since it may not need the concept for a long time.

## What changes

- The TeamLeader prompt sentence is now `Your Team's workspace <path> is a managed git worktree (cleanup: delete-on-close).`, or `(cleanup: keep)`, or `… is a reused directory (cleanup: keep).`
- The TeamLeader-facing `dissolve` description defines what each cleanup mode means at dissolve: under `delete-on-close` Dreamux removes the managed worktree, so the leader checks it for uncommitted, untracked, or unmerged work first; under `keep`, and in a reused directory, nothing is removed and nothing blocks the dissolve.
- `team-leader-prompt.test.ts` pins the three sentences and asserts the prompt does not mention dissolving outside the server map's `team` entry.
- **From the review (R35).** `team.status` reports `worktree_mode` and `worktree_cleanup_mode` next to the lifecycle state `worktree_cleanup`. The lifecycle state reads `managed-active` for every open managed worktree, so the Dispatcher, whose `dissolve` and `force` are phrased in `delete-on-close` / `keep` terms, could not tell whether a dissolve removes anything once the default flipped to `delete-on-close`. `list` and `history` stay compact; the Dispatcher-facing `dissolve` description names the two fields; the `team.status` output schema is an open object, so no schema change. `team-view.test.ts` pins both fields.
- The design (§3.1, §3.14), the dispatcher-orchestration page, the task record, and the pending `@excitedjs/dreamux` change note follow. The change note moves to a new filename because `rush change --verify` counts only added files.

## Validation

`rush build`, `rush lint`, `rush test`, `rush typecheck:tests`, `.agents/scripts/check.sh`, `rush change --verify --target-branch origin/next --no-fetch`, `git diff --check`: green.


packages/dreamux/src/service/team-collection/mcp-delegate.ts
```diff
@@ -251,7 +251,7 @@ function teamToolDescriptors(
...
...ace and its cleanup mode. Under cleanup: delete-on-close Dreamux removes the managed worktree when the Team dissolves, so first check it for uncommitted, untracked, or unmerged work; if there is any, or you cannot tell, do not dissolve: report it and ask the user. Under cleanup: keep, and in a reused directory, nothing is removed and nothing blocks the dissolve. Submit a dissolve of this descriptor-bound Team. It returns a receipt as soon as the request is accepted ({ accepted, team_name, status: submitted }) and never reports how the dissolve went: the Team\'s Workflow, TeamMates, and this TeamLeader are stopped behind that receipt, so expect this call to lose its response. note is required and records why the Team stopped. Uncommitted, untracked, or unmerged work in a managed delete-on-close worktree leaves the Team open and running instead of closing it. force: true only overrides a delete-on-close removal blocked by uncommitted, untracked, or unmerged work, by discarding that work; under cleanup: keep the checkout and its changes are retained; never the branch, its commits, a reused directory, or the source repository; deleting them is a separate decision that is the user\'s.',
```
ryanxiang7, line 254: Non-blocking, and adjacent rather than in this diff — but R34's own argument points at it.

R34's premise is that the leader must be *told* its cleanup mode because it cannot see it from inside the directory, and this description now leans on that ("Your system prompt names the Team's workspace and its cleanup mode"). The dispatcher-facing `dissolve` twin below (line ~482) makes the same demand — *"Uncommitted, untracked, or unmerged work in a managed delete-on-close worktree leaves the Team open"*, and `force` is defined entirely in terms of `delete-on-close` vs `keep` — but a Dispatcher has no surface that reports the mode:

- `list`, `history`, and `status` all project `worktree_cleanup: team.worktree.cleanup_state` (`team-collection/read-model.ts`, `team-service/team-view.ts`), i.e. `managed-active` / `kept` / `deleted` / `not-managed` — never the `cleanup` mode itself. The field name reads like the mode, which makes the gap easy to miss.
- Before R31 the Dispatcher could infer it (`keep` unless it passed `cleanup` itself). Now the default is `delete-on-close`, so the inference is gone exactly where it matters most.

So after this PR the leader knows and the Dispatcher — the caller that can dissolve a Team it did not create — still has to guess. Projecting `worktree.cleanup` alongside `worktree_cleanup` on `team.status`, or naming where to read it in the dispatcher-facing description, would close it. Fine as a separate follow-up; nothing here is wrong.

.agents/tasks/mcp/refine-model-facing-surfaces/requirement.md
```diff
@@ -266,6 +266,15 @@ a recommendation each, and the operator ruled on 2–7 in one message
   envelopes; its position is not a contract, and the two assertions that
   pinned it are removed.
 
+After the merge of #380 (2026-09-06 15:38) the operator read the R29 sentence
+and ruled on it (15:40, follow-up PR):
+
+- R34 (the TeamLeader's workspace sentence): "应该是这个样子，就是这个地方只写。WorkTree和清理模式。只有他自己需要解散的时候，才把解散相关内容写到解散工具的 Description 里。\n不然他刚开始干活时，前面的提示词里就有一个"解散"，然后怎么怎么样。他可能在很长一段时间都不需要感知到解散。"
```
ryanxiang7: Nit: the verbatim quote carries a transcription artifact — a literal `\n` mid-sentence (`...写到解散工具的 Description 里。\n不然他刚开始干活时...`) and an inner unescaped `"解散"` inside the `"`-delimited quote, so the quote's own boundary is ambiguous.

The other rulings in this file quote the operator cleanly (e.g. `R4: "这个保密的话，我感觉可以直接去掉"`). Since these quotes are the record future readers adjudicate against, worth normalising the newline and the nesting.

ryanxiang7 (commented):
Reviewed the whole PR. Text-only — a prompt sentence, a tool description, tests, and the records — no behavioural surface touched, and CI is green on both runners.

The change itself is right and the reasoning holds:

- `teamWorkspaceSentence` collapses to `${kind} (cleanup: ${workspace.cleanup})`, and the `mode` union is `'reuse-cwd' | 'managed'`, so the binary is total — no unhandled case behind the ternary.
- The mode a leader can actually be shown is sound: every entry point (`teammate.spawn`, `team.create`, and both admin Commands) projects `repo` through `repoWorktree()`, which drops `cleanup` on the `reuse-cwd` branch, so a reused directory always records `keep`. There is no reachable "reused directory (cleanup: delete-on-close)" that would read as a removal promise the code does not keep.
- The `dissolve` description now defines both modes in place and scopes the removal to *the managed worktree*, so it stays correct even for a reused directory; it no longer disagrees with the sentence further down the same string.
- Moving the change note to a new filename to satisfy `rush change --verify` preserves the note verbatim apart from the one reworded sentence — I diffed it; nothing from #380 was dropped.
- `.agents/domains/dispatcher-orchestration.md`, the task record, and `verification.md` all follow, and `verification.md` is honest that the prompt still names `dissolve` once inside the server map — which is also exactly what the new test scrubs before asserting.

Two non-blocking notes inline. Neither should hold the merge.

packages/dreamux/src/service/team-collection/mcp-delegate.ts
ryanxiang7: Confirmed with the operator, and I have now pinned down exactly how much the Dispatcher is missing — it is narrower than my comment implied, and there is a second, smaller defect next to it.

**What the Dispatcher can see.** All three Team projections carry the same single field:

- `TeamView` (`team.status`), `TeamListRow` (`team.list`), `TeamHistoryRow` (`team.history`) — each `worktree_cleanup: AgentEntityWorktreeIdentity['cleanup_state']`, filled from `team.worktree.cleanup_state` in `read-model.ts` and `team-view.ts`.

**What that field settles, and what it does not.** `cleanup_state` is the *lifecycle* value, not the mode:

- `not-managed` → a reused directory. The Dispatcher **can** conclude nothing is removed. ✅
- `managed-active` → `worktree/manager.ts` writes this at prepare time for a managed worktree **regardless of `cleanup`**, so `keep` and `delete-on-close` are indistinguishable here. ❌
- `kept` / `deleted` / `retained-error` / `cleanup-pending` only appear **after** an assessment has run, i.e. at or after dissolve — too late to inform the decision to dissolve.

So for the only state that matters — an open Team the Dispatcher is deciding whether to dissolve — `managed-active` is all it gets, while the dispatcher-facing `dissolve` asks it to reason in terms of `delete-on-close` vs `keep`, and defines `force` entirely in those terms. Before R31 the gap was coverable by inference (`keep` unless the caller passed `cleanup`); the flipped default removed that.

To be explicit about the mechanism: **this should not be prompt injection.** The TeamLeader needed the prompt because its Team catalog is `dissolve` and nothing else — it has no read tool for its own Team. The Dispatcher has `list` / `status` / `history`, so the fact belongs in a read result.

**Second, smaller defect.** `TeamListRow`'s doc comment says the machine-local `worktree_mode` is *"no longer projected — reach for `team.status` for detail"*, but `TeamView` does not carry the mode either. The pointer sends a reader to a surface that does not have what it was sent for.

Suggested shape: project `worktree.cleanup` (and, if `worktree_mode` is meant to be reachable, the mode) on `TeamView` only, leaving `list` and `history` compact — which is what that doc comment already promises. Still out of this PR's scope; happy to open a tracking issue instead if that is easier to schedule.

.agents/tasks/mcp/refine-model-facing-surfaces/requirement.md
YourWildDad: Fixed in 127aca8a: the line break is joined and the inner marks are the operator's own curly ones, with a one-line note saying so.

packages/dreamux/src/service/team-collection/mcp-delegate.ts
YourWildDad: Confirmed against the same three files, and the operator ruled to add the fields (R35 in the task record). In c11a9bdb `team.status` reports `worktree_mode` and `worktree_cleanup_mode` next to `worktree_cleanup`; `list` and `history` stay compact as the `TeamListRow` comment promises, which is now true for `worktree_mode`. The Dispatcher-facing `dissolve` description names the two fields. The output schema is an open object, so nothing else changes; `team-view.test.ts` pins both fields for a managed delete-on-close worktree and a reused directory. Bundled here rather than tracked because the gap was opened by the default flip in the same unreleased batch.

ryanxiang7 (approved):
Approving. Both notes are resolved, and the second one is fixed properly rather than papered over.

**R35 / `team.status`.** `TeamView` now carries `worktree_mode` and `worktree_cleanup_mode` straight off the record's worktree identity, and I verified the pieces that make that safe and sufficient:

- The `team.status` output is `closedObjectSchema({ team: OPEN_OBJECT, ... })`, so the two added fields need no schema change — the claim in your reply checks out.
- `teamView()` has exactly two consumers (`read-model.ts` for `team.status` and `TeamService.view()`), so nothing else picks the fields up; and both values are enums, not machine-local paths, so the `TeamView` "projected away" contract is intact.
- `list` and `history` are untouched, which makes `TeamListRow`'s own comment — *"`worktree_mode` no longer projected — reach for `team.status` for detail"* — true for the first time.
- The Dispatcher-facing `dissolve` now names both fields, so the caller that has to reason in `delete-on-close` / `keep` terms is told where to read them. That closes the gap the R31 default flip opened, and it does it through a read result rather than a prompt, which is the right split: the TeamLeader needs the prompt precisely because its Team catalog is `dissolve` and nothing else.

**R34 quote.** Line break joined, the inner marks identified as the operator's own curly ones, with a one-line note recording both — so the record still reads as verbatim without the transcription artifact.

The task record and `dispatcher-orchestration.md` both carry the reasoning, `verification.md` records the review round honestly (including the stale `TeamListRow` pointer), the change note picks up the new fields, and CI is green on both runners. Nothing further from me.

PR #384, opened 2026-09-06
Replace Claude request windows with resident-session settlement

YourWildDad:
Claude Code can start native turns when background work finishes. Dreamux treated a result without an active request window as a protocol failure and terminated the resident process, also losing its background agents.

This change removes the request window. One resident session accepts every input through the same path. RPC owns one table of unanswered requests; native consumption and matching submitted UUIDs identify the requests answered by each result.

- Background results emit activity and end events without settling unrelated requests or terminating the process. Folded inputs share one immutable completion; queued inputs wait. Core retains its recipient selection and completion-object deduplication.
- Genuine errors remain failures, including UUID-less errors and `subtype: success` with `is_error: true`. Consumed commands retain result membership through `cancelled`, which can follow native failure and does not prove a user stop. Failure cleanup preserves the original exit cause for racing admissions.
- The aggregate completion promise, terminal drainage, runtime request queue and initial/steer split are removed. No origin filter or second request registry is added.

The PR changes 10 source paths in the Claude package: seven modified, one added and two deleted, counting old and new activity-module paths separately. Package source count falls from 29 to 28. Eight test/fixture paths change. Core and the neutral runtime ABI are unchanged.

**Extension compatibility:** custom session factories and direct `ClaudeCodeStreamRpc` consumers replace `submitTurn`/`steerTurn` with `submit` returning `RuntimeAdmission`. RPC consumers replace `failPending` with `fail` or `stop`; options require `sessionId`, optionally accept `outputSchemaEnabled`, and pass an Error to `reapOnTimeout`. Session exit callbacks carry the failure cause. Native callbacks observe results and lifecycle independently of settlement; parsed results expose `terminalReason`. The package README and Rush breaking minor note cover these changes.

**Validation**

- [GitHub CI](https://github.com/excitedjs/dreamux/actions/runs/34102098737) passed on `6b7e8327`: all nine jobs, including Linux and macOS Rush gates.
- Full-workspace Rush build, lint, test and `typecheck:tests` passed. Claude package: 12 test files, 212 cases.
- Failure replays verify that A cannot receive B's later answer. Production runtime/supervisor regressions with controlled child processes distinguish unexpected exit from explicit stop during admission.
- Real Claude Code 2.1.263 against a controlled local 401 API emitted `success/is_error: true/api_error` followed by `cancelled`. Its captured stream replays through the corrected RPC as failed with its actual message.
- Earlier real provider/Core runs on the resident-session replacement passed pure background, folded steer and queued input. These captures also replay through the corrected RPC; they are not new end-to-end runs of this correction.
- Task/knowledge checks, change-file validation and `git diff --check` passed. TeamLeader whole-diff pre-review completed; the operator omitted dynamic-workflow review.

**Evidence limits:** command lifecycle is an internal CLI seam. Missing-start sequences remain compatibility coverage. Setup failure without consumption or UUID evidence reports detailed native failure, but the subsequent named cancelled request receives a generic failure; the adapter does not guess its error owner. The complete wrong-answer incident remains a synthetic reproduction. Exceptional tool ordering and external cancellation have no new live coverage. The existing outstanding-request idle policy is unchanged.


packages/agent-runtime/claude-code/src/rpc.ts
```diff
@@ -440,25 +442,8 @@ export class ClaudeCodeStreamRpc {
...
-            `claude result envelope for unsubmitted command ${commandUuid}; ` +
-              'native completion ownership is ambiguous',
-          );
-          this.options.log?.('error', error.message, error);
-          this.settlePending(error)?.reject(error);
-          this.options.reapOnTimeout();
-          break;
-        }
+        this.aggregator.accept(line);
+        const outcome = this.aggregator.takeOutcome()!;
```
ryanxiang7, line 223: **The interrupt artifact now consumes and wipes the aggregator it is documented not to touch.**

`accept()` + `takeOutcome()` were moved *above* the artifact guard. `takeOutcome()` calls `discard()`, which clears `lastAssistantText` and `result` — and then line 454 throws the outcome away and `break`s. Pre-diff the guard ran before `accept`, so the artifact never entered the aggregator.

This matters because the aggregator is now session-wide, and `outcome()` falls back to `lastAssistantText` when `r.text` is empty:

1. Command A streams assistant text -> `lastAssistantText = 'partial answer'`.
2. An `error_during_execution` artifact (no `result`, no `user_message_uuid`) arrives -> accepted, `takeOutcome()` discards, `lastAssistantText = ''`, outcome dropped.
3. A's real `result` lands with an empty `result` field -> `outcome.text` is `''` instead of `'partial answer'`, and the caller gets `resultText: null`.

The existing regression test (`rpc.test.ts`, *"does not forward the error_during_execution artifact..."*) uses `resultLine('interrupting answer', steerUuid)` — a result **with** text — so it never exercises the fallback and passes either way.

Moving the guard back above `accept` restores the documented "not a native answer boundary" behaviour.

packages/agent-runtime/claude-code/src/rpc.ts
```diff
@@ -406,21 +406,23 @@ export class ClaudeCodeStreamRpc {
...
+          state: line.state,
+        });
         // `command_lifecycle` does double duty: it coordinates live-steer
         // admission (writeWaiters) and proves lifecycle capability, and its
         // terminal states are the drainage gate when the CLI represents several
         // started commands with one `result`.
         const newlySupported = this.lifecycleSupported === null;
-        if (newlySupported) this.lifecycleSupported = true;
+        this.lifecycleSupported = true;
+        if (line.state === 'cancelled') this.aggregator.discard();
```
ryanxiang7: **`discard()` on `cancelled` is unscoped, on an aggregator that is no longer per-turn.**

This line runs before `if (pending === null) break`, and without asking whether `line.commandUuid` belongs to this window. The aggregator moved from `PendingTurn` to the RPC instance in this PR, so its `lastAssistantText` is now shared by every native turn in the resident session — while the discard trigger stayed per-command.

Two reachable consequences, both newly possible *because* background turns now coexist with submitted ones:

- A and B are folded into one window, both `started`; the model streams `'running answer'`; an interrupt cancels A. B survives, its shared `result` carries an empty `result` field, and B's completion is `resultText: null` instead of `'running answer'`.
- A cancelled **background** command wipes the fallback text of an unrelated in-flight submitted window.

Same root cause as my note on line 446: hoisting `TurnAggregator` to session scope without re-scoping who may discard it. Gating on `pending?.submitted.includes(line.commandUuid)` would confine it to the window that owns the text.

packages/agent-runtime/claude-code/src/rpc.ts
```diff
@@ -469,15 +454,30 @@ export class ClaudeCodeStreamRpc {
...
+            commandUuids.add(commandUuid);
+          }
+          // Without lifecycle, admission allows only the initial input. Its
+          // legacy UUID-less result is attributable, but an explicit foreign
+          // UUID is not. Lifecycle-capable background results stay unassigned.
+          if (this.lifecycleSupported !== true && commandUuid === null) {
+            commandUuids.add(pending.submitted[0]!);
+          }
+          pending.startedSinceResult.clear();
+          if (commandUuids.size > 0) pending.hasAttributedResult = true;
```
ryanxiang7: **The single-submission attribution fallback was deleted, and the replacement only covers lifecycle-less builds — a lifecycle-capable result with no matching uuid now hangs the window.**

`sawResult` is set only when `commandUuids` is non-empty, and the only uuid-less rescue is line 469, gated on `lifecycleSupported !== true`. The old net was in `completeStartedGroup`:

```ts
if (representative === undefined && active.submissions.size === 1) { ... }
```

removed here, together with its test — whose own comment states the case is real:

> `attributes an unstarted result to the sole pending submission of a single-command turn`
> *"Some Claude Code builds omit `started` for a fast turn; a turn with exactly one submission still has an unambiguous owner."*

Walk it with `init` advertising `msg_lifecycle_v1`, a `result` with no `user_message_uuid`, and `started` omitted (`completed` still arrives):

- `commandUuids` is empty -> `completeSubmittedGroup` returns at line 137 without settling; `sawResult` stays `false`.
- `settleIfReady`: first escape needs `lifecycleSupported !== true` — no. `startedSinceResult` is empty. Every submitted command is terminal. `sawResult` is `false`. `ranAnyCommand` is `true` (from `completed`), so it `return`s.

The window never drains, `RuntimeSubmission.settled` never resolves, and the only exit is the idle deadline — which this file's own comment calls *"not an acceptable backstop here: it reaps the resident child"*. Either keep a sole-submission fallback for the lifecycle-capable path too, or state why builds that advertise lifecycle are guaranteed to emit `started`.

packages/agent-runtime/claude-code/src/rpc.ts
```diff
@@ -406,21 +406,23 @@ export class ClaudeCodeStreamRpc {
...
+          commandUuid: line.commandUuid,
+          state: line.state,
+        });
         // `command_lifecycle` does double duty: it coordinates live-steer
         // admission (writeWaiters) and proves lifecycle capability, and its
         // terminal states are the drainage gate when the CLI represents several
         // started commands with one `result`.
         const newlySupported = this.lifecycleSupported === null;
-        if (newlySupported) this.lifecycleSupported = true;
+        this.lifecycleSupported = true;
```
ryanxiang7: **`lifecycleSupported` is tri-state, and both new uses treat it as a boolean.**

Two separate slips:

**1. This line drops the old `if (newlySupported)` guard.** Pre-diff it was `if (newlySupported) this.lifecycleSupported = true;`, i.e. promote only from `null`. Now a single `command_lifecycle` frame flips a session that `init` already decided is lifecycle-less from `false` to `true`. That disables both anti-hang paths for the current window at once — `settleIfReady`'s first escape (`sawResult && lifecycleSupported !== true`) and the uuid-less attribution at line 469 — while the build still emits no lifecycle for submitted commands, so `terminal` never fills. `newlySupported` is also `false` on that transition, so capability waiters are not flushed either.

**2. Line 469 tests `!== true`, which folds `null` into `false`.** `decideLifecycleSupport` leaves the flag `null` until `init` is parsed, so before `init` a uuid-less result is attributed to `pending.submitted[0]` and immediately settles the window through the first escape — even though nothing was decided and the result may not be that command's. The comment above it says *"Without lifecycle, admission allows only the initial input"*, which describes the decided-`false` state, not the undecided one.

Both read cleanly if the two sites say what they mean: `if (this.lifecycleSupported === null) this.lifecycleSupported = true;` here, and `this.lifecycleSupported === false` at 469.

ryanxiang7 (commented):
Read the five source files in full and walked the state machine rather than the diff; the `.agents` records and the two COT task notes were read for intent. CI is green.

The core idea is right, and the two things it set out to fix are fixed: an unbound `result` no longer reaps the resident child, and a steer that joins a background turn shares that turn's completion. What I would want looked at before merge is mostly one theme — **`TurnAggregator` was promoted from per-`PendingTurn` to a session-wide singleton, but the things that consume or discard it are still scoped to a single command**:

- line 446: the interrupt artifact now `accept`s + `takeOutcome`s (i.e. `discard`s) before the guard that says it is not a boundary;
- line 421: `cancelled` discards the shared aggregator before the `pending === null` check and without checking ownership.

Alongside that, three independent ones: a deleted attribution fallback that can leave a window permanently undrained (line 473), `lifecycleSupported` being tri-state while both new call sites treat it as a boolean (line 420), and an unattributed background `turn.ended` closing the operator's open Feishu card (`runtime-submissions.ts:114`). Then two smaller ones and a nit.

Two things I checked and am **not** raising:

- **The `BREAKING:` change note has neither `Review:` nor `Rebuild:`, and I think that is correct here.** CLAUDE.md scopes those clauses to incompatible config / state / path shape changes; this is a pure compile-time break in an extension seam. The repo has direct precedent for exactly this form — `packages/channel/feishu-channel/CHANGELOG.md:69` (*"BREAKING: Remove the root-exported createFakeFeishuBot test helper... runtime behavior and persisted state are unchanged"*). The note already carries the analogous non-impact sentence.
- **The double-`turn.ended` risk in `runtime.ts:526`** (an admitted-but-unbound turn can publish an end here and a second one from `markTurnFailed`). The COT adapter documents this away: *"with no card open there is nothing to finish... That also makes a repeated end harmless."* Benign by design.

packages/agent-runtime/claude-code/src/runtime-submissions.ts
```diff
@@ -128,47 +116,25 @@ export function handleProtocolEvent(
       event.outcome.isError ? turnFailureMessage(event.outcome) : null,
       context.activitySink,
     );
-    completeStartedGroup(active, event.outcome, context);
+    context.activity.tools.clear();
+    if (active !== null) completeSubmittedGroup(active, event.commandUuids, event.outcome, context);
```
ryanxiang7: **An unattributed background result now publishes `turn.ended`, which closes whichever card the operator currently has open.**

Settlement is gated on `active !== null` here, but `endNativeTurn` five lines above is not — so it also runs for a result with `commandUuids: []` — the case `runtime-background.test.ts` asserts (`'assistant.message', 'tool.call', 'tool.call', 'assistant.message', 'turn.ended'`).

`turn.ended` is not advisory downstream. `feishu-cot-adapter.ts`:

> *"This is the only terminal a card has. It names no logical turn — a provider folded whatever it folded — so it closes whatever this recipient currently has open, which is exactly the one card the operator is watching."*

The harmful window is a **queued** request: the Channel opens its card at submission, before Core runs; the request then waits behind the running native turn, so it is not in `startedSinceResult`; a background result lands, `commandUuids` is empty, and the card closes as `completed` while the request has not started. Its later output then opens a second card.

Pre-diff this could not happen — `onProtocolEvent` returned early on a null `activeTurn`, and the RPC failed/reaped the unbound result instead.

Note the docs updated here (`feishu-cot-conversation-cards/README.md` and `continued-optimization.md`) supersede the **fail-loud / reap** behaviour, which is right, but neither mentions that unbound results now emit a card terminal. If that is intended, the COT requirement is where it needs saying; if not, the end belongs inside the `active !== null` branch (or gated on non-empty `commandUuids`).

packages/agent-runtime/claude-code/src/runtime-activity.ts
```diff
@@ -128,47 +116,25 @@ export function handleProtocolEvent(
       event.outcome.isError ? turnFailureMessage(event.outcome) : null,
       context.activitySink,
     );
-    completeStartedGroup(active, event.outcome, context);
+    context.activity.tools.clear();
```
ryanxiang7, line 73: **`tools.clear()` at every result boundary, but a request window legally spans several results.**

The map this replaced (`ActiveTurn.tools`) lived for the whole window and was never cleared. The `ActiveTurn` doc comment right above still says so:

> *"A window is opened by an initial command and can legally produce several sequential `result` boundaries — one native turn each — as commands steered or queued into it run after the earlier ones were answered."*

So: a `tool_use` registers `call-1` in `activity.tools`; a result boundary lands — an earlier command's, or a background one, which is exactly the interleaving this PR introduces — and clears the map; `call-1`'s `tool_result` arrives afterwards. `toolResultActivity` finds nothing and emits a `tool.call` with `toolName: 'tool'` and `arguments: null`: an unnamed tool row in the operator's card.

Clearing per result is the natural instinct for a per-turn map, but this map is correlation state whose lifetime is the tool call, not the turn.

packages/agent-runtime/claude-code/src/rpc.ts
```diff
@@ -406,21 +406,23 @@ export class ClaudeCodeStreamRpc {
...
-        if (pending === null || line.commandUuid === null) break;
-        if (line.state !== null) {
-          this.options.onProtocolEvent?.({
-            kind: 'command_lifecycle',
-            commandUuid: line.commandUuid,
-            state: line.state,
-          });
-        }
-        if (line.state === 'started') pending.startedSinceResult.add(line.commandUuid);
+        if (line.commandUuid === null || line.state === null) break;
```
ryanxiang7: **This guard now drops an unmodelled lifecycle state before capability proof and write-waiter resolution.**

Pre-diff only the `onProtocolEvent` forward was gated on a non-null state; `lifecycleSupported = true`, `resolveWriteWaiter`, and the capability flush all still ran. Now a frame whose `state` the parser does not model (`parseCommandLifecycle` maps unknown states to `null`) breaks out before any of them.

`stream.ts`'s module contract is explicit that this is the direction not to regress in — *"a wider or newer CLI build cannot break a turn"*. Concretely: a newer CLI emits `{"type":"command_lifecycle","command_uuid":"<steer>","state":"running"}` and its `init` omits the capability string. Lifecycle is never proved, the write waiter for that steer is never resolved, and live steering stays rejected for the life of the session.

Keeping the two effects that do not depend on the state value above this guard would preserve the old tolerance.

packages/agent-runtime/claude-code/src/runtime-activity.ts
```diff
@@ -99,25 +101,11 @@ export function createRuntimeSubmission(): SubmissionDeferred {
...
-    if (
-      active.submissions.has(event.commandUuid) &&
-      !active.completedCommands.has(event.commandUuid) &&
-      !active.started.includes(event.commandUuid)
-    ) {
-      active.started.push(event.commandUuid);
-    }
-    return;
-  }
+  if (event.kind === 'command_lifecycle') return;
```
ryanxiang7, line 57: Nit / entropy: the `command_lifecycle` member of `ClaudeProtocolEvent` is now dead weight. The RPC still builds and forwards one object per lifecycle frame (`rpc.ts:410-414`), and this — its only production consumer — discards it on the first line. Nothing else in `src/` reads the variant.

Since this release is already breaking the `onProtocolEvent` callback shape by adding `commandUuids`, it is the cheap moment to either drop the variant or say in the type why an external `ClaudeCodeSessionFactory` still needs to receive it.

packages/agent-runtime/claude-code/src/rpc.ts
```diff
@@ -42,442 +11,256 @@ import {
...
+        // lifecycle settles the aborted requests; this is not a completion.
         if (
           line.outcome.subtype === 'error_during_execution' &&
-          line.outcome.userMessageUuid === null &&
-          line.outcome.text === null
-        ) {
-          this.options.log?.('warn', 'claude interrupt result artifact ignored');
-          break;
+          line.outcome.userMessageUuid === null && line.outcome.text === null
+        ) break;
```
ryanxiang7: **[F1+F2 — BLOCKER] The artifact guard is too broad and strands the settlement set.**

A `result` is a native turn boundary regardless of subtype. This three-field match (`error_during_execution` + `userMessageUuid === null` + `text === null`) is not specific to the 2.1.231 interrupt artifact: the same zeroed shape is produced by a crashed worker (per the KB's own protocol note) and by hard failures such as `turn_setup_failed`/`model_error`. Swallowing those results means (a) the caller never gets a `failed` settlement, and (b) `consumed` is never cleared, so the *next* result mis-settles the stale started requests — A's caller receives B's answer.

Fix: treat every result as a turn boundary that clears `consumed`; distinguish the true abort artifact (the CLI's own abort predicate is `terminal_reason ∈ {aborted_streaming, aborted_tools}` per the 2.1.251 schema) from real error results, which should settle their consumed requests as `failed` with the `errors` text and emit a failed native end. Restore the warn log that the old code emitted here.

packages/agent-runtime/claude-code/src/rpc.ts
```diff
@@ -42,442 +11,256 @@ import {
...
-          this.flushCapabilityWaiters(pending);
+        const { commandUuid, state } = line;
+        if (commandUuid === null || state === null) break;
+        if (state === 'started') this.consumed.add(commandUuid);
+        const request = this.requests.get(commandUuid);
+        if (request?.write === null) {
+          this.acceptRequest(request);
+          if (state === 'cancelled' || state === 'discarded' || state === 'refused') {
+            this.requests.delete(commandUuid);
+            request.settle(state === 'cancelled'
```
ryanxiang7: **[F4 — MAJOR] `cancelled` is overloaded: a hard-failure cancel is settled as a neutral `stopped`, not `failed`.**

The CLI emits `command_lifecycle: cancelled` for two very different situations: (a) a Remote-Control `priority:'now'` interrupt of a queued/started command, and (b) a command the CLI could not run and cancels as a failure. Here both map to `{ kind: 'stopped' }` (and at line 216 `cancelled` is also treated as a turn boundary that discards the aggregator). The old code let the runtime classify stopped-vs-failed via `classifySteerFailure`; this rewrite hard-codes the optimistic reading.

Combined with F1/F2, a hard-failure `cancelled` both strands `consumed` (the result that would have cleared it is swallowed by the artifact guard) and reports the caller as cleanly stopped. Please distinguish the two — at minimum, a `cancelled` that arrives without a preceding `started` for a written request is a failure, not a stop.

packages/agent-runtime/claude-code/src/supervisor.ts
```diff
@@ -121,42 +122,16 @@ class LiveClaudeCodeSession implements ClaudeCodeSession {
...
-    if (this.stopRequested || this.stopped) {
-      return Promise.reject(
-        new ClaudeSteerAdmissionError(
-          'failed',
-          'claude resident session is stopped before live steer',
-        ),
-      );
-    }
+  ): Promise<RuntimeAdmission> {
+    if (this.stopRequested || this.stopped) return Promise.resolve({ status: 'stopped' });
```
ryanxiang7, line 131: **[A1 — MAJOR] The child now authors the neutral `stopped` admission, and core can no longer tell a stop from a failure.**

When `stopRequested`/`stopped` is set, `submit()` returns `{ status: 'stopped' }` directly. But `stopped` is a *neutral* admission in `RuntimeAdmission` — it means "the runtime was asked to stop," not "this submit failed." The old code rejected with `ClaudeSteerAdmissionError('failed')` and let `classifySteerFailure(error, this.stopped)` decide stopped-vs-failed at the runtime layer.

Concretely: `turn-recording.ts:208` maps `stopped` → `{ status: 'unsupported', reason: 'runtime stopped' }`, so a submit that races a stop is now recorded as "unsupported" rather than failed, and the caller cannot distinguish "I was stopped" from "the session is broken." The stop-vs-failure decision belongs behind the provider, not hard-coded in the session. Please route this through the same classification the old code used (or justify why `stopped` is always the correct neutral reading here).

common/changes/@excitedjs/agent-runtime-claude-code/fix-claude-background-turns_2026-09-07-06-30.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "BREAKING: Custom ClaudeCodeSessionFactory implementations must replace submitTurn/steerTurn with submit returning RuntimeAdmission and per-request settlement. Session specs now provide sessionId and optional outputSchemaEnabled; exit handlers receive an Error, and protocol callbacks observe results and interrupted boundaries rather than performing settlement. Remove request-window coordination, retain native background work, and settle folded or queued inputs directly. Core and neutral runtime contracts are unchanged.",
```
ryanxiang7: **[T3 — MINOR] The change note covers `ClaudeCodeSessionFactory` but misses the exported `ClaudeCodeStreamRpc` break.**

`ClaudeCodeStreamRpc` is part of this package's public API (re-exported from `src/index.ts:32`), and this PR renames its public methods `submitTurn`/`steerTurn` → `submit` and changes `ClaudeCodeStreamRpcOptions` (`reapOnTimeout` now takes an `Error`; `sessionId`/`outputSchemaEnabled` added). The note only calls out the factory contract. Any consumer importing `ClaudeCodeStreamRpc` directly breaks on upgrade with no `BREAKING:` line to tell them why. Please add the `ClaudeCodeStreamRpc` surface to this note (it can stay in the same change entry).

packages/agent-runtime/claude-code/src/runtime-activity.ts
ryanxiang7: Rejecting this one. In the old window model `tools` had to survive across folded results because one turn could produce several, and `startedSinceResult` tracked that. Under per-request settlement each `result` is a turn boundary and the tools it carried belong to the turn that just ended — clearing is correct, and keeping them would leak tool state into the next turn. The window entity that made this a concern no longer exists.

packages/agent-runtime/claude-code/src/runtime-activity.ts
ryanxiang7: Agreed it is under-documented, but I would keep it: `command_lifecycle` is still emitted on the protocol callback (rpc.ts:221) as an observation seam for custom `ClaudeCodeSessionFactory` implementations that track command state directly. The activity projection not consuming it is fine — it is a display layer, and lifecycle is settlement-internal. Non-blocking; a one-line comment on the `ClaudeProtocolEvent` type ("emitted for factory observation; not consumed by the built-in activity projection") would close it.

packages/agent-runtime/claude-code/src/rpc.ts
ryanxiang7: Agreed the behavior itself is correct: the artifact arrives after discard() on cancel (line 217), so the aggregator is already empty and takeOutcome() on the empty artifact is a no-op wipe. But the silent break needs the warn log the old code emitted here, plus a comment explaining why this is not a settlement boundary. Folded into the F1/F2 blocker on line 233 — the deeper issue is that this three-field match cannot distinguish an interrupt artifact from a crashed-worker zeroed result (they are byte-identical on the wire), so when it is actually the latter it strands consumed and the next result mis-settles.

ryanxiang7 (changes requested):
## Verdict: REQUEST_CHANGES

The settlement model is a genuine entropy reduction — 653-line `rpc.ts` + 376-line `runtime-submissions.ts` collapse to 361 + 195, the diff is confined to `packages/agent-runtime/claude-code/` (+ `.agents/`, change file), and the neutral ABI (`RuntimeAdmission` / `RuntimeSubmission` / `RuntimeCompletion`) is untouched. The happy paths (folded commands, queued requests, pure-background results) are sound and live-probed. But the error path has a blocker that can deliver the wrong answer to the wrong caller.

### Blocker

**F1+F2 — the `error_during_execution` artifact guard is too broad and strands the settlement set** ([rpc.ts:233](https://github.com/excitedjs/dreamux/pull/384#discussion_r3947378124)).
The three-field match (`error_during_execution` + `userMessageUuid === null` + `text === null`) cannot distinguish the 2.1.231 interrupt artifact from a crashed-worker zeroed result — they are byte-identical on the wire, and the KB's own protocol note (`.agents/research/claude-code-stream-json-protocol.md:170`) documents that a crashed worker produces exactly this zeroed shape. When it is the latter, `consumed` is never cleared, so the *next* result mis-settles the stale started requests: A's caller receives B's answer. A `result` is a native turn boundary regardless of subtype. Fix: clear `consumed` on every result, and use the CLI's own abort predicate (`terminal_reason ∈ {aborted_streaming, aborted_tools}` per the 2.1.251 schema) to distinguish the true interrupt artifact from real error results — the latter should settle their consumed requests as `failed` with the `errors` text and emit a failed native end. Restore the warn log the old code emitted here.

### Major

**F4 — `cancelled` is overloaded** ([rpc.ts:208](https://github.com/excitedjs/dreamux/pull/384#discussion_r3947385773)).
`command_lifecycle: cancelled` covers both a Remote-Control `priority:'now'` interrupt and a command the CLI could not run and cancels as a failure. Both map to `{ kind: 'stopped' }` (and line 216 also treats `cancelled` as a turn boundary that discards the aggregator). The old code let the runtime classify stopped-vs-failed; this rewrite hard-codes the optimistic reading. A `cancelled` that arrives without a preceding `started` for a written request is a failure, not a stop.

**A1 — the child now authors the neutral `stopped` admission** ([supervisor.ts:130](https://github.com/excitedjs/dreamux/pull/384#discussion_r3947386826)).
When `stopRequested`/`stopped` is set, `submit()` returns `{ status: 'stopped' }` directly. But `stopped` is a *neutral* admission — it means "the runtime was asked to stop," not "this submit failed." `turn-recording.ts:208` maps `stopped` → `{ status: 'unsupported', reason: 'runtime stopped' }`, so a submit that races a stop is now recorded as "unsupported" rather than failed, and the caller cannot distinguish "I was stopped" from "the session is broken." The old code rejected with `ClaudeSteerAdmissionError('failed')` and let `classifySteerFailure` decide at the runtime layer. The stop-vs-failure decision belongs behind the provider.

### Required before merge

**F3 — held-result ordering during background Tasks needs a live probe.** The verification doc admits "Cancellation/Remote Control interleaving and exceptional tool-result ordering have no fresh live probe." The 2.1.251 binary reading suggests a background Task can hold a result while a foreground turn streams; if the held result arrives after the foreground result, the settlement order is undefined. This is the exact interleaving the new model is most fragile under, and it is untested. Please add a live probe (or a fake-stream test that exercises the held-then-released ordering) before merge.

### Minor

**T3 — the change note misses the exported `ClaudeCodeStreamRpc` break** ([change file:4](https://github.com/excitedjs/dreamux/pull/384#discussion_r3947387593)).
`ClaudeCodeStreamRpc` is re-exported from `src/index.ts:32`, and this PR renames its public methods (`submitTurn`/`steerTurn` → `submit`) and changes `ClaudeCodeStreamRpcOptions` (`reapOnTimeout` now takes an `Error`). The note only calls out `ClaudeCodeSessionFactory`. Add the `ClaudeCodeStreamRpc` surface to the same entry.

### Considered and rejected

- **Test coverage regression (T1).** I checked: the new `rpc.test.ts` has 59 tests (vs 32) with a dedicated cancellation suite (`cancels only the queued request`, `discards cancelled native text with interrupt artifact before/after/absent`, `retains other consumed fold members`, `discards cancelled internal native text`), and `error_during_execution` *with* `errors` text is settled as `failed` (line 527). The dropped stderr assertion in `session.test.ts` was replaced by a stronger callback-based RC assertion. No regression.
- **Operator's `runtime-activity.ts:73` (`tools.clear()`).** Rejecting — under per-request settlement each `result` is a turn boundary and the tools it carried belong to the turn that just ended; the window entity that made this a concern no longer exists.

### Non-blocking nit

- **Operator's `runtime-activity.ts:57` (`command_lifecycle` dead weight).** Agreed it is under-documented, but the event is still emitted on the protocol callback as an observation seam for custom `ClaudeCodeSessionFactory` implementations. A one-line comment on `ClaudeProtocolEvent` ("emitted for factory observation; not consumed by the built-in activity projection") would close it.

### Summary

The architecture direction is right and the happy paths are solid. The blocker is a correctness issue on the error path (wrong-answer delivery), and F4/A1 are settlement-classification regressions from the old code. Once F1+F2 is fixed (clear `consumed` on every result, distinguish via `terminal_reason`, restore the warn log), F4 and A1 are routed through proper classification, and F3 has a live probe, this is approvable.

ryanxiang7 (approved):
## Verdict: APPROVE

Re-reviewed the fix commits (`72af66e58a`..`6b7e83279e`) against the round-1 findings. All accepted findings are correctly resolved, and the F4 fix is better than the heuristic I proposed.

### Round-1 findings

**F1+F2 (blocker) — fixed.** The `error_during_execution` artifact guard is deleted; every `result` is now a turn boundary that clears `consumed`. The repeated wrong-answer replay confirms A now fails with its native error and B receives only B's answer. The `isError` parser fix (`subtype !== 'success' || o['is_error'] === true`) correctly handles the live-captured API-failure shape (`subtype: success, is_error: true, terminal_reason: api_error`), and `turnFailureMessage` now preserves `terminalReason` and success-arm error text.

**F4 (major) — fixed, with a better approach.** Instead of my proposed no-start heuristic, consumed commands that report `cancelled` wait for their native result (which carries the failure evidence via `isError`/`terminalReason`), and only unconsumed `cancelled`/`discarded`/`refused` commands settle directly as `failed`. The fresh 2.1.263 401 probe (`queued, started, success/is_error:true/api_error result, cancelled`) validates that the result — not the lifecycle — carries the failure. Tests cover both cancelled-before-result and result-before-cancelled orderings, folded commands, setup failure, and unbound internal failures.

**A1 (major) — fixed.** `exitError → failed` now takes precedence over `stopRequested → stopped` in `submit()`, and `onChildExit` retains the failure cause through cleanup. The `it.each(['exit','stop'])` session test covers the exact narrowed scenario: unexpected exit during identity publication → `failed`; deliberate stop → `stopped`.

**T3 (minor) — fixed.** The change entry now documents the full `ClaudeCodeStreamRpc` public surface: renamed submission methods, `fail(error)`/`stop()`, `reapOnTimeout` receiving an `Error`, required `sessionId`, and `terminalReason` exposure.

### F3 — rejected with evidence, accepted

The held-result inversion I flagged from binary reading is not established: the installed 2.1.263 engine holds results only after input closes and drains older held results before the current one. No speculative ordering machinery was added, and the evidence gap is documented honestly in `verification.md`.

### Non-blocking note

The `interrupted` protocol event is no longer emitted by the built-in RPC (the type is retained as a public seam for custom factories). This is intentional per ruling R5 — the Claude provider has no user-cancellation entry point, so `cancelled` is always a native failure and the native turn correctly ends as `failed`. The activity projection's `interrupted` branch is defensive code for the seam; a one-line comment there would help future readers.

### Validation

- CI: 9/9 green on `6b7e83279e`.
- 12 test files, 212 cases; failure-path tests use realistic error envelopes and do not bypass the guard.
- `.agents/` delta is complete (`pr-review-fixes.md`, `verification.md` closeout, `rulings.md` R4/R5).
- Diff remains confined to `packages/agent-runtime/claude-code/` + `.agents/` + change file; neutral ABI untouched.

The settlement model is a genuine entropy reduction and the error path is now correct. Approving.

Issue #383, opened 2026-09-06
Solution review: refine Feishu COT and route notification cards


YourWildDad:
# Final technical solution

## Outcome and boundary

Keep this change inside the Feishu Channel while consuming the existing neutral
`team.status` command for canonical Team facts. Core already owns and returns
the TeamLeader name, configured Agent Runtime ID, and runtime cwd, so this task
adds no Core contract, provider interface, persisted record, routing schema, or
provisioning state.

The implementation changes five user-visible behaviors:

1. Plain-text COT tool results keep at most ten content lines before the
   existing Feishu byte fitting runs. JSON objects and arrays bypass this line
   rule.
2. Replacing a non-null COT anchor closes the superseded card as completed.
   Retiring an anchor and all genuine lifecycle interruptions remain
   interrupted.
3. Manual and provisioned route binds obtain the same canonical Team card facts
   before the route becomes visible.
4. Route-bound notifications render the selected Team-focused Card 2.0 design,
   with the runtime cwd in the selected full-width detail-panel treatment.
5. Explicit MCP unbind, confirmed Team dissolution, and pre-final route removal
   each use copy that states only the cause the Channel actually knows.

Collaboration-space policy cards and the general Card 1.0 status-card helper
remain unchanged.

## 1. Bound plain-text tool results after JSON classification

Extend `toolResultOutput` in
`/packages/channel/feishu-channel/src/feishu-cot-events.ts`, where the complete
result is already parsed and classified. If the parsed value is a non-null
object or array, pretty-print it as JSON exactly as today and do not count its
lines. Treat JSON scalars and all non-JSON values as plain text.

For plain text, scan the original string by LF and CRLF source boundaries. A
terminal delimiter terminates the preceding content line but does not create an
additional line. A delimiter followed by content, including another blank-line
delimiter, does start another content line. When an eleventh content line
exists, keep the first ten complete lines, preserve their original separators,
and append the existing English `TRUNCATION_MARKER` on a new line. Apply
`preserveSpacing` only after this source-text cut.

The existing `assembleToolResultContent` byte-fitting pipeline remains the
final bound for JSON and text. This adds no Core processing, configuration, or
second byte budget.

## 2. Give anchor replacement its own terminal

Change `FeishuCotAdapter.advanceAnchor` so its terminal follows the transition:

- old anchor to a new non-null anchor: `completed`;
- old anchor to `null`: `interrupted`.

The adapter already owns both the old presentation and the replacement target,
[...]

PR #386, opened 2026-09-07
Refine Feishu COT and route cards

YourWildDad:
## Summary

- cap non-JSON Feishu COT tool results at ten content lines while preserving JSON object/array output and the existing byte limit
- complete superseded COT cards successfully and keep genuine lifecycle interruptions unchanged
- restore TeamLeader runtime ID and cwd on route-bound cards, using the enriched team.create receipt for provisioning
- distinguish explicit unbind, final Team dissolution, and pre-final route-ended notifications with the selected English Card 2.0 designs

## Architecture

- manual bind performs one canonical team.status read; TEAM_NOT_FOUND propagates unchanged and closed is read from the successful response
- automatic provisioning consumes leader_agent_runtime and runtime_cwd directly from team.create and performs no redundant status read
- no persisted routing/config/state change, retry, polling, or compensation mechanism

## Verification

- PASS: affected builds for @excitedjs/feishu-channel and @excitedjs/dreamux
- PASS: @excitedjs/dreamux-types (106 tests)
- PASS: @excitedjs/feishu-channel (416 tests)
- PASS: 977 deterministic @excitedjs/dreamux tests
- LIVE ENVIRONMENT: 5 Codex live tests timed out or observed no mid-turn activity; no failures were on the changed contract paths
- PASS: rush change --verify --target-branch origin/next --no-fetch
- PASS: .agents/scripts/check.sh
- PASS: git diff --check and pre-commit gitleaks/internal-content gates

## Follow-up on this draft

- affected lint and test typechecking
- deterministic Dreamux rerun with live Codex skipped only for environment isolation
- independent implementation review and full repository gates

Refs #383

### DIFF
--- a/.agents/domains/channel.md
+++ b/.agents/domains/channel.md
@@ -482,8 +482,9 @@ one, and send the notification card.
 
 Refusing to route to a Team that cannot answer is the one moment the question can
 be asked: `team.status` is invoked before the row is written and before the
-conversation is told, and only a definite answer refuses — `TEAM_NOT_FOUND`,
-`TEAM_CLOSED`, or a `closed` status. A Team that dissolves afterwards converges
+conversation is told. A nonexistent Team throws Core's `TEAM_NOT_FOUND`
+unchanged; a closed Team is a successful status response whose `team.status` is
+`closed`, which the Channel rejects. A Team that dissolves afterwards converges
 through the `team.state` event instead.
 
 What Core owns is the lease scope. Every Channel MCP call carries a
@@ -529,10 +530,13 @@ provisioning begins on first accepted topic inbound.
 The run order is the one that degrades honestly — create the Team, commit the
 route, announce it, deliver the message — and every exit before `team.submit`
 answers `unsubmitted`, so the message it was carrying goes to the Dispatcher
-Agent like any other message this Channel could not hand to a Team. Failing to
-provision is not a reason to drop what somebody wrote. A failure after
-`team.create` leaves an ordinary Team nothing routes to: an accepted orphan an
-operator can see and use, deliberately not compensated.
+Agent like any other message this Channel could not hand to a Team. The
+`team.create` receipt carries the created Team's leader name, configured runtime
+ID, and runtime cwd, so the Channel can render the route card without an
+immediate `team.status` round trip. Failing to provision is not a reason to drop
+what somebody wrote. A failure after `team.create` leaves an ordinary Team
+nothing routes to: an accepted orphan an operator can see and use, deliberately
+not compensated.
 
 Idempotency comes from one choice with several consequences: the `team.create`
 `request_id` is the inbound message id, used bare.
@@ -583,23 +587,30 @@ Routing notification cards are rendered from this Channel's own records, at the
 moment this Channel changes them, and no Core event is involved — Core publishes
 no binding fact, because it holds none.
 
-There are four renderers, and their field sets are the content contract:
+There are six renderers, and their field sets are the content contract:
 
 - **route bound** (manual bind, or an automatic provisioning announcement):
-  target display, binding kind (topic or group), Team name, and the space name
-  when the row was installed for a Space;
-- **route unbound** (manual unbind, and every route a closing Team gives up):
-  target display, Team name, and a status line;
+  target display, binding kind (topic or group), Team name, TeamLeader name,
+  configured Agent Runtime ID, and runtime cwd;
+- **route unbound** (manual Channel MCP unbind): target display, binding kind,
+  Team name, and an explicit statement that only the route was removed and the
+  Team remains active;
+- **Team dissolved** (routes removed after final `team.state(status: 'closed')`):
+  target display, binding kind, Team name, and the confirmed Team-closure cause;
+- **route ended** (early route removal after an inbound `TEAM_CLOSED` rejection):
+  target display, binding kind, Team name, and cause-neutral wording because
+  dissolution may still be in progress and may be refused;
 - **space bound**: space name, container group display, TeamLeader runtime, the
   workspace — which renders the configured repository path when the policy has
   one — base ref, and a note that new topics get their own Team;
 - **space unbound**: space name, container group display, and a status line
   saying provisioning stopped while existing Teams and bindings are unchanged.
 
-Cards use Feishu `plain_text` elements only and render display fields, Team
-facts, and the Space's own policy without rendering prompts or raw errors. The
-repository path is deliberately visible to the members of the bound conversation;
-the user-visible half of that disclosure is
+The three route lifecycle cards use Card 2.0 with English fixed copy and literal
+`plain_text` dynamic values. Collaboration Space cards keep their existing
+schema. Cards render display fields, Team facts, and the Space's own policy
+without rendering prompts or raw errors. The repository path is deliberately
+visible to the members of the bound conversation; the user-visible half of that disclosure is
 [`/.agents/product/README.md`](/.agents/product/README.md).
 
 Where a card goes follows the target. A route card for a topic replies under the
--- a/.agents/domains/channel.md
+++ b/.agents/domains/channel.md
@@ -606,11 +606,12 @@ There are six renderers, and their field sets are the content contract:
 - **space unbound**: space name, container group display, and a status line
   saying provisioning stopped while existing Teams and bindings are unchanged.
 
-The three route lifecycle cards use Card 2.0 with English fixed copy and literal
-`plain_text` dynamic values. Collaboration Space cards keep their existing
-schema. Cards render display fields, Team facts, and the Space's own policy
-without rendering prompts or raw errors. The repository path is deliberately
-visible to the members of the bound conversation; the user-visible half of that disclosure is
+The route-bound, route-unbound, and Team-dissolved cards use Card 2.0 with
+English fixed copy and literal `plain_text` dynamic values. The cause-neutral
+route-ended card and Collaboration Space cards keep their existing schema.
+Cards render display fields, Team facts, and the Space's own policy without
+rendering prompts or raw errors. The repository path is deliberately visible
+to the members of the bound conversation; the user-visible half of that disclosure is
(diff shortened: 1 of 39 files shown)


.agents/domains/channel.md
```diff
@@ -552,23 +556,30 @@ Routing notification cards are rendered from this Channel's own records, at the
...
   workspace — which renders the configured repository path when the policy has
   one — base ref, and a note that new topics get their own Team;
 - **space unbound**: space name, container group display, and a status line
   saying provisioning stopped while existing Teams and bindings are unchanged.
 
-Cards use Feishu `plain_text` elements only and render display fields, Team
-facts, and the Space's own policy without rendering prompts or raw errors. The
-repository path is deliberately visible to the members of the bound conversation;
-the user-visible half of that disclosure is
+The three route lifecycle cards use Card 2.0 with English fixed copy and literal
```
ryanxiang7: Nit: this says all three route lifecycle cards use Card 2.0, but `bindingRouteEndedCard` still renders through `buildFeishuStatusCard` (the Card 1.0 shape — `wide_screen_mode` config, `header.template`, `i18n_content` fields). The bound/unbound/dissolved cards are Card 2.0; route-ended is not. Either move route-ended to a Card 2.0 builder or adjust this sentence (and the matching line in `.agents/product/README.md`) to say two.

ryanxiang7 (changes requested):
## Review: REQUEST_CHANGES

The code changes themselves look good — clean layering, no glue, tests are thorough. Two merge-hygiene blockers prevent approval right now, plus one documentation nit (inline).

### Blockers

1. **Merge conflict with `next`** (5 files: `.agents/domains/channel.md`, `.agents/product/README.md`, `feishu-cot-events.ts`, `feishu-cot-tool-rows.test.ts`, `mcp-delegate.ts`). The conflicts are mechanical: `next` already merged #378 with the byte-identical `preserveSpacing` implementation, and #380/#382 touched the delegate and the docs. A rebase that keeps both sides should be straightforward, but it must happen before this can merge.
2. **Zero CI check runs on head `804873a`.** No checks have ever run for this SHA, so there is nothing green to approve against. The rebase/re-push should trigger them; please make sure the required checks (and the internal-content scan once #367 lands) pass.

### Nit

- See the inline comment on `.agents/domains/channel.md`: the doc says all three route lifecycle cards use Card 2.0, but `bindingRouteEndedCard` still renders through the Card 1.0 `buildFeishuStatusCard` builder.

### Observations (no change required, confirming intent)

- `bindChannel` now refuses a Team whose leader runtime context is incomplete (for example a `starting` Team whose leader identity is not durable yet), where the old check only proved existence. The new bound card's content contract needs those fields, and the tests lock the refusal, so this looks deliberate — just flagging the semantic change.
- On the `TEAM_CLOSED` rejection path the route is removed and the "route ended" card is sent; if that wins the race against the final `team.state(closed)` event, the dissolved card never arrives (the routes are already gone). This race existed before and the new wording is more honest about it; the dissolved card still appears whenever the final event arrives first. Fine as designed.

### What is good here

- The `team.create` receipt gains `leader_agent_runtime` / `runtime_cwd` from data Core already owns (`TeamRecord`), and automatic provisioning consumes the receipt directly instead of a redundant `team.status` round trip — the right capability in the right layer.
- No persisted routing/config/state change; the changelog files are all new `Added` files with 0.x-safe types.
- The ten-line cap and NBSP preservation are well covered, including CRLF, terminal-LF, JSON-scalar-as-text, and the post-conversion byte budget.

ryanxiang7 (approved):
## Review: APPROVED

Round 2 on a58eea0. Both blockers from round 1 are resolved:

- **Rebase clean**: the branch now sits on current `next` (mergeable). I verified the conflict resolutions file by file against the round-1 head — every source file is byte-identical to the reviewed version, and `mcp-delegate.ts` correctly keeps both #380/#382's model-facing changes and this PR's `team.create` receipt fields.
- **CI green**: 9/9 required checks pass, including the rush build/test matrix and the internal-content tree scan.

The code review from round 1 stands: the `team.create` receipt extension is the right capability in the right layer, provisioning drops its redundant status read, the three-way route-removal distinction is clean, and the COT line cap / spacing preservation is well tested.

One non-blocking note for a fast-follow: the round-1 nit still stands — `.agents/domains/channel.md` says all three route lifecycle cards use Card 2.0, but `bindingRouteEndedCard` still renders through the Card 1.0 `buildFeishuStatusCard`. Doc or implementation, align one side when convenient.

packages/dreamux/src/service/team-collection/types.ts
```diff
@@ -325,7 +325,9 @@ export interface TeamHistoryResult {
   next_cursor: string | null;
 }
 
-export interface TeamCreateResult extends TeamSummary {
+export interface TeamCreateResult extends Omit<TeamSummary, 'leader'> {
+  /** A successful creation always publishes its TeamLeader identity. */
+  leader: AgentEntityRuntimeStatus;
```
YourWildDad: 这一段是给谁用的

packages/channel/feishu-channel/src/feishu-channel.ts
```diff
@@ -439,17 +436,13 @@ export class FeishuChannelSession {
...
-      // for that reason. `TEAM_NOT_FOUND` is a row pointing at nothing, which
-      // is this Channel correcting its own document and stays silent.
+      // TEAM_CLOSED also covers a pending dissolve that may still be refused.
+      // Announce only route removal until the final Team state proves closure.
+      // TEAM_NOT_FOUND is silent correction of a row pointing at nothing.
       await this.forgetTeamRoutes(
         plan.teamName,
         outcome.status === 'rejected' && outcome.code === 'TEAM_CLOSED'
-          ? 'team_closed'
+          ? 'route_ended'
```
YourWildDad, line 445: 这个地方为什么改了

packages/channel/feishu-channel/src/feishu-cot-adapter.ts
```diff
@@ -362,7 +362,7 @@ export class FeishuCotAdapter {
     anchor: VisibleMessageAnchor | null,
   ): void {
     state.generation += 1;
-    this.detach(key, state, 'interrupted');
+    this.detach(key, state, anchor === null ? 'interrupted' : 'completed');
```
YourWildDad, line 365: 这个地方为什么改？

packages/channel/feishu-channel/src/feishu-cot-events.ts
```diff
@@ -282,7 +283,17 @@ export function toolResultOutput(resultJson: string | null): ToolResultOutput |
...
+  for (let line = 0; line < 10; line += 1) {
+    boundary = text.indexOf('\n', boundary + 1);
+    if (boundary === -1 || boundary === text.length - 1) {
+      return { kind: 'text', text: preserveSpacing(text) };
+    }
+  }
+  return {
+    kind: 'text',
+    text: preserveSpacing(text.slice(0, boundary + 1) + TRUNCATION_MARKER),
+  };
```
YourWildDad, line 296: 这个调整是什么意思？

.agents/domains/channel.md
YourWildDad: Fixed in b4e592ea. The docs now state the actual boundary explicitly: route-bound, route-unbound, and Team-dissolved use Card 2.0; the cause-neutral route-ended card and Collaboration Space cards keep their existing schemas. I updated both the domain KB and product catalog rather than changing the card implementation, because Card 2.0 was selected only for the three designed cards.

packages/channel/feishu-channel/src/feishu-channel.ts
YourWildDad: This separates an early admission rejection from a final lifecycle fact. TEAM_CLOSED can be returned once the Team closing fence is raised, before the final team.state(status=closed) event; the dissolve can still fail or be refused. Therefore this path may remove the now-unusable route, but it must only announce the cause-neutral route-ended state. The confirmed Team dissolved card is reserved for the later final closed event. TEAM_NOT_FOUND remains a silent stale-route correction.

packages/channel/feishu-channel/src/feishu-cot-adapter.ts
YourWildDad: The anchor value distinguishes replacement from retirement here. A non-null anchor means a newer accepted inbound message superseded the old presentation: the old card completed successfully and ongoing activity moves to the new anchor. A null anchor means a lifecycle fence is retiring the presentation (route release, Team/session close), so it remains interrupted. This implements the requirement that changing the Feishu COT anchor must not label the old card as an interrupted task while preserving real interruption semantics.

packages/channel/feishu-channel/src/feishu-cot-events.ts
YourWildDad: This implements the new non-JSON 10-content-line limit. The loop locates the first ten LF boundaries. If any boundary is absent, the text has at most ten lines; if the tenth LF is the final character, it is only a terminal newline and does not create an eleventh content line. In either case the result is unchanged. If content exists after the tenth LF, the code keeps the first ten lines and appends the truncation marker. JSON objects/arrays return before this block and bypass the line rule; the existing byte-budget fitting still runs afterward. The tests cover 10 vs. 11 lines, terminal LF/CRLF, JSON exemption, and the byte limit.

packages/dreamux/src/service/team-collection/types.ts
YourWildDad: You are right to question the inheritance itself. I traced its history and current producer/consumers: TeamCreateResult inherited TeamSummary because create originally returned a post-create status snapshot plus the first-turn result. The public team.create contract has since become a dedicated receipt, but this internal shape was never narrowed. In the current path, createFromRequest only needs team_name, leader_name, leader_agent_runtime, and runtime_cwd; member_count, submission status, and error have no consumer. Commit 2328afee therefore makes TeamCreateResult an independent four-field result, removes the unused memberCount read and stale submission projections, and leaves TeamSummary solely on the status/read path.

PR #387, opened 2026-09-07
feat(dreamux): add provider-neutral channel.list command

YourWildDad:
A host managing a dispatcher needs to enumerate its configured Channels and observe their live state without asking a TeamLeader to take a turn. Add the provider-neutral v1 `channel.list` Core command, available as `dreamux server-ctl channel list --id <dispatcher>`.

The command takes no payload and returns `{ channels: [{ channel_id, provider, identity, live }] }` in configuration order. It reads the existing configured and live Channel views, uses an empty string for an absent identity, and works for stopped dispatchers without starting sessions or exposing provider configuration. This is additive: existing commands, tools, and state are unchanged.

Changed files:

- `packages/dreamux/src/service/channel-service/commands.ts`: canonical command and input/output schemas.
- `packages/dreamux/src/command/catalog.ts`: register the Channel command group between dispatcher and Team commands.
- `packages/dreamux/src/cli/server-ctl.ts`: verb mapping and usage text.
- `packages/dreamux/src/service/dispatcher-service/index.ts`: `listChannels()` metadata projection.
- `packages/dreamux/tests/commands.test.ts`: four command cases covering metadata, missing/unknown dispatcher IDs, and stopped-dispatcher reads through the real Server.
- `packages/dreamux/tests/core-command-adapters.test.ts`: CLI-to-command mapping through the admin socket.
- `packages/dreamux/tests/core-command-registry.test.ts`: pin the new command and namespace.
- `packages/dreamux/tests/helpers/command-harness.ts`: `listChannels` override, defaulting to `[]`.
- `packages/dreamux/CLAUDE.md` and `packages/dreamux/src/service/CLAUDE.md`: document ownership and behavior.
- `common/changes/@excitedjs/dreamux/feat-core-channel-list_2026-09-07-08-59.json`: generated by non-interactive `rush change` after staging the package changes, with a `minor` bump for `@excitedjs/dreamux`.

Verification (all exited 0):

| Directory | Command | Result |
| --- | --- | --- |
| Root | `node common/scripts/install-run-rush.js install` | Passed |
| Root | `node common/scripts/install-run-rush.js build --to @excitedjs/dreamux` | 8 operations passed |
| `packages/dreamux` | `rushx test --exclude tests/codex-live.test.ts` | 71 files, 1012 tests passed; live Codex file excluded |
| `packages/dreamux` | `rushx typecheck:tests` | Passed |
| `packages/dreamux` | `rushx lint` | Passed |
| Root | `bash .agents/scripts/check.sh` | Passed; 170 files reachable |
| Root | `rush change --verify --target-branch origin/next --no-fetch` | Passed after commit; found the generated change file |
| Root | `git diff origin/next...HEAD --check` | Passed |

The mandatory pre-commit hook passed, including staged ESLint, author identity, gitleaks, and internal-content checks. Rush emitted its Node.js 22.16.0 compatibility warning; it did not fail any gate.


packages/dreamux/CLAUDE.md
```diff
@@ -60,6 +60,11 @@ Two settled shape rules govern where code lives:
   delivery) inside the Dispatcher Service.
 - Own the canonical Command catalog and expose it identically to `admin.sock`
   and to an in-process Channel's `invoke` port.
+- Expose `channel.list` (CLI: `dreamux server-ctl channel list --id <ID>`) for
```
ryanxiang7: **F1 (blocker):** the command documented here cannot run. The public `dreamux` CLI has a closed `.strict()` command list (`src/cli/commands/index.ts`) with no `server-ctl` and no `channel`; `server-ctl.ts` is an internal delegated module. `dreamux server-ctl channel list --id flow` exits 2 with `Unknown argument: server-ctl`. Either add a public `dreamux channel list` wrapper (mirror `createDispatcherVerbCommand` in `src/cli/commands/dispatcher.ts`) and fix this spelling plus the change-file comment, or drop the CLI claim and document the command as admin-socket-only. The change-file comment lands verbatim in `CHANGELOG.md`, so the false spelling ships either way.

packages/dreamux/src/service/channel-service/commands.ts
```diff
@@ -0,0 +1,43 @@
...
+  BOOLEAN,
+  NO_INPUT,
+  STRING,
+  arrayOf,
+  objectSchema,
+} from '../../command/schema.js';
+import type { DispatcherService } from '../dispatcher-service/index.js';
+
+interface ChannelListResult {
+  channels: ReturnType<DispatcherService['listChannels']>;
```
ryanxiang7: **F3 (should-fix):** `ReturnType<DispatcherService['listChannels']>` derives the Command's output contract from an implementation's inferred type, and `listChannels()` itself has no explicit return type. The output schema is closed (`objectSchema` → `additionalProperties: false`): adding a field to the mapping widens this type silently (TS stays green) while `validateOutput` rejects every `channel.list` call at runtime as INTERNAL. Name the DTO (e.g. export a `ChannelMetadata` interface), give `listChannels()` an explicit return type, and reference the named type here — mirroring `DispatcherListResult { dispatchers: DispatcherSummary[] }`.

packages/dreamux/tests/commands.test.ts
```diff
@@ -1,6 +1,142 @@
...
+          {
+            channel_id: 'secondary',
+            provider: 'npm:@example/secondary',
+            identity: 'secondary-identity',
+            live: false,
+          },
+        ],
+      });
+
+      expect(fake.sessions.size).toBe(0);
```
ryanxiang7, line 79: **F4 (should-fix):** no test produces `live: true` through the real code path — case 1 stubs the whole result, and case 4's dispatcher is stopped so the live map is empty. A wrong key in `live.has(...)` (e.g. `live.has(channel.provider)`) passes all four cases. Add a case that builds and `adopt()`s a live map (see `tests/channel-service.test.ts`) and asserts `live: true` for the adopted channel.

packages/dreamux/src/cli/server-ctl.ts
```diff
@@ -76,6 +77,7 @@ function resolveMethod(obj: string | undefined, verb: string | undefined): strin
   const o = obj ?? '';
   const v = verb ?? '';
   if (o === 'server' && v === 'status') return 'server.status';
+  if (o === 'channel' && v === 'list') return 'channel.list';
```
ryanxiang7: **N1 (nit):** this `channel` branch sits before the `dispatcher` block, while the usage text below and the catalog registration (`src/command/catalog.ts`) both place channel after dispatcher. Pick one order.

packages/dreamux/src/service/channel-service/commands.ts
```diff
@@ -0,0 +1,43 @@
+/** The Channel namespace's canonical Commands. */
```
ryanxiang7: **N2 (nit):** sibling `commands.ts` files carry a rationale docblock; this one-line header is also where it's worth stating that `channel.*` is a read-only inventory surface — the namespace was previously retired (`DELETED_NAMES` in `tests/core-command-registry.test.ts`: `channel.invoke_tool`, `channel.mcp.*`), so a reader could otherwise mistake this for a revival of the Channel MCP proxy.

ryanxiang7 (changes requested):
**Verdict: REQUEST_CHANGES.** The Core command itself is well-built — clean layering, no provider config leak, no session starts on stopped dispatchers, solid contract tests — but there are two blockers at the edges.

**F1 (blocker) — the documented CLI command does not exist.** `packages/dreamux/CLAUDE.md` and the Rush change file instruct `dreamux server-ctl channel list --id <ID>`. The public `dreamux` CLI has a closed `.strict()` command list (`src/cli/commands/index.ts`): no `server-ctl`, no `channel`. `server-ctl.ts` is an internal delegated module (`.agents/decisions/cli-and-package-naming.md`, `global-bin-onboard-serve.md`), reachable only through public wrappers (`dreamux dispatcher list` → `execEntry(serverCtlEntry, ['dispatcher','list',...])`). `channel.list` is currently the only Core command with no public CLI path, and the change-file comment ships verbatim into `CHANGELOG.md` — users are told to run a command that exits 2. Fix: add `src/cli/commands/channel.ts` mirroring `createDispatcherVerbCommand` (`dreamux channel list --id <ID>`), register it in `createDreamuxCommands`, and correct the spelling in both CLAUDE.md files and the change file. (Inline on `packages/dreamux/CLAUDE.md:63`.)

**F2 (blocker) — the required `.agents/` knowledge delta is missing.** A new Core command + CLI verb is a CLI-surface change; the root `CLAUDE.md` makes a same-PR `.agents/` update mandatory. CI's KB job only checks links/orphans, so it passed on a now-false KB:
- `.agents/proposals/admin-control-plane-surface.md:77` — "There is no first-class `channel.list` or `channel.status` Command." Now false; the same proposal's "Channel session lifecycle detail" row is partly answered by the new `live` field, and its remaining-slices / out-of-scope sections name channel inventory.
- `.agents/domains/service-topology.md` — the ChannelService row names only `index.ts` and `mcp-delegate.ts`; the module now also owns `commands.ts`.
- `.agents/domains/channel.md` — Core now also publishes a read-only configured-channel inventory.

Update these and run `.agents/scripts/check.sh`.

**F3 (should-fix) — the command's output contract is derived from an inferred implementation type.** `ChannelListResult = ReturnType<DispatcherService['listChannels']>` and `listChannels()` has no explicit return type. The output schema is closed (`objectSchema` → `additionalProperties: false`), so adding a field to the mapping widens the inferred type silently (TS stays green) while `validateOutput` rejects every invocation at runtime as INTERNAL. Sibling pattern: `DispatcherListResult { dispatchers: DispatcherSummary[] }` with a named exported interface. (Inline on `commands.ts:17`.)

**F4 (should-fix) — `live: true` is never produced by the real code path.** Case 1 stubs the whole result; case 4 runs against a stopped dispatcher, so the live map is empty. A wrong key in `live.has(...)` (e.g. `channel.provider`) passes all four cases. (Inline on `commands.test.ts:121`.)

**Nits.** N1: `resolveMethod` places the `channel` branch before `dispatcher`, while the usage text and catalog registration place it after — pick one order (`server-ctl.ts:80`). N2: the new `commands.ts` carries a one-line header where siblings carry a rationale docblock — worth stating this is a read-only inventory surface, not a revival of the retired `channel.*` MCP-proxy namespace (`commands.ts:1`).

**Verified clean:** no provider config leak and no session start (only the config array and the adopted map are read; `build()` is the sole path to `createSession`); catalog registration matches the established pattern; the harness default `[]` mirrors `listTeams` and hides nothing (case 4 exercises the real `Server`); `identity: ''` matches the established representation; change file type `minor` is correct for the 0.x line; tests are not weakened (`FROZEN_NAMESPACE_TABLE` is the sanctioned update path, `DELETED_NAMES` untouched); no public-repo red-line issues. CI 9/9 green.

packages/dreamux/CLAUDE.md
YourWildDad: Accepted and fixed in [5f1f597a](https://github.com/excitedjs/dreamux/commit/5f1f597a9105539e805f5a84b87c31f4b55ccacb). `packages/dreamux/src/cli/commands/channel.ts` now exposes `dreamux channel list --id <ID>` through the existing admin delegate, and `src/cli/commands/index.ts` registers it. Both package guides now use the public spelling, and the regenerated Rush change file uses it too. `packages/dreamux/tests/commands.test.ts` exercises the compiled public bin against a real Server/admin socket and verifies the returned inventory without starting Channel sessions.

Two small factual clarifications: the existing Team and TeamMate Core Commands also lack public CLI wrappers, and the original change-file comment said `server-ctl channel list ...` without a `dreamux` prefix. The documented public invocation was still unusable, so the finding and the public-wrapper fix are accepted.

packages/dreamux/src/service/channel-service/commands.ts
YourWildDad: Fixed in [5f1f597a](https://github.com/excitedjs/dreamux/commit/5f1f597a9105539e805f5a84b87c31f4b55ccacb). `packages/dreamux/src/service/channel-service/types.ts` exports `ChannelMetadata`; `channel-service/commands.ts` now declares `ChannelListResult.channels` as `ChannelMetadata[]` instead of deriving it with `ReturnType`. In `dispatcher-service/index.ts`, `listChannels()` explicitly returns `ChannelMetadata[]`, and its map callback explicitly returns `ChannelMetadata` so extra fields on the constructed object are checked at compile time. The closed wire schema is unchanged. Source and test typechecks pass.

packages/dreamux/tests/commands.test.ts
YourWildDad: Fixed in [5f1f597a](https://github.com/excitedjs/dreamux/commit/5f1f597a9105539e805f5a84b87c31f4b55ccacb). The new case in `packages/dreamux/tests/commands.test.ts` starts a real Dispatcher backed by the fake Channel provider. Production startup runs `ChannelService.build()` and `startPreparedChannels()`/`adopt()`, then the real `channel.list` Command returns `live: true` for both configured Channel IDs. Their provider refs differ from those IDs, so checking the live map by provider ref would fail this assertion. The test also stops the Dispatcher and verifies that both entries return to `live: false`. The original four command cases remain, and the suite passes.

packages/dreamux/src/cli/server-ctl.ts
YourWildDad: Fixed in [5f1f597a](https://github.com/excitedjs/dreamux/commit/5f1f597a9105539e805f5a84b87c31f4b55ccacb). `packages/dreamux/src/cli/server-ctl.ts` now resolves `channel` after `dispatcher`, matching the usage text and Core catalog ordering. The new public wrapper is also registered immediately after the dispatcher wrapper in `src/cli/commands/index.ts`.

packages/dreamux/src/service/channel-service/commands.ts
YourWildDad: Fixed in [5f1f597a](https://github.com/excitedjs/dreamux/commit/5f1f597a9105539e805f5a84b87c31f4b55ccacb). The rationale docblock in `packages/dreamux/src/service/channel-service/commands.ts` now explains the read-only inventory use case and explicitly distinguishes it from the retired `channel.invoke_tool` / `channel.mcp.*` proxy Commands. Provider tools remain behind runtime-bound MCP delegates and the generic `mcp.*` transport Commands.

YourWildDad:
Addressed F1–F4 and N1–N2 in [5f1f597a](https://github.com/excitedjs/dreamux/commit/5f1f597a9105539e805f5a84b87c31f4b55ccacb).

- F1: Added the public `dreamux channel list --id <ID>` wrapper and registration, corrected both package guides and the Rush change-file comment, and added a compiled-public-bin test against the real Server/admin socket.
- F2: Updated `.agents/domains/channel.md`, `.agents/domains/service-topology.md`, and `.agents/proposals/admin-control-plane-surface.md` to describe the implemented inventory and remaining control-plane work.
- F3: Exported `ChannelMetadata`, used it in the Command output type, and explicitly typed both `listChannels()` and its mapping callback.
- F4: Added real Dispatcher startup coverage through Channel build/start/adopt, verified `live: true` by Channel ID, and checked that stopping clears liveness. The original four command cases remain.
- N1/N2: Aligned dispatcher/channel ordering and added the inventory rationale, including the boundary with retired Channel MCP proxy Commands.

Validation on the submitted patch:

| Directory | Command | Exit code / result |
| --- | --- | --- |
| `packages/dreamux` | `rushx build` | 0 |
| `packages/dreamux` | `rushx lint` | 0 |
| `packages/dreamux` | `rushx typecheck` | 0 |
| `packages/dreamux` | `rushx typecheck:tests` | 0 |
| `packages/dreamux` | `rushx test --exclude tests/codex-live.test.ts` | 0; 71 files, 1014 tests passed |
| Root | `bash .agents/scripts/check.sh` | 0; 170 files reachable |
| Root | `rush change --verify --target-branch origin/next --no-fetch` | 0; rerun after commit |
| Root | `git diff origin/next...HEAD --check` | 0 after commit |
| Root | `packages/dreamux/bin/dreamux channel list --help` | 0 |

The live Codex file was excluded as planned. Rush emitted its Node.js 22.16.0 compatibility warning without failing a gate. The mandatory pre-commit hook passed, including staged ESLint, author identity, gitleaks, and internal-content checks. The change remains a `minor` bump for `@excitedjs/dreamux`.

packages/dreamux/CLAUDE.md
YourWildDad: Following the maintainer's decision, F1 now follows your admin-socket-only alternative.

The maintainer's rule is that the public `dreamux` CLI is reserved for host lifecycle operations (onboard/serve/status/doctor/daemon/dispatcher/config/changelog/mcp); Commands consumed programmatically by Channels, MCP, or the admin socket (Team, TeamMate, Channel inventory) do not get CLI verbs, and `server-ctl` only backs the public wrappers.

Commit `1b5376bb563ce3c463f250b8a4b3df95e24ea8f3` removes `packages/dreamux/src/cli/commands/channel.ts` and its registration introduced in `5f1f597a`, and restores `packages/dreamux/src/cli/server-ctl.ts` byte-for-byte to its pre-PR version. `channel.list` remains available through the admin socket and a Channel's in-process `invoke` port. Both package guides, `.agents/domains/channel.md`, and the Rush change note now reflect this boundary.

The named DTO, explicit return type, rationale docblock, and real start/adopt/stop coverage are retained. The CLI-dependent tests are replaced with direct admin-socket/Channel-invoker equivalence coverage in `packages/dreamux/tests/core-command-adapters.test.ts`.

Build, lint, both typechecks, all 1,013 tests with `tests/codex-live.test.ts` excluded, knowledge checks, and change-file verification pass. An actual `channel list --id x` invocation is rejected with exit code 2; `--help` retains the existing root-help behavior and exits 0.

YourWildDad:
Commit `1b5376bb563ce3c463f250b8a4b3df95e24ea8f3` withdraws the public Channel CLI wrapper introduced in `5f1f597a`, following the maintainer's decision to reserve the public CLI for host lifecycle operations. F1 now takes the admin-socket-only route: `channel.list` is callable through the admin socket and a Channel's in-process `invoke` port, without a public CLI verb or `server-ctl` mapping.

The CLI command registry, CLI entry point, and `server-ctl.ts` are byte-identical to their pre-PR versions. Both package guides, the Channel domain documentation, and the Rush-generated minor change note describe the programmatic surface. The `ChannelMetadata` DTO, explicit return type, inventory rationale docblock, and real start/adopt/stop test remain. Direct admin-socket/Channel-invoker equivalence coverage replaces the CLI-dependent adapter test.

Validation for this commit (package commands run in `packages/dreamux`, other commands at repository root):

| Command | Exit code / result |
| --- | --- |
| `rushx build` | 0 |
| `rushx lint` | 0 |
| `rushx typecheck` | 0 |
| `rushx typecheck:tests` | 0 |
| `rushx test --exclude tests/codex-live.test.ts` | 0; 71 files, 1,013 tests passed |
| `bash .agents/scripts/check.sh` | 0; 170 files reachable |
| `rush change --verify --target-branch origin/next --no-fetch` | 0, before and after commit |
| `git diff HEAD --check` | 0, before and after commit |
| `git diff HEAD^ HEAD --check` | 0 |
| `packages/dreamux/bin/dreamux channel list --id x` | 2; unknown arguments, as expected |
| `packages/dreamux/bin/dreamux channel list --help` | 0; existing root-help behavior, with no Channel command listed |

The `--help` probe does not produce a nonzero exit because the existing yargs help branch skips argument validation. The actual invocation is rejected, and the public parser behavior is unchanged.

ryanxiang7 (approved):
**Verdict: APPROVE.** Re-reviewed the two fix commits (`5f1f597`, `1b5376b`) against the previous findings; all are resolved.

- **F1 (blocker) — resolved by design decision.** The public-CLI claim was dropped rather than papered over: the `channel list` verb mapping and usage lines are removed from `server-ctl.ts`, both CLAUDE.md files and the change-file comment now state the command is reachable via the admin socket and the in-process `invoke` port only, and the rationale ("the public CLI is reserved for host lifecycle operations") is recorded. No `server-ctl channel` reference remains anywhere in the tree. The adapter test was rewritten accordingly: it now proves adapter equivalence (admin.sock vs Channel invoker) instead of a CLI mapping that no longer exists.
- **F2 (blocker) — resolved.** The `.agents/` delta is thorough and accurate: `domains/channel.md` gains a "Configured Channel inventory" contract section with source links and the no-public-CLI statement; `domains/service-topology.md` names `commands.ts` in the ChannelService row; `proposals/admin-control-plane-surface.md` updates the intro, both gap-table rows (channel inventory landed; `live` flag partly answers session lifecycle detail), the out-of-scope list, and the reviewer checklist. `check.sh` passes in CI.
- **F3 (should-fix) — resolved.** `ChannelMetadata` is now a named exported interface in `channel-service/types.ts`; `listChannels(): ChannelMetadata[]` has an explicit return type with a per-element annotation; `ChannelListResult` references the named type. The drift hazard (closed output schema vs inferred type) is gone.
- **F4 (should-fix) — resolved.** A new case drives the real dispatcher start path (`build` → start → `adopt`), asserts `live: true` for both adopted channels, then stops and asserts `live: false`. A wrong `live.has(...)` key would now fail. The stopped-dispatcher case was refactored onto a shared `createChannelListServer()` fixture without losing coverage.
- **N1 — moot** (the `resolveMethod` branch was removed). **N2 — resolved** (the docblock now states the read-only inventory rationale and explicitly distinguishes it from the retired `channel.*` MCP-proxy Commands).

Verified clean on the new head: CI 9/9 green; no stale CLI references; `FROZEN_NAMESPACE_TABLE` still pins `channel.list`; change file type `minor` correct for the 0.x line.

One non-blocking nit: the PR description still says "available as `dreamux server-ctl channel list --id <dispatcher>`" — worth updating to match the admin-socket/invoke-port decision, since the description is what future readers land on first.

PR #389, opened 2026-09-07
fix(dreamux): fence completion delivery across teardown

YourWildDad:
## Summary

- retire every not-yet-started TeamMate completion at explicit close, Team dissolve, Dispatcher shutdown/restart, and failed-start rollback fences
- make asynchronous TeamMate, Team, member, and TeamLeader construction inherit the aggregate fence before publication or submission
- keep Turn settlement convergence and never retract a completion delivery that already started
- suppress Workflow terminal delivery only when explicit or aggregate stop wins the existing first-terminal-intent arbitration
- reduce Workflow terminal delivery bookkeeping to one nullable source-owned obligation while preserving retry and natural completed/failed delivery

## Regression history

PR #350 made every settled `EntityTurn` start its completion closure, including close-induced `stopped` settlements. Before that PR, deliberate close happened not to push because completion-token eligibility gated the path, but that architecture also hid unrelated failed or unexpected stopped outcomes. PR #380 only added the automated-notification wording.

This change keeps PR #350's useful settlement visibility and retires the obligation at the named lifecycle owner instead of filtering rendered notification text downstream.

Closes #388

## Verification

- `node common/scripts/install-run-rush.js update`
- `node common/scripts/install-run-rush.js build`
- `node common/scripts/install-run-rush.js lint`
- `node common/scripts/install-run-rush.js test`
- `node common/scripts/install-run-rush.js typecheck:tests`
- `node common/scripts/install-run-rush.js change --verify --target-branch origin/next --no-fetch`
- `.agents/scripts/check.sh`
- `git diff --check`
- mandatory pre-commit gitleaks and internal-content scans
- independent implementation review: `ACCEPTABLE`
- Fable architecture review: `ACCEPTABLE`

All listed checks pass locally. Node.js 22.16.0 emits the repository's untested-version warning, but every gate exits successfully.

Issue #388, opened 2026-09-07
Solution: drop lifecycle-stop completion pushback


YourWildDad:
## Problem

When an owning Agent deliberately tears down a TeamMate or Workflow, the pending runtime result currently converges to `stopped` and is delivered back as another model turn. The owner already knows it requested the stop, so this extra input consumes tokens and can make the model reason about its own action. Team dissolve and Dispatcher shutdown/restart can produce a storm of the same cleanup completions while recipients are themselves stopping.

The same Core paths serve TeamLeader and Dispatcher owners.

## History

- PR #149 introduced ordinary reverse completion delivery for completed, failed, and stopped TeamMate turns. That contract remains valid for independently terminating work.
- PR #338 consolidated TeamMate lifecycle teardown and established Workflow terminal delivery through the current finalization path.
- PR #350 directly introduced the TeamMate regression by making every selected `EntityTurn` outcome deliver, including close-induced `stopped`. It also later locked explicit Workflow-stop delivery in a test, but did not originate Workflow terminal delivery.
- PR #380 only made the automated-notification wording more explicit.

## Required behavior

- Explicit TeamMate close abandons every completion delivery that has not started.
- Explicit `workflow_stop` abandons terminal delivery only when stop wins the run's existing first-terminal-intent race.
- Team dissolve abandons pending member-to-TeamLeader, TeamLeader-to-Dispatcher, and active Team Workflow terminal delivery.
- Dispatcher shutdown/restart abandons pending delivery from materialized TeamMates, TeamLeaders, and active Dispatcher Workflows.
- Turn settlement, Workflow terminal persistence, cleanup, and stop receipts still converge.
- Independent TeamMate completed/failed/stopped outcomes still deliver normally.
- A Workflow completed/failed intent selected before stop still delivers exactly once.
- Delivery already started before teardown is not retracted.
- Token folding, recipient FIFO ordering, durable closed state, and public Command/MCP shapes remain unchanged.

The TeamLeader self-dissolve consequence is intentional: preserving only that result would require a new causal mode through Team closing. The operator delegated this edge to code-complexity judgment, so the solution keeps one uniform teardown rule.

### TeamMate turns

Keep the decision at the source Turn in Core.

1. Make `EntityTurn`'s stored delivery closure mutable and add `abandonPendingDelivery()`. Clear only while delivery has not started.
[...]

PR #391, opened 2026-09-08
Fence completion delivery at the recipient's scope

YourWildDad:
## Summary

Redo of #389 on a fresh branch from `next`. Same requirement (Refs #388), different shape: the scope that owns a recipient publishes whether it still wants news, and delivery reads that fence once, at the moment it would report. Producers never learn their owner's lifecycle.

One rule: **a settled turn is reported unless its owner ended it.** Three reads implement it:

- **Dispatcher scope.** `CompletionDeliveryPolicy.deliverRuntime` drops a completion when the dispatcher's admission gate (`DispatcherTaskDrain.accepting`) is closed, before folding or queueing. Nothing that settles behind a stopping dispatcher's fence is delivered; a delivery that already started is never retracted.
- **Entity scope.** `EntityTurn` asks its owner `owesCompletion()` (`phase === 'active' && hostStop === null`) once at settlement. A negative answer releases the delivery closure for good, so a later `ensureDelivery()` cannot revive the report. `releaseHostRuntime` now drains and converges retained turns even when the native stop fails, while `hostStop` is still set.
- **Workflow scope.** `WorkflowRun` holds a nullable `deliverTerminal`; the terminal's `closeAdmission('stopped')` clears it, so a stopped run reports nothing and every other terminal reports once.

Team scope needed no code change: `TeamService.admit()` / `enterOrdinaryMutation` already refuse during dissolve.

## Not ported from #389

`CompletionDeliveryScope` epochs, `abandonPendingDelivery` sweeps, construction-time inheritance (`retireIfStale` / `observeConstruction` / `onConstructed`), `rearm*`, `TeamCompletionDelivery` / `aggregateAbandoned`, and `reserveStop(): boolean`. The behavior tests from #389 (`teammate-completion-lifecycle.test.ts`, `controlled-runtime-provider.ts`, the workflow race tests) are reused with the same substance; two names were updated to describe the fence they exercise.

## Knowing changes

Test contracts changed on purpose (reviewed against the source contract, not against a green run):

- `entity-turn.test.ts`: a turn the owner's own close ended is settled and never reported.
- `workflow-service.test.ts`: a stopped Workflow run delivers no terminal report.

Two behavior deltas versus #389, both accepted by inclusion in the approved redo and recorded in the task record:

- After a failed teardown, an entity that is still active reports again.
- A leader turn that completes naturally inside a dissolve's Workflow-stop window reaches the dispatcher.

## Shared modules replacing copied mechanics

The fence reads pushed `TeammateService` and `WorkflowRun` past the 700-line lint gate. It is met by removing duplication, not by splitting files:

- `service/closed-fact.ts` (`ClosedFactPublisher<Fact>`): one durable-close broadcast, replacing `TeamClosedPublisher` (`team-service/closed-fact.ts`, since #350) and the identical listener loop `TeammateService` carried inline. Each owner builds its own fact beside the fact's type.
- `service/in-flight-work.ts` (`InFlightWork`): one tracker for work a scope admitted and must join before it stops, replacing the `Set<Promise>` drain loops in `DispatcherTaskDrain`, three in `WorkflowRun`, one in `WorkflowService`, and the counter plus idle waiters in `TeammateService`. The unexplained extra microtask after `WorkflowRun` joined its materializations (#338) goes with the loop; the Workflow stop race tests are the ordering evidence.
- The terminal completion fact stays inline in `WorkflowRun.finalize()`; the dead `deferred()` helper in `run-support.ts` is deleted, and so is `DispatcherTaskDrain.trackAccepted`, whose only caller left with #350.

Recorded, not done here: the serialized-tail idiom is copied in four places beside `KeyedAsyncQueue`; unifying it touches the state-write lease re-check and the admission line, so it is its own change.

## Records

- Task record: `.agents/tasks/completion-routing/suppress-owner-close-stop-pushback/` (requirement with operator quotes, final design, superseded #389 design).
- KB: `dispatcher-orchestration.md`, `service-topology.md`, `service/CLAUDE.md`, product catalog, maintenance `service-lifecycle.md`.
- Rush change file (`patch`).

## Verification

`rush build`, `rush lint`, `rush test`, `rush typecheck:tests`, `.agents/scripts/check.sh`, and `git diff --check` all pass on the committed HEAD.

Supersedes #389. Refs #388.

https://claude.ai/code/session_015GDei27iEAeUEFahreY6xq


ryanxiang7 (approved):
## Verdict: Approved

The redo is materially simpler than the superseded producer-side design in #389. Placing "is this report still wanted" on the recipient's scope — which already publishes its fence for admission reasons — needs none of the epoch propagation, population sweeps, construction-time staleness inheritance, or re-arm paths that made #389 read like 各处的堵漏. Each of the three fences is one read of a fact the scope already owned for another reason:

- **Entity fence** — `owesCompletion = phase === 'active' && hostStop === null`, read once at settlement; a negative read releases the closure permanently, so no `ensureDelivery()` revival path exists.
- **Dispatcher fence** — `DispatcherTaskDrain.accepting` read at the top of `deliverRuntime`, before folding and before queueing, so a folded token is not a way past the gate. `stop()`, `beginShutdown()`, and failed-start rollback close it synchronously; successful rollback reopens.
- **Workflow fence** — nullable `deliverTerminal` cleared in the `closeAdmission('stopped')` callback; only `reserveStop()` reserves `stopped`, so a completed/failed intent that wins first keeps its report.

No-retraction is structural in all three (synchronous delivery-task capture, finalize-after-intent with a local const, queued tails never re-checked) rather than enforced by convention.

The 700-line gate was met by extracting genuinely shared mechanics, not by splitting classes: `ClosedFactPublisher` unifies the Team and TeamMate durable-close broadcasts (deleting `team-service/closed-fact.ts` and the inline listener loop), and `InFlightWork` replaces six copies of the admitted-work tracker. The entropy accounting in the final design doc matches the diff.

The test contract inversions are requirement-driven, not weakened assertions: the baseline close-induced-stopped delivery test is inverted because the requirement itself inverts that behavior, while the PR #149 rule (an independently failed/stopped turn is still delivered once) is kept as an explicit test. The new lifecycle suite observes actual owner runtime submissions across the real race windows, including the two subtle cases the requirement calls out (a leader turn completing naturally inside the dissolve's Workflow-stop window is still delivered; an admission in flight when a host stop fails still settles behind the entity fence while the failure is rethrown).

Docs delta is complete (requirement with verbatim rulings, final design, product catalog, both domain docs, service CLAUDE.md, maintenance reference) and the knowledge-base check passes in CI. Change file is `patch` on the 0.x line as required.

PR #392, opened 2026-09-08
Strengthen independent review prompts

YourWildDad:
## Summary

Make Dreamux solution consultants and implementation reviewers independent challengers of TeamLeader framing while keeping entropy reduction a hard repository acceptance criterion.

The identity blocks own that standing posture. Work prompts carry only task-specific inputs, assigned outputs, write boundaries, and report contracts, so the same instruction is not duplicated across identities and prompts.

## Scope

- Split the solution identity set into a shared posture plus role-correct author and draft-reviewer responsibilities.
- Strengthen the implementation-review identity without overlaying it onto shared finder angles.
- Keep simple and complex solution-consultation prompts task-specific and pass the relevant identities intact.
- Preserve all existing reviewer counts, turns, workflow topology, schemas, approval authority, and the shared dynamic code-review method.

The shared review method continues to own architecture challenge through its cleanup finder's simplification and Altitude angles. Dreamux keeps exactly two review deltas: approved requirement/solution context in scope, and the requirement-fidelity finder.

## Review follow-up

- Addressed all five requested inline changes and recorded their adjudication.
- Replaced duplicated entropy definitions with a pointer to engineering-whitepaper Section 0 while retaining the hard rejection boundary.
- Moved draft-specific reconstruction, growth alarms, and no-finding language out of the common author/reviewer block and into the solution-reviewer block.
- Rewrote the branch from current `next` as one clean commit, removing the previously committed development home path from reachable PR history.
- Recorded the broader absolute-path guardrail gap as an out-of-scope follow-up rather than expanding this skill-only patch.

This is intentionally a documentation-only correction. Prompt effectiveness remains observable by the operator through later real review output; this PR does not add a test harness or review mechanism for it.

## Verification

- Skill frontmatter validation
- Task record validation
- `.agents/scripts/check.sh`
- `git diff --check`
- Public-content scan over the final diff

## Records

- `.agents/tasks/architecture/strengthen-independent-review-prompts/`

Refs #389 and #391.


.agents/skills/dev-workflow/references/implementation-review.md
```diff
@@ -44,7 +44,10 @@ verdicts, grouping, report assembly, gear parameters, and failure semantics appl
 unchanged, and picking the gear above selects one of those parameters rather than
 adding a third delta. Both deltas close one gap: the scope stage resolves the diff
 and never learns what the operator approved, so no shared angle can check the
-implementation against it.
+implementation against it. The common challenger identity reaches this review
+through the requirement-fidelity finder below, not through the shared angles: a
+shared finder's candidate needs a user-visible failure scenario and the verifier
```
ryanxiang7: The rationale for dropping the third delta is contradicted by the shared method it cites, in three ways:

1. The shared finder set already owns an **Altitude** angle — "Special cases layered on shared infrastructure are a sign the fix is not deep enough" (`packages/dreamux/skills/shared/dynamic-workflow/references/code-review.md:223`). That is an architecture challenge without a trigger.
2. Altitude candidates explicitly do not need a user-visible failure: their `failure_scenario` "states the concrete cost — what is duplicated, what work is wasted, what becomes harder to maintain — instead of a crash" (`code-review.md:242-244`).
3. REFUTED has only four rungs — factually wrong / provably impossible / already handled / pure style with no observable effect (`code-review.md:285-287`). A named-concept altitude candidate is none of those, so it lands PLAUSIBLE and becomes a finding.

The paragraph also self-refutes: line 60 below sends the requirement-fidelity finder's candidates through this same verifier "as correctness". If the ladder truly had no rung for trigger-less architecture challenges, the escape hatch this PR relies on would die too.

The conclusion (drop the delta) is correct; the reason is not. The accurate reason is **ownership**: the shared method already owns the altitude angle, and this file (lines 61-63) requires the shared method to keep "sole ownership of its angles and gears" — composing Dreamux posture into shared finders would break that seam. The same wording is mirrored in `technical-design/final.md:52-56`, `requirement.md:85-90`, and `verification.md:32`; all four sites need the reason replaced together.

.agents/skills/dev-workflow/SKILL.md
```diff
@@ -127,6 +127,12 @@ pushes, PR actions, and the final merge outcome. Technical consultation may
 delegate only each seat's disjoint proposal or solution-review task file. TeamMate
 agreement and vote count never replace TeamLeader judgment or an operator decision.
 
+Every technical-solution consultant and implementation-review TeamMate carries the
+common challenger identity from
+[solution-identities.md](references/solution-identities.md) or
+[reviewer-identities.md](references/reviewer-identities.md). Those blocks own the
+posture: pass them, do not paraphrase them into a work prompt.
```
ryanxiang7, line 134: This new rule is contradicted by this same diff. `solution-consultation.md:69-76` paraphrases the common identity three times in the hand-down prompt:

- "reconstruct the user story, authoritative owner, and simplest coherent mechanism" ≈ identity lines 10-12
- "a hypothesis to challenge rather than an answer to complete" ≈ line 13
- "an evidence-backed no-finding result is valid" ≈ line 24

A TeamLeader following this rule keeps posture out of the prompt; one following `solution-consultation.md` puts it in — opposite instructions from two references loaded for the same hand-down, and the paraphrase will drift the next time the identity is edited (the exact drift cited in `verification.md:34` as the reason for this very `SKILL.md` change).

Fix in the direction this PR already took for `SKILL.md`: convert `solution-consultation.md:69-76` into a pointer ("the common identity owns the reviewer's posture; pass it and keep the prompt to paths, files, and write boundaries"). The seat already receives the identity block via the `identity` field (lines 63-64), so the prompt does not need to restate it.

.agents/skills/dev-workflow/references/solution-identities.md
```diff
@@ -6,22 +6,35 @@ field. Keep task paths and the assigned output path in the work prompt.
...
-`.agents/skills/engineering-whitepaper/SKILL.md` first and design with its taste:
-minimal mechanism, no unrequested defenses, and existing design treated as
-evidence rather than a verdict.
+You are an independent technical solution challenger, not a confirmer of the
+TeamLeader's framing. Before extending or defending a proposed answer, reconstruct
+the user story, authoritative owner, and simplest coherent mechanism from the
+recorded requirement and current source. Treat the TeamLeader's proposal and the
+existing design as evidence to evaluate, never as a verdict to complete.
+
+Entropy reduction is a hard repository acceptance criterion. A solution or
```
ryanxiang7: This paragraph restates definitions the block already points at. The identity itself says "Read `.agents/skills/engineering-whitepaper/SKILL.md` first", and that whitepaper already owns all three points more sharply: §0 "The root rule: entropy reduction" (with the operational definition — the count of concepts, entities, mechanisms, persisted facts, states, special cases, and cross-layer hops), §3 "Write the user story before the contract", §4 "Existing design is evidence, not a verdict". Restating them in weaker prose creates a second owner that will drift.

Nuance: the task's acceptance criterion (`requirement.md:67`) requires the identities to *carry* the posture, so this should not be deleted wholesale. Keep "entropy reduction is a hard repository acceptance criterion" as the carried statement and compress the definition into a pointer to whitepaper §0.

Please preserve the genuinely new teeth this PR adds — the challenger stance, "the approved solution bounds what the writer was authorized to change", the authority closure, and the architecture-blocker definition in `reviewer-identities.md:35-37`, which is the only place this PR actually operationalizes entropy reduction.

.agents/skills/dev-workflow/references/solution-identities.md
```diff
@@ -6,22 +6,35 @@ field. Keep task paths and the assigned output path in the work prompt.
...
 
 ```identity
-You are an independent technical solution consultant for a clarified Dreamux task.
-Use the recorded requirement and current code as authority. Do not modify product
-implementation. Write only the single proposal or review file assigned to your
-seat, and make ownership, boundaries, trade-offs, and verification explicit. Read
-`.agents/skills/engineering-whitepaper/SKILL.md` first and design with its taste:
-minimal mechanism, no unrequested defenses, and existing design treated as
-evidence rather than a verdict.
+You are an independent technical solution challenger, not a confirmer of the
```
ryanxiang7: The common block is reviewer-voiced, but it is also handed to **solution-author** seats in the complex three-proposal path (`solution-consultation.md:86-88`), where no TeamLeader draft exists in round 1. "Not a confirmer of the TeamLeader's framing", "before extending or defending a proposed answer", and "treat the TeamLeader's proposal as evidence" have no referent there, and "allow a no-finding result" is wrong for a seat that must deliver a complete proposal — owner, boundary, contracts, verification, risks (`solution-consultation.md:98-106`). The author-facing half this block used to carry ("make ownership, boundaries, trade-offs, and verification explicit") was removed and now survives only in the prompt's bullet list.

Suggest moving the challenger-framing sentences into the "Solution reviewer" seat block and keeping the common block to what both seats share.

.agents/tasks/architecture/strengthen-independent-review-prompts/verification.md
```diff
@@ -0,0 +1,54 @@
...
+- The common solution and implementation-review identities now carry the
+  operator's challenger and entropy-reduction boundary.
+- Prompt guidance requires independent requirement-and-source reasoning before
+  comparing with a TeamLeader answer.
+
+## Independent review adjudication
+
+| Finding | Decision | Resolution |
+| --- | --- | --- |
+| `SKILL.md` still counted two Dreamux review deltas while the reference counted three | Accepted | Updated the entrypoint to count and name the challenger-identity delta. |
```
ryanxiang7: These first two rows describe edits that commit 2 rolled back: row 1 says `SKILL.md` was "updated to count and name the challenger-identity delta" (three deltas), but the final state counts two, and row 2's angle-identity edit was dropped along with the delta. The supersede note only appears in the second table's row 2, so a reader hits the stale rows first. Mark these two rows' Decision as `Superseded` directly.

ryanxiang7 (changes requested):
Verdict: REQUEST_CHANGES.

The direction is right — the challenger stance and the architecture-blocker definition in `reviewer-identities.md` are real improvements, and I verified the topology claims: no seats/turns added, schemas and approval authority unchanged, `packages/dreamux/skills/shared/` untouched, and no change file is required (`.agents/skills/dev-workflow` is not a bundled skill; `BUNDLED_SKILL_NAMES` covers only `packages/dreamux/skills/`). The PR #389 retrospective labeling also checks out against that branch's review records.

What blocks: two of the reasons this PR uses to justify itself do not survive a check against the source it cites, and one new rule is contradicted by the same diff. Details in the five inline threads:

- **Major** — the rationale for dropping the third delta is factually wrong about the shared method (`implementation-review.md:49`).
- **Major** — the new "do not paraphrase" rule is violated by this same diff (`SKILL.md:134` vs `solution-consultation.md:69-76`).
- **Minor** — the entropy paragraphs restate the whitepaper the block already loads (`solution-identities.md:15`).
- **Minor** — the common identity is reviewer-voiced but is also handed to author seats (`solution-identities.md:9`).
- **Minor** — the first two adjudication rows describe edits commit 2 rolled back (`verification.md:18`).

Two non-inline items:

1. **Commit history (operator decision).** Commit `acaafe7` is public on this same-repo branch and still contains a developer home path with the company name in `verification.md:29` (`/Users/bytedance/.codex/skills/...`). The head tree is clean, so squash-merge keeps `next` clean, but the commit stays reachable in PR history. Options: force-push a rewritten branch before merge, or accept. Separately, the guardrail has a gap this PR exposed: `common/scripts/check-internal-content.sh` `FORBIDDEN_RE` matches `/home/<user>/` but has no `/Users/` pattern, so the macOS form passed pre-commit, the CI tree scan, and the release gate. That fix is outside this PR's declared scope and should be its own change.
2. **PR body accuracy (nit).** Scope item 3 says the `SKILL.md` rule replaces a restatement "instead of restating them"; relative to `next` there was no restatement there to replace — the net change is a new 6-line routing paragraph. Worth rewording so the body matches the diff.

A pre-existing observation, not introduced here and not blocking: the challenger identity is slotted into a recall-biased finder (`implementation-review.md:56-60`) while the shared method says "suppression belongs to verify, not to find" (`code-review.md:258`), and the candidate schema has no blocker/improvement fields. Worth a line in the task record or a follow-up so it is not lost.

Reviewed at head `7f5684f`. CI 9/9 green; `.agents/scripts/check.sh` passes locally.

.agents/skills/dev-workflow/references/solution-identities.md
YourWildDad: Addressed in 86ba752. Draft reconstruction, growth alarms, evidence requirements, and the valid no-finding outcome now live in the solution-reviewer block. The common block is role-neutral, and the author block still requires a complete proposal.

.agents/skills/dev-workflow/SKILL.md
YourWildDad: Addressed in 86ba752. The solution-consultation prompts now carry only task-specific inputs, outputs, write boundaries, and report contracts; they pass the common and seat identities intact instead of paraphrasing their posture.

.agents/tasks/architecture/strengthen-independent-review-prompts/verification.md
YourWildDad: Addressed in 86ba752. The first two adjudication rows are now explicitly marked Superseded and point to the later PR review that removed those edits.

.agents/skills/dev-workflow/references/solution-identities.md
YourWildDad: Addressed in 86ba752. The common solution identity keeps the hard architecture-rejection boundary and now points to engineering-whitepaper Section 0 instead of restating its entropy definition.

.agents/skills/dev-workflow/references/implementation-review.md
YourWildDad: Addressed in 86ba752. The rationale now follows ownership: the shared cleanup finder already owns simplification and Altitude angles, so adding the Dreamux identity would duplicate responsibility. The verifier-ladder claim is removed.

ryanxiang7 (approved):
APPROVE.

All five findings from the requested-changes review are resolved at head `86ba752`, and the branch rewrite also purges the home-path commit from the public PR history (the old `acaafe7` is no longer reachable; the new tree is clean).

Verified per finding:

- **Dropped-delta rationale** — replaced at all four mirror sites with the ownership reason that actually holds: the shared method already owns architecture challenge through the cleanup finder's Simplification and Altitude angles (`code-review.md:213,223`), so composing the Dreamux identity there would give one responsibility two owners. The false "user-visible failure / no verifier rung" claims are gone from `implementation-review.md`, `final.md`, `requirement.md`, and `verification.md`.
- **Paraphrase self-contradiction** — `solution-consultation.md` now passes the identities intact and keeps work prompts to task inputs, assigned outputs, write boundaries, and report contracts, matching the `SKILL.md:134` rule instead of contradicting it. A grep across the skill finds no restatement of the identity posture outside the two identity files.
- **Whitepaper restatement** — the common identity now carries the hard acceptance boundary and points to engineering-whitepaper Section 0 for the definition. The new teeth are intact: the challenger stance, "the approved solution bounds what the writer was authorized to change", the authority closure, and the architecture-blocker definition in `reviewer-identities.md`.
- **Author-seat voice** — the common block is seat-neutral again; draft reconstruction, growth alarms, and no-finding language moved to the Solution reviewer block, while the author block keeps complete-proposal ownership and gains "challenge inherited ownership and architecture assumptions instead of copying the current shape".
- **Stale adjudication rows** — the first two rows are marked `Superseded` with dispositions linked to the later table.

Holistic re-check: topology, review counts, schemas, and approval authority unchanged; scope stays within `.agents/skills/dev-workflow/**` plus the task record; CI 9/9 green; no change file required (`.agents/skills/` is not a bundled skill). The new "PR #392 requested-changes adjudication" table records the dispositions accurately.

One non-blocking nit: `technical-design/final.md:22` reads "findings remain evidence-based and no finding is valid", which parses as "every finding is invalid" — the identity blocks use the clear "a no-finding result is valid" form; worth aligning the task record when this file is next touched.

Follow-ups outside this PR (for the operator, not blockers): the `FORBIDDEN_RE` guardrail gap remains — it matches `/home/<user>/` but not `/Users/`, so macOS home paths still pass pre-commit, CI, and the release gate. A separate change should add the macOS form.

PR #396, opened 2026-09-08
Unwrap Codex shell launchers in tool display

YourWildDad:
## Summary
- Show the inner script of recognized Bash/Zsh/Sh and PowerShell commands in Codex tool summaries and expanded invocations, following the pinned Codex TUI parsing rules. Preserve unknown or undecodable input and the Windows quoting round-trip rule.
- Keep parsing in the Codex display owner: raw arguments, execution, approval, action labels, redaction, and Feishu rendering remain unchanged. No new dependency or public API.
- Add display/runtime/Feishu regression coverage, rust-shlex attribution, Rush change notes, and the completed task and provider knowledge updates. The original workspace's unrelated Workflow commit is excluded.

Task: `.agents/tasks/channel/align-codex-command-display/README.md`
Upstream baseline: `openai/codex@49db349ffdd888f5e3c91abf9b7519d8631e6e9a`.

## Test plan
- [x] `node common/scripts/install-run-rush.js build`
- [x] `node common/scripts/install-run-rush.js lint`
- [x] `node common/scripts/install-run-rush.js test` — 150 files, 2,300 tests, no skips; includes 9 real Codex tests. The existing operator proxy was supplied through HTTP_PROXY/HTTPS_PROXY for live model access.
- [x] `node common/scripts/install-run-rush.js typecheck:tests`
- [x] `node common/scripts/install-run-rush.js change --verify --target-branch origin/next --no-fetch`
- [x] Task validator, `.agents/scripts/check.sh`, and `git diff --check`
- [x] Independent read-only review: no functional or architecture blocker; both accepted cleanup findings resolved with operator approval.
- [x] In-memory provider-to-Feishu payload probe for quoting, multiline scripts, unknown shells, malformed input, and Windows text.
- [ ] Feishu client visual verification was not performed; payload tests do not establish the rendered client result. No live service was restarted or deployed.

## Review scope
Only the command display fix and its tests/docs/change notes are included. Local gates were rerun on the isolated branch based on `next` after transferring the reviewed patch; the only transfer conflict was concurrent task-index additions, both retained.

### DIFF
--- a/.agents/domains/provider-runtime.md
+++ b/.agents/domains/provider-runtime.md
@@ -755,8 +755,13 @@ invocation; file tools are their path; searches their pattern or query;
 `Agent` its description with the prompt as invocation; `Skill` its name.
 Codex follows the TUI's own exec, patch and web-search cells: a uniformly
 parsed command is the files it read, the paths it listed, or `query in
-path`, otherwise the command's first line, always with the command as
-invocation; a patch is its paths with the diffs codex prepared as invocation;
+path`, otherwise the display command's first line.
+`/packages/agent-runtime/codex/src/tool-display.ts` decodes shell-word quoting
+and unwraps recognized Bash/Zsh/Sh and PowerShell launchers for display.
+Unrecognized or undecodable commands, including Windows drive-path text that
+fails Codex's quoting round-trip check, keep their original spelling. The full
+display command is the invocation; raw arguments stay unchanged.
+A patch is its paths with the diffs codex prepared as invocation;
 a `webSearch` item, which carries no tool name and was dropped before, is a
 `web_search` row. Anything outside those tables — every MCP tool — reports
 `null` and displays as its name. Core sanitizes both facts exactly as it does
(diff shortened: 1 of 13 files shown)


ryanxiang7 (approved):
Reviewed against the pinned upstream baseline (`openai/codex@49db349`) and the approved task record.

**Upstream fidelity.** The `splitShellWords`/`quoteShellWord` port matches rust-shlex 1.3.0 rule-for-rule (single/double quote tables, comment handling, malformed-input rejection, the Windows `:\` round-trip gate, file-stem shell detection, exact `[shell, -lc|-c, script]` Bash shape, and PowerShell flag ordering/case rules). Differential fuzzing against an independent oracle of the Rust sources (split/quote plus end-to-end display) produced no mismatches; the acceptance cases in the task record, including the upstream Windows snapshot, behave identically.

**Layering and scope.** Display-only decoding stays in the Codex provider's existing display owner; core projection, redaction, Feishu rendering, raw activity arguments, and other runtimes are untouched. No new dependency, exported API, or shared-core parser. The runtime test proves raw `arguments` remain the wrapped command on both started/completed while the display facts are decoded.

**Docs/changelog/safety.** `.agents` delta is accurate and current-state-only; the Codex `patch` note is correct for a display-only fix and the Feishu `none` note matches its test-only change; the rust-shlex MIT attribution is appropriate; the diff contains no internal identifiers. CI is 9/9 green on the current head.

Two non-blocking observations (no change required to merge):
- A decoded empty script (`bash -c ''`) now yields `invocation: ''`; downstream `nonEmpty`/`runtimeToolTitle` already collapse it to a bare row, so behavior is safe — a Feishu-level assertion pinning that row shape would be the only optional addition.
- The NUL-byte edge differs from upstream's `try_join` failure path, but every codex producer substitutes a marker string via `shlex_join` before this code sees the input, so it is unreachable; the task record documents the operator ruling to leave the branch out.

Approving.

PR #401, opened 2026-09-09
feat: show COT tool details and concise notification labels

YourWildDad:
Tool rows now show invocation or structured arguments before a labeled result area. Failed appears inside that result area instead of before the command. Automated completion and schedule inputs use concise labels on the Feishu card while the model still receives their complete bodies.

- Preserve fixed action titles and icons. Other rows use the runtime summary or complete tool name; remove the separate 80-byte name cap while retaining the existing START title fitting and event-size check.
- Prefer invocation, then formatted JSON/text arguments, with no ARGUMENTS heading. Argument and invocation values bypass masking and path rewriting under the approved display policy; other payload fields keep their existing redaction rules.
- Emit a separate RESULT/divider segment before failure text and output. Rows with items remain list-only, including failed calls; existing whitespace, line limits, and list-overflow behavior remain.
- Carry completion kind and producer through the existing input projection for TEAMMATE CALLBACK, WORKFLOW FINISHED, CRON TRIGGERED, and SYSTEM RESTARTED labels. No provider language-classification contract is added.
- Forward Codex web-search query/action as arguments. Add the requested workflow clarification rule and concise TeamMate briefing examples.

Validation: Rush build, lint, test (2418 passed), typecheck:tests, and smoke-built-cli pass. Skill validation, KB checks, git diff --check, mandatory pre-commit scans, and Rush change verification against origin/next pass. Fresh Feishu client probing has not been performed for this implementation. [GitHub CI](https://github.com/excitedjs/dreamux/actions/runs/34393066051) passed all 9 checks, including Linux and macOS build/test jobs. Final external review is pending.

[Requirement and operator decisions](.agents/tasks/channel/refine-cot-tool-details-and-notifications/requirement.md) · [Verification](.agents/tasks/channel/refine-cot-tool-details-and-notifications/verification.md)

Alpha: `@excitedjs/dreamux@0.25.0-alpha.g3180b18cd009`, published from this implementation by [release run 34393391714](https://github.com/excitedjs/dreamux/actions/runs/34393391714).

### DIFF
--- a/.agents/domains/channel.md
+++ b/.agents/domains/channel.md
@@ -830,15 +830,15 @@ an event changes only that catalog and its consumers.
 No event carries a turn identity: presentation correlation is the `source_id` a
 caller supplied, echoed back on its own input, and nothing exposes a
 runtime-native Turn object or transcript. Conversation events may contain
-redacted user/assistant display text and redacted tool arguments/results, whole:
-Core bounds nothing, and the Channel cuts a string only where it would exceed
-Feishu's per-event limit (「core那边只做脱敏，不做截断 … Channel这边先去解析JSON，
-然后在发送接口之前去做截断」, 2026-09-04). Redaction keeps a structured payload
-parseable — a quoted secret becomes the quoted string `<redacted>`, and a bare
-one stops at the quote or bracket that closes it — because the Channel parses
-`result_json` to decide how to show it. Other
-events contain no prompt or assistant text. No event contains native transcript
-paths, raw errors, or platform user identity.
+redacted user/assistant display text and redacted tool results, whole. Argument
+payloads and invocation strings are the operator-selected exception: Core
+serializes arguments and preserves invocation text without secret masking or path
+rewriting ("参数全给我放开，不要做脱敏了", 2026-09-09; see
+[the task record](../tasks/channel/refine-cot-tool-details-and-notifications/README.md)).
+Core bounds nothing; the Channel parses JSON and cuts a string only at its native
+send limit. Other event metadata does not expose a runtime transcript or its
+path. Raw arguments can contain the values submitted to a tool; they are not
+covered by the redacted-result/body guarantee.
 
 Delivery is live and best-effort: Core invokes listeners in publication order
 without awaiting them, and a listener's exception or rejection never escapes into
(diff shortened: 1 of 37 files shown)


ryanxiang7 (approved):
## Final review: APPROVE

Read the full diff against the stated baseline `00b858e`, traced every changed
seam through the layer above it, and cross-checked the requirement, the final
technical design, and the R1–R7 dispositions. CI is 9/9 green on `b70d224`.

### Verified against the ruled behavior

- **Invocation-first, unredacted arguments.** `toolArguments`
  (`packages/channel/feishu-channel/src/feishu-cot-presentation.ts`) selects a
  nonempty `invocation` for every tool before `arguments_json`; only an empty
  invocation falls back to the JSON/text payload. Core
  (`packages/dreamux/src/channel/conversation-projection.ts`) now crosses
  `invocation` and `arguments_json` verbatim while summary, items, and result
  keep their existing redaction. The new privacy test pins both sides of that
  boundary: the same secret-shaped value is byte-identical in arguments but
  `<redacted>` in the result. The exception is field-specific, not a global
  redaction removal, and no config flag was introduced.
- **Title/name and icons.** The runtime summary is the title as-is; an untitled
  row sends the tool's own name; action icons are unchanged. Action-verb
  composition and the name-leaf truncation are gone, per ruling.
- **List-only.** A call with items renders the list alone for every status,
  including a failed call — the old `status !== 'failed'` branch is removed and
  both success/failure list rows are tested. The pre-existing list budget and
  many-short-items fallback are untouched (R2, explicitly deferred).
- **RESULT as its own segment.** `RESULT\n\n---` is a separate text segment
  emitted only when there is output, immediately before the unchanged
  JSON/text result; the blank line prevents a setext heading. JSON vs text
  classification, the ten-line text rule, `preserveSpacing`, and native event
  order are preserved.
- **Byte fitting (R7).** The three unreachable whole-payload/status fallbacks
  are gone; the assembler keeps only result-then-argument string fitting. I
  worked through the budget math: after the two cuts content is at least 32
  bytes under `maxBytes`, labels are always emitted with their payload (no
  orphan labels), fixed labels plus markers are far under the budget, and the
  outer 4096-byte check still falls back to a status word. "Two cuts suffice"
  holds.
- **Automated-input labels.** `TeammateInputCompletion` is a minimal neutral
  contract (producer kind, plus a TeamMate's name); it rides the existing
  admitted-input path and is never read by `renderSubmission`, so the model
  still receives the full notification body. The producer name is redacted
  like any display string; the kind is Core's own word. `inputDisplayText`
  maps only structured facts (completion kind/name, the existing
  `cron`/`system` source names) and never parses the body. Workflow carries no
  name, per ruling.
- **R4 web-search mapping.** Codex `argumentsFor` projects the native
  `{query, action}` through the existing argument seam inside the provider;
  pairing and cache behavior are unchanged, and nothing is inferred from the
  summary.
- **Layering / no glue.** The neutral `invocationLanguage` fact and the
  completion metadata live in `dreamux-types` with no channel/runtime leakage;
  protocol-specific extraction (bash/powershell/javascript/diff, webSearch)
  stays in the providers. No magic-string prefixes, no prompt smuggling, no
  state re-derived in Core. KB delta (domains, product README, task record)
  matches the final code; change files are all `minor` on the 0.x line with no
  `BREAKING`, and there is no config/state shape change, so no maintenance-skill
  sync is due.

### Non-blocking nits (no change required to merge)

1. `packages/channel/feishu-channel/tests/feishu-cot-tool-rows.test.ts` — the
   baseline asserted `role: 'tool'` on a `TOOL_CALL_RESULT` event (old
   `:161`); after the rewrite no test covers that wire field, though the source
   still sets it (`feishu-cot-events.ts:169,183`). A one-line `toMatchObject`
   would restore the coverage.
2. `.agents/domains/channel.md` — the current-state domain page inlines the
   task ruling ids "R2"/"R3" in two spots; the section already ends with a
   proper History pointer, so dropping the bare ids would match the
   current-state-only convention more closely. The same rewrite also no longer
   states the still-true "no raw errors / no platform user identity" event
   invariants (they hold in code: errors cross as redacted result text, and
   `source_id` is a Feishu message id, not an open_id).

Neither affects correctness, the redaction boundary, or the card contract.

Happy for this to proceed to Alpha.

ryanxiang7 (approved):
Re-reviewed the reset re-implementation at `3180b18` (base `7ed1d886`) against the restated requirements and verification record, with the focus on code complexity / entropy. **No blocking findings.** This version removes more mechanism than it adds and lands every ruled behavior on the existing Runtime → Core → Channel seams.

Complexity / entropy:

- The inferred `invocationLanguage` cross-layer contract is gone — from the neutral types, both runtimes' tool-display, and the Channel. Code-segment language is the Channel's existing action-based rule (`tool_action === 'run'` ⇒ `bash`, else `text`), which is the one notation the card distinguishes; per-tool language detection no longer burdens a provider adapter.
- The baseline fixed action names and title words (Read / List / Search / Edit / Bash, with the action-verb prefix on the summary) are preserved as next already had them, and the 80-byte tool-name cap is removed under the later explicit ruling while the START title fitting and the final per-event size check stay in place.
- Automated provenance is one neutral `notice` fact, constructed only on the completion-delivery path (`submitCompletionInput`); ordinary Channel/cron submissions never set it, so the one-line label cannot be spoofed from the open submit surface. The producer name crosses verbatim because `TEAMMATE_NAME_PATTERN` admits no separator, colon, or space a redaction rule could match — which also deletes the projection-side producer-redaction step entirely.
- The `jsonText` contract now states the honest guarantee: whole serialization, no post-redaction parsability promise; the display already treats an unparsable payload as text.

Ruled behaviors verified in code and tests:

- A nonempty `invocation` always wins over `arguments_json` for every tool; both argument members cross Core verbatim while summary, items, result, and input/assistant bodies keep their existing redaction. Privacy tests add home-path and secret-shaped-argument verbatim assertions rather than weakening any.
- Items produce a list-only row for every status, including failed; list budgeting and the many-short-items status-word fallback are unchanged.
- `RESULT\n\n---` is its own text segment; on failed non-list rows the order is argument code → RESULT label → Failed → output, and a failed row with no output still gets the RESULT area. Exact segment order is asserted.
- Codex `webSearch` projects its `query`/`action` as arguments (absent members omitted rather than emitted as `null`); native pairing untouched.
- Automated inputs show one-line labels (TEAMMATE CALLBACK · producer / CRON TRIGGERED / WORKFLOW FINISHED / SYSTEM RESTARTED) from structured facts, never by parsing the notification body, and the model still receives the full body.

APPROVE.

PR #402, opened 2026-09-10
fix: route question answers from their actual card

YourWildDad:
Question cards now reply directly to the supplied message ID, including messages
absent from the session's observation ledger. Answers and expiry notices resolve
the actual card's chat/topic through Feishu message details and retain that card's
ID in the inbound envelope and presentation anchor. A failed message lookup is a
delivery failure instead of a guessed parent-chat route.

The question registry no longer stores a precomputed target or a duplicate expiry
message ID. Feishu message reads expose chat/topic metadata, and ordinary inbound
messages and card settlements share the existing chat-mode projection. Core,
runtime providers, persisted state and unrelated notification sending are unchanged.

Validation: Rush build, lint, test, typecheck:tests, built-CLI smoke, knowledge links,
change declarations and commit hooks. Regression coverage includes unobserved reply
IDs, no-ID creation, submitted/cancelled answers, consecutive rounds, expiry and
message-lookup failures. Existing router ledger tests were retained in their owning
suite. Rebased onto next `6bb8a713` before final verification.

Each settlement adds one message-details lookup. Thread-bearing cards also use the
existing cached chat-mode lookup; its ordinary-inbound fallback behavior is
unchanged. Message details omit chat type, so readback uses the existing group
default when no topic is resolved; unbound direct-chat answers still reach the
Dispatcher. Live Feishu validation is recorded below.

DevBox review requested by the operator; the earlier approval covered the
superseded minimal patch.

Task: [requirements and implementation record](https://github.com/excitedjs/dreamux/blob/fix/feishu-question-card-routing/.agents/tasks/channel/fix-question-card-topic-routing/README.md).

CI: [all 9 checks passed](https://github.com/excitedjs/dreamux/actions/runs/34435750357)
on `94ee3505778a1b37756cb0edb5b308b06c5f110e`, including Linux and macOS gates.

Alpha published from `94ee3505778a1b37756cb0edb5b308b06c5f110e`:
`@excitedjs/dreamux@0.25.0-alpha.g94ee3505778a`
([release run](https://github.com/excitedjs/dreamux/actions/runs/34436982446),
[tarball](https://registry.npmjs.org/@excitedjs/dreamux/-/dreamux-0.25.0-alpha.g94ee3505778a.tgz)).
Registry metadata and downloaded tarball SHA-512 integrity were verified.

```sh
npm install -g @excitedjs/dreamux@0.25.0-alpha.g94ee3505778a
```

Live Feishu validation passed on `0.25.0-alpha.g94ee3505778a`: after the operator
upgraded, the first question was answered, and the second question was sent using
exactly the first settlement envelope's `message_id`. Both answer envelopes
retained the original topic, and the operator confirmed that the second card
rendered in that same topic. The two-card test is complete.

### DIFF
--- a/.agents/domains/channel.md
+++ b/.agents/domains/channel.md
@@ -165,20 +165,22 @@ field descriptions, so a model reads the two as one tool: 1-4 `questions`, each
 with a `header` chip, a `question`, and 2-4 `options` of `label` +
 `description`. Four fields differ, and two of them are the chat. A `chat_id` is
 required, because a chat tool needs a destination and AskUserQuestion has no
-such concept; a `message_id` is optional, and when the model names the message
-its question came out of, the card is addressed the way a reply is — into that
-message's topic, under the anchor the target router already holds for it. With
-no message named, the card is addressed at the chat, which in a topic group
-opens a topic of its own: right for a question belonging to no particular
-message, wrong for one that does. A message id from another chat is ignored
-rather than obeyed, the same rule `reply` follows, so a stale id cannot
-redirect a question into a conversation it was not meant for.
+such concept; a `message_id` is optional. When supplied, that exact id addresses
+an interactive-card reply, just as it does for `reply`, without an observed-message
+lookup or anchor substitution. Without it the card is created in the named chat;
+in a topic group it opens a topic of its own.
 
-After a successful question-card send, the session records the returned message
-ids against the question's target in the existing observed-message ledger. The
-answer envelope and presentation anchor carry the answered card's message id.
-A subsequent question using that id therefore inherits the same topic; it
-does not substitute the earlier human message or the round's deduplication id.
+Question rounds store their sent card id, not a precomputed routing target. On
+submission or cancellation, the card callback identifies the answered card. Feishu
+message lookup supplies its actual chat and optional thread id for existing
+chat-mode and binding routing. Expiry resolves the sent card id through the same
+path. The answer envelope and presentation anchor carry that card id, never the
+earlier human message or the round's deduplication id. A failed lookup is logged
+as a delivery failure; it is not a reason to route to a guessed parent chat.
+
+This adds one message query per settlement. The existing Feishu message reader
+exposes chat/topic metadata alongside content; Core and runtime providers acquire
+no Feishu-specific fields or routing responsibilities.
 
 The other two fields are gone. There is no `multiSelect`: the operator ruled
 multi-select out of this channel, so every question takes exactly one answer.
--- a/.agents/domains/channel.md
+++ b/.agents/domains/channel.md
@@ -174,6 +174,12 @@ message, wrong for one that does. A message id from another chat is ignored
 rather than obeyed, the same rule `reply` follows, so a stale id cannot
 redirect a question into a conversation it was not meant for.
 
+After a successful question-card send, the session records the returned message
+ids against the question's target in the existing observed-message ledger. The
+answer envelope and presentation anchor carry the answered card's message id.
+A subsequent question using that id therefore inherits the same topic; it
+does not substitute the earlier human message or the round's deduplication id.
+
 The other two fields are gone. There is no `multiSelect`: the operator ruled
 multi-select out of this channel, so every question takes exactly one answer.
 And there is no `preview`: it was first drawn as a column beside the options,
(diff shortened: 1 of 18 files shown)


ryanxiang7 (approved):
Final review of `1e01198` (base `next`, stacked on #401). **No blocking findings — APPROVE.**

The fix is one registration loop in `askUserQuestion` (`feishu-session-ops.ts:478`): after a question card is sent, each returned message id is recorded against the resolved target via the existing `targetRouter.observe(...)`.

Why it is correct and minimal:

- **Root cause is real.** A follow-up question addressed to an answered card was routed by `outboundTarget(chatId, messageId)`. Question-card sends were never entered into the observed-message ledger, so that lookup missed and the new card fell back to the chat root — opening a new topic in a topic group. Recording the send makes the card id addressable exactly the way every inbound and notification message already is.
- **Reuses the existing mechanism; no new concept.** `observe` is the same call `projectInbound`, `sendReply` (team-leader path), and `onNotificationSent` already use, and it populates both ledgers a reply needs (message→target and the target's newest anchor). No new field, branch, or indirection.
- **Target consistency holds.** The card is physically sent to `notificationTarget(target)` but recorded against `target` — identical to `onNotificationSent`'s established convention. When `target` is a topic, its anchor already exists (established by the inbound `observe` that routed the triggering message), so there is no case where a message recorded as a topic was actually delivered to the chat root. A `messageId` from another chat is still ignored by `outboundTarget`, so the stale-id guard is unchanged.
- **Answer identity preserved.** The envelope and presentation anchor still carry the actual answered card id (`settlement.cardMessageId`); the fix only adds routing registration and does not touch that id. `observe('')` is a no-op, so an unknown card id registers nothing.

Validation: the added two-round send/click/submit regression drives the real registry, session ops, and router, asserting both the card destination (`{chatId, replyToMessageId}` in-topic) and the answer identity (`thread_id` + answered `message_id` in attrs, anchor pointing at the answered card). It fails on the second round's missing anchor before the fix. Changelog is `patch` for the 1.0+ `@excitedjs/feishu-channel` package with no `BREAKING`, and the `channel.md` delta states the routing rule accurately. CI 9/9 green.

ryanxiang7 (approved):
Final review of the reworked approach at `94ee350` (rebased onto latest `next`, which now contains #401 and #403). Reviewed the full PR diff. **No blocking findings — APPROVE.** This rework is simpler than the earlier one-line-observe version and removes a mechanism rather than adding one.

What changed in the approach:

- **Send**: the question card is addressed exactly like `reply` — an explicit `messageId` sends it as `replyTo` that message (`feishu-session-ops.ts`); with none, it is a new message in the chat. No precomputed target and no send-time ledger registration.
- **Settle**: a submitted, dismissed, or expired answer is routed from the card it belongs to. `askUserCardRoute` reads the card back with `im.v1.message.get`, takes the reported `chat_id`/`thread_id`, and resolves the target through the router's chat-mode logic. `round.target`, `AskUserOpenInput.target`, and the expiry's separate `messageId` are all deleted; the registry no longer holds routing state.

Requirement consistency and routing correctness:

- The read-back is ground truth. `thread_id` is the same field inbound parsing already uses (`parse/content.ts`), so a topic answer cannot be misrouted by a stale or missing local observation — the class of "observation miss redirects a reply" bug the prior version patched around is removed, not papered over.
- Chat-mode resolution is reused, not duplicated: `projectInbound` now delegates to a new `project({chatId, threadId, chatType?})` that holds the one topic-vs-reply-chain decision. A message read back reports no `chat_type`; it defaults to `group`, whose only behavioral difference is bindability (`isBindableTarget`), and a p2p chat id can have no binding row, so routing still falls to the Dispatcher. The code comments state this reasoning explicitly.
- Failure semantics are correct and tested: if the card cannot be located (read throws, or no chat reported), the answer is delivered **nowhere** rather than to the chat the question happened to be asked from — a guessed parent chat would show one conversation's answer in another. The expiry card repaint is independent and still happens. No retry; consistent with the existing "log and drop, the user can say it again" delivery policy.
- Send-fail safety is preserved: the round (and its TTL) is armed only inside `activate()`, called after `sendCard` resolves, so a failed send leaves no question whose timer would later report an unanswered question.

Complexity:

- Net concept reduction: the round no longer caches a target that could disagree with reality; one authoritative read at delivery replaces both the cached target and the send-time `observe` for this flow. The ordinary `reply` ledger path is untouched.
- Transport gains two projected fields (`chatId`, `threadId`) on the message-read item — additive, the same source the inbound event uses.

Tests cover consecutive answers staying in-topic, ordinary group delivery, dismissal routed from its card, expiry, and both lookup-failure modes (read error / empty result → zero deliveries, repaint still occurs), plus router ledger behavior and the transport projection. Changelogs are `patch` for `@excitedjs/feishu-channel` and `minor` (additive read fields) for `@excitedjs/feishu-transport`, no `BREAKING`. CI 9/9 green.

PR #403, opened 2026-09-10
Align COT tool results and remove the RESULT heading

YourWildDad:
Tool calls now separate their arguments and results with a horizontal divider, without a RESULT heading. Calls with actual output show it directly below the divider; calls without output show Complete or Failed, including calls that have arguments but no output.

The change stays in Feishu result assembly. List-only rendering, output formatting and truncation, event-size fallback, and runtime status facts retain their behavior. The task and product records capture the operator's final display rule.

Validation: Rush build, lint, full test (including real Codex), typecheck:tests, knowledge checks, and change-file verification passed. Regression coverage includes text/JSON output, both no-output outcomes with and without arguments, and the existing byte-budget and list cases.

Independent review approved both the initial status change and the final heading removal (implementation commit cc3904cf). All nine CI checks passed on that implementation. Subsequent commits only close the task records. No service installation or restart was performed.


ryanxiang7 (approved):
Approved.

The change replaces three non-list result shapes (a Failed line before output, a conditional RESULT header, and bare Completed / args-alone for empty success) with one uniform structure: arguments when present, the RESULT header and divider, then actual output when it exists or Complete/Failed when it does not — matching the operator's final display rule.

Read-only verification against head 85c34b7a:

- **Logic**: `toolResultSegments` covers all six outcome × output × arguments cases. Status comes from `event.status`, never inferred from output text. List rows bypass this path and the event-size fallback is unchanged; the non-list path still converges far inside the 4,096-byte budget (minimum content is the header plus one status word), so the fallback remains list-oversize-only. The fallback's bare status word is the deliberately preserved existing behavior.
- **Entropy**: net removal — conditional header branch, unconditional Failed line, empty-success special case, and the single-segment return optimization all deleted; no new mechanism, entity, or cross-layer hop. Change stays inside Channel presentation.
- **Tests**: rewritten assertions encode the new operator requirement rather than weakening a contract — failure with text/JSON output, both no-output outcomes with and without arguments, success with output, list rendering, list-oversize fallback, and dual truncation with the byte-budget assertion all remain in place.
- **Records**: `.agents/domains/channel.md`, the product catalog, and the task requirement/design/verification records describe exactly what the code does; change file is patch-typed and display-only, with no upgrade impact.

CI 9/9 green.

ryanxiang7 (approved):
Re-approved at cc3904cf.

The follow-up implements the operator's pre-merge refinement literally: the fixed segment between arguments and output is now `{type: "text", text: "\n\n---"}` — the RESULT word removed, the divider and its own segment kept. The output/no-output status rules approved at 85c34b7a are unchanged: actual output follows the divider directly with no status line for either outcome; without output, Complete or Failed fills the area whether or not arguments exist; list rows and the event-size fallback are untouched.

Read-only verification of the incremental diff (85c34b7a..cc3904cf, two commits):

- **Source**: `RESULT_HEADER` → `RESULT_DIVIDER` with the value `\n\n---`; comment and docblock updated to describe a divider instead of a label. No stale references; the constant's two call sites and the byte-fitting path behave identically (the segment is shorter than before, so budget convergence only improves).
- **Tests**: the shared expected segment constant moved with the code; every existing ordering/status/fallback assertion now expects the divider without a heading — contract updated for the new requirement, not weakened.
- **Records**: domain doc, product catalog, task README/requirement/design/verification all quote the operator's refinement verbatim and describe the divider-only segment accurately; the historical approval at 85c34b7a is recorded faithfully. Patch change file comment updated to match; still display-only with no upgrade impact.

CI 9/9 green on cc3904cf.

PR #405, opened 2026-09-10
Restore a subtle RESULT heading in Feishu COT

YourWildDad:
Removing the RESULT label in #403 made expanded Feishu tool results harder to distinguish from their arguments. Restore it as a sixth-level Markdown heading before the existing divider, using the style selected after native COT probes. Output/status rules and list-only rows retain their behavior.

Validation: Rush build, lint, test (including real Codex), and typecheck:tests; task and knowledge checks; existing text/JSON, empty-output, list, and event-budget coverage. The operator selected sample B in the Feishu desktop client; grey markup was rejected. No installation or service restart was performed.

Task: [.agents/tasks/channel/refine-cot-tool-details-and-notifications](.agents/tasks/channel/refine-cot-tool-details-and-notifications/README.md).


ryanxiang7 (approved):
Reviewed the full diff at 929f11a6 against `next`. The change is exactly the selected probe B and nothing more.

**Scope**
- Production change is one fixed segment in `feishu-cot-events.ts`: `'\n\n---'` → `'###### RESULT\n\n---'` (renamed `RESULT_DIVIDER` → `RESULT_HEADER`), still emitted as its own text segment ahead of the result area. No new mechanism, option, or cross-package surface.

**No regressions in the surrounding contracts**
- Text vs. JSON output selection and order are untouched; the header precedes either segment.
- Empty-output behavior is untouched: `Complete`/`Failed` still fills the area for every arguments/status combination. The four empty-output tests only swap the shared `RESULT` expectation constant; no ordering or status assertion was weakened.
- List-only rows take the independent `{ type: 'list' }` path and are not touched at all; the status-only overflow fallback is unchanged.
- Byte budget holds: the heading sits inside the measured content the existing shrink loop fits against (13 bytes of growth vs. the 256-byte reserve), and the long-argument + long-output test still asserts both truncation markers with the header in the middle and total content ≤ 4096 bytes. The `checkedEvent` gate remains the final guarantee.
- No stale references: no remaining `RESULT_DIVIDER` or old bare-divider literal in src, tests, or docs.

**Docs/changelog**
- Product catalog, channel domain page, and the task record (requirement/design/verification/index state) consistently describe the sixth-level heading; the task record follows the current-amendment + preserved-history convention.
- Rush patch note matches a display-only label change — no config/state/path contract moves, so no breaking note is warranted.

CI is 9/9 green. Approving.

PR #412, opened 2026-09-11
Redact tool call invocations and arguments

YourWildDad:
## What

Two things, and the second is why the first took four commits.

1. The conversation projection ran three of a tool call's five payload members
   through the redactor and handed `invocation` and `arguments_json` to the
   Channel as the runtime wrote them. That exemption is deleted.
2. A tool call's structured payloads are now redacted by **walking the JSON
   value**, not by pattern-matching its serialization — and the whole redaction
   capability moved out of core into `@excitedjs/dreamux-utils`.

## Why the exemption went

It bought legibility — a masked command is a command nobody can judge — and
cost a gate. A token pasted into a shell command, or an agent writing a `.env`
through a tool's arguments, reached a chat surface unmasked, on the same card
whose result member was already being protected.

## Why walking, and not a better pattern

Widening redaction to `arguments_json` put `INLINE_SECRET_RE` in front of a
shape it had never been shown: a JSON string nested inside another. It failed,
and each fix exposed the next failure — the value of `TOKEN=\"abc\"`, then every
line of a multi-line `.env` but the first, then a redacted line swallowing the
lines below it, then a serialized backslash read as a boundary, then a nested
JSON *object* whose key no longer sits beside its separator, then a nested value
containing an escaped quote. The last two were found in review, not by me.

Six symptoms, one cause, and the cause is level rather than coverage. The same
payload leaks when serialized and is redacted correctly when raw:

```
{"content":"{\"client_secret\":\"shhSECRET\"}"}   → passes through
{"client_secret":"shhSECRET"}                      → {"client_secret":"<redacted>"}
```

A regex cannot count nesting depth, so every patch buys one shape and leaves the
next. `arguments` and `result` arrive as `JsonValue`, so `redactJson` walks them
before serialization: a field whose *name* says secret has its value destroyed,
every string leaf goes through `redactText`, numbers and booleans travel as they
are. Whatever parsed the structure already peeled one level of escaping, so the
text the pattern sees is always ordinary text. Depth stops mattering.

Three things follow. A payload that arrived as JSON is still JSON afterwards, so
the parsability caveat earlier revisions of this PR carried is gone. A
low-entropy or oddly-punctuated secret under a secret-named key is covered for
the first time. And **`INLINE_SECRET_RE` needed no change at all** — it is
`next`'s pattern, recomposed from the shared name list and otherwise untouched.
The extension made while chasing symptoms is reverted, which also closes the
scope question earlier revisions raised.

## The capability moved to utils

It is pure and depends on types alone, so core no longer owns rules it only
consumes; `conversation-projection.ts` drops from 387 lines to 227.

Its secret key names had been written out **three times** — the inline pattern,
`config-helpers.ts`, and `logger.ts` — and the copies had drifted. The logger's
was missing `api_key`, `private_key`, and `client_secret`, so those three were
never hidden in a host log. One list closes that gap with it. That is the one
behavior change here that reaches past `tool.call`, and it is deliberate.

## Scope

- `packages/dreamux-utils/src/redaction.ts` (new) — secret key names, the text
  rules, `isSecretKeyName`, `redactSecretKeyValues`, `redactJson`.
- `packages/dreamux/src/channel/conversation-projection.ts` — the `tool.call`
  branch; the rules it used to own now live in utils.
- `packages/dreamux/src/config/config-helpers.ts`, `src/platform/logger.ts` —
  call the shared capability instead of keeping their own key-name copy.
- `packages/dreamux-types/src/teammate.ts` — doc comments only; no field shape
  changes, so no persisted file, protocol frame, or config is affected.
- Tests: the rules' own cases move to `packages/dreamux-utils/tests/redaction.test.ts`
  with new `redactJson` coverage for every shape above; the projection suite
  keeps every projection-level case, with the three that locked the old
  exemption rewritten against the acceptance criteria. Two pinned surface tests
  gain the new module.
- `.agents/product/README.md`, `.agents/domains/channel.md`,
  `.agents/domains/current-architecture.md`.

## Verification

`rush build`, `rush lint`, `rush test`, `rush typecheck:tests` all pass.
`.agents/scripts/check.sh` clean.


packages/dreamux/tests/cot-projection-privacy.test.ts
```diff
@@ -424,15 +422,40 @@ describe('conversation projection: content visible after redaction is unchanged,
...
     expect(tool.kind === 'tool.call' && tool.invocation)
-      .toBe('post https://example.test with Bearer abcDEF123.ghiJKL456-_9');
+      .toBe('post https://example.test with Bearer <redacted>');
     expect(tool.kind === 'tool.call' && tool.arguments_json)
-      .toBe(JSON.stringify({ token: 'abc123secret', home: `${HOME}/keys` }));
-    // Nothing the redactor still owns fired, so the row is not marked redacted:
-    // the argument members never count toward it.
-    expect(tool.kind === 'tool.call' && tool.redacted).toBe(false);
+      .toBe(JSON.stringify({ token: '<redacted>', home: '~/keys' }));
+    expect(tool.kind === 'tool.call' && tool.redacted).toBe(true);
```
ryanxiang7, line 428: Acceptance criterion 4 requires `redacted: true` when the only rewrite is in either `invocation` or `arguments_json`, but this case rewrites both, and the dedicated case below isolates only `invocation`. If `args` were later removed from the aggregate at `conversation-projection.ts:239`, every current assertion would still pass because another member is redacted in each `redacted: true` fixture. Please add an arguments-only fixture (ordinary/null summary, invocation, items, and result) that asserts both the masked `arguments_json` and `redacted: true`.

ryanxiang7 (changes requested):
The production change is correctly layered and the rewritten assertions distinguish the old contract, but acceptance criterion 4 is not fully locked: the suite isolates invocation-only redaction, not arguments-only redaction. Because this is a privacy boundary and the requirement explicitly covers either member, please add that regression case.

packages/dreamux/src/channel/conversation-projection.ts
```diff
@@ -15,15 +15,32 @@ import type { AgentEntityIdentity } from '../service/agent-entity/types.js';
...
+ * The same four are a boundary at the key end too, where `\b` cannot see one:
+ * a newline serialized into JSON is the two characters `\n`, and its `n` is a
+ * word character, which would hide every line of a multi-line `.env` but the
+ * first. A key still needs a boundary, so `mytoken` is not one.
+ *
+ * Together these keep the shape *around* the secret intact, so a payload that
+ * arrived as JSON is still JSON afterwards and the Channel can parse it to
+ * decide how to show it, and a redacted line does not take the lines after it
+ * along.
  */
```
ryanxiang7, line 42: The nested alternative still leaks in two serialized shapes. First, an escaped quote inside the secret is mistaken for the lazy branch terminator: `arguments = { command: "echo TOKEN=\"prefix\\\"TOP-SECRET-SUFFIX\"" }` serializes to `TOKEN=\"prefix\\\"TOP-SECRET-SUFFIX\"`, but projects as `TOKEN=\"<redacted>\"TOP-SECRET-SUFFIX\"`; the suffix remains visible and the JSON still parses, so Feishu displays it. Second, `arguments = { content: "{\"token\":\"sk-live-value\"}" }` is returned completely unchanged with `redacted: false`, because the nested key is `\"token\"` while the key branch only accepts bare optional quotes. These are outside the documented raw `\n`/`\r`/`\t` tradeoff and directly violate criterion 2. Please parse/redact string leaves structurally before re-serialization, or make both nested keys and delimiters escape-depth aware, and add both cases as exact regressions.

.agents/tasks/channel/redact-tool-call-arguments/technical-design/final.md
```diff
@@ -0,0 +1,84 @@
...
+Two places: the projection branch that decided who gets redacted, and the
+secret pattern that had never been shown a tool argument.
+`projectedActivity()`'s `tool.call` case in
+`packages/dreamux/src/channel/conversation-projection.ts` ran three of its five
+payload members through `redactText` and handed `invocation` and
+`arguments_json` to the Channel untouched. The exemption is deleted: the branch
+now redacts all five through one local `redact(text: string | null)` helper, and
+`redacted` is the OR over every member instead of the three that used to count.
+
+Nothing else moves. The five secret patterns, the path-prefix renaming, and
```
ryanxiang7, line 18: This current final design still says all five secret patterns and `redactText` are unchanged, but the same document later explains that `INLINE_SECRET_RE` changed and the requirement now marks that scope extension as pending operator confirmation. Please remove or amend this stale statement so the closeout record has one truthful shipped boundary.

ryanxiang7 (changes requested):
F1 is closed, and the new tests correctly lock the basic nested-string and multiline cases. However, the rewritten security regex still leaks secrets in two common serialized shapes, and the requirement expansion is explicitly still awaiting operator confirmation. Please fix and regress the concrete counterexamples inline before this privacy boundary ships.

packages/dreamux/src/channel/conversation-projection.ts
```diff
@@ -15,15 +15,32 @@ import type { AgentEntityIdentity } from '../service/agent-entity/types.js';
...
-const INLINE_SECRET_RE = /(["']?\b(?:secret|password|passwd|token|authorization|cookie|credential|api[_-]?key|private[_-]?key|client[_-]?secret)\b["']?)(\s*[:=]\s*)("(?:[^"\\]|\\.)*"|'[^']*'|`[^`]*`|[^\s,;"'`)\]}]+)/giu;
+const INLINE_SECRET_RE = /(["']?(?:\b|(?<=\\[nrt]))(?:secret|password|passwd|token|authorization|cookie|credential|api[_-]?key|private[_-]?key|client[_-]?secret)\b["']?)(\s*[:=]\s*)(\\"(?:[^"\\]|\\.)*?\\"|"(?:[^"\\]|\\.)*"|'[^']*'|`[^`]*`|(?:\\\\|(?!\\["nrt])[^\s,;"'`)\]}])+)/giu;
```
ryanxiang7: **F1 — a nested JSON *object* leaks; only the unquoted-key `.env` form is covered.**

The value side learned the opening `\\"`, but the key side did not. In serialized compact JSON the text is `\\"client_secret\\":\\"shhSECRET\\"`. The key group matches `"client_secret` (the leading `["']?` eats only the bare `"`, leaving the `\`), so the next characters before the separator are `\` + `"`, not `:`. Separator group 2 needs `\s*[:=]` immediately, the match fails entirely, and the secret passes through untouched.

Repro (all are the `arguments_json` path, outer JSON.parse succeeds in, and the value survives):

```js
JSON.stringify({ body: JSON.stringify({ client_id: 'x', client_secret: 'shhSECRET' }) })
// -> {"body":"{\"client_id\":\"x\",\"client_secret\":\"shhSECRET\"}"}  // shhSECRET survives, redacted:false
JSON.stringify({ file_path: 'config.json', content: '{"token":"abcSECRET"}' })
JSON.stringify({ body: '{"token": "spcSECRET"}' })
JSON.stringify({ content: '[{"token":"abcSECRET"}]' })
```

I confirmed against `next` that this specific shape missed before too — but this PR's stated purpose is precisely \"recognize secrets written inside a nested JSON string,\" and acceptance criterion 2 names `arguments_json`. A `Write` of a JSON config or a POST whose body is a stringified JSON object is the most likely real instance of that, more so than the `.env` line. The new tests only exercise the unquoted-key `.env`/`export` form, so they stay green while this misses — the \"green is not evidence\" caveat applies here again.

Minimal fix, verified: let the key end carry an optional escaped quote before group 2 — change the key tail `\b["']?)` to `\b(?:\\")?["']?)`. I ran that against the full 31-shape corpus (plain JSON, nested `.env` LF/CRLF, single/backtick quotes, `C:\\name`, values ending in `\`, multi-field objects, arrays, number/boolean values): it closes 8 nested-object shapes, introduces zero new misses, and every output that parsed before still parses. Please add the compact-object cases above as arguments-only tests that fail on the current head (deleting the key-tail allowance must turn them red).

ryanxiang7: **F2 — an escaped quote *inside* a nested value closes it early and breaks JSON parsability.**

The nested value alternative `\\"(?:[^"\\]|\\.)*?\\"` is non-greedy and treats the first `\\"` as the closing quote, with no pairing against an escaped backslash. Repro:

```js
JSON.stringify({ content: 'TOKEN="ab\\"cdSECRET"' })
// content text is a .env line whose value contains an escaped quote:
// actual text -> TOKEN="ab\"cdSECRET"
// redacted   -> TOKEN=\"<redacted>\"cdSECRET\"
```

`cdSECRET` survives **and** the outer `JSON.parse` of the projected `arguments_json` throws (I verified true→false), so the Channel falls back to plain text — that contradicts this block's own new promise that \"a payload that arrived as JSON is still JSON afterwards.\" The `.env` value `"ab\"cd…"` is a normal escaped value, not a malformed one.

The closing `\\"` has to be distinguished from an interior one: in serialized text a real close is a lone `\\"` (the quote's backslash is not itself escaped), while an interior quote arrives as `\\\\\"` (a `\\\\` pair immediately before the `"`). The value body should consume escaped pairs as a unit and only close on an unescaped `\\"`. Please make this shape redact whole *and* stay parseable, and cover it with a test (it should assert both no `cdSECRET` and `JSON.parse` success).

ryanxiang7 (changes requested):
_Duplicate of the earlier CHANGES_REQUESTED review (#5176583494) submitted 48 minutes earlier on this same commit by the Devbox review seat. Its inline comments (the same two findings, independently re-derived) have been removed to avoid double noise. Please treat review #5176583494 as the authoritative one._

packages/dreamux-utils/src/redaction.ts
```diff
@@ -0,0 +1,282 @@
...
+  let redacted = false;
+  const walk = (node: JsonValue): JsonValue => {
+    if (typeof node === 'string') {
+      const result = redactText(node, cwd, homePaths);
+      redacted ||= result.redacted;
+      return result.value;
+    }
+    if (Array.isArray(node)) return node.map(walk);
+    if (node === null || typeof node !== 'object') return node;
+    const out: { [key: string]: JsonValue } = {};
```
ryanxiang7: **Re-assessed 2026-09-11 after operator challenge — downgraded from blocker to non-blocking nit.**

The mechanism below is verified real (reproduced against the shipped module), but its severity in this path is low: this is a display-only projection, the failure over-*removes* content rather than leaking anything, it does not pollute the global prototype or throw (the swapped prototype lives only on the local `out`, and the event carries only `JSON.stringify` text), and real tool execution uses the original object. It also needs a structured argument/result that actually carries a `__proto__` *key* (a nested JSON document inside a string value is a string leaf and is unaffected). Keeping it as a hardening nit, not a merge gate.

~~Original claim, kept for the record:~~ When rebuilding an object, bracket assignment on the fresh literal invokes the legacy `Object.prototype.__proto__` setter for that key instead of creating a data property, so `redactJson(JSON.parse('{"__proto__":"keep-me","a":1}'))` projects `{"a":1}` (the verbatim serializer on `next` showed the field) and an object value replaces `out`'s own prototype. Reachability comes from `JSON.parse` creating `__proto__` as an own data property.

**Correction to my original comment:** the same-hazard claim against `redactSecretKeyValues` was wrong. It mutates in place, and assigning `__proto__` onto an object where it already exists as an own data property (the `JSON.parse` case) is an ordinary overwrite — verified it keeps the own property and writes `<redacted>`. Only the rebuild-into-a-fresh-object path in `redactJson` has the issue.

Optional hardening if desired (also reasonable as a fast follow-up, since `redactJson` is now a shared exported capability): `Object.defineProperty(out, key, { value, writable: true, enumerable: true, configurable: true })` plus one `__proto__`-key case in `redaction.test.ts`.

packages/dreamux-utils/src/redaction.ts
```diff
@@ -0,0 +1,282 @@
...
+  redacted = redacted.replace(BEARER_RE, 'Bearer <redacted>');
+  redacted = redacted.replace(JWT_RE, '<redacted-jwt>');
+  redacted = redacted.replace(COMMON_ACCESS_KEY_RE, '<redacted-access-key>');
+  redacted = redacted.replace(
+    INLINE_SECRET_RE,
+    (_match, key: string, separator: string, secret: string) => {
+      // A quoted secret stays a quoted (now empty of meaning) string, so the
+      // text around it keeps whatever grammar it had — JSON included. A nested
+      // JSON string is quoted by its escape, not by the bare character.
+      const quote = secret.startsWith('\\"')
```
ryanxiang7: **Nit — the `\`"` quote branch is dead code after the revert.**

Group 3 can only start with a literal `"`, `'`, backtick, or a bare-word character (which excludes `"`); no alternative can begin with backslash-quote, and the walk guarantees a string leaf never carries JSON serialization escapes, so a leaf containing `\"` holds two actual characters rather than an escaped quote. I probed 500k randomized matches of the shipped pattern and group 3 never once started with `\"`. This is the leftover hook from the round-2 nested-escape attempt that this commit otherwise fully reverts, and it sits oddly against the comment above the pattern saying the pattern deliberately does not understand nested JSON. Dropping it makes the "`next`'s pattern, untouched" claim literal.

ryanxiang7 (changes requested):
Round 3 review of the restructured change (109a358). The new shape is the right call: walking `JsonValue` removes the nesting problem instead of chasing it with regex escape rules, and consolidating the redaction capability in `@excitedjs/dreamux-utils` is a genuine entropy reduction — core goes 387→227 lines and only calls the capability. I independently confirmed the two claims this rewrite rests on:

- **Regex revert is byte-identical.** I extracted the actual `new RegExp(...)` construction from `redaction.ts` (no hand transcription) and compared it to the `next` literal: `source` and flags (`giu`) are exactly equal, and 500k randomized replace runs produced zero behavioral differences.
- **No cases were lost in the test split.** The 18 moved rule cases are byte-identical to their `next` bodies apart from the import rename; every retained projection case is still present; the 11 new utils cases cover the walk; the rewritten exemption cases are the operator-ordered contract change, not weakened assertions. The pinned surface additions (4 exports + 1 module representative in `index-exports.test.ts`, `./redaction.js` in `package-boundary-guards.test.ts`) extend the contract by the book.

One blocker on the walk, plus two non-blocking cleanups; details inline.

**Non-blocking cleanup trail:** a fourth copy of the key-name list survives at `packages/channel/feishu-transport/src/transport/diagnostics.ts:167` with the same drift (still missing `api_key`, `private_key`, `client_secret`). The requirement correctly scopes this PR to the projection/logger/config sites, and the transport keeps its dependencies deliberately minimal, so no change is requested here — but the final design's "written out three times" will read as exhaustive when a fourth copy exists; please open a follow-up issue/task record for it.

ryanxiang7 (approved):
**Approved.** Re-assessment after review of the round-3 finding against the actual Activity projection path:

The `__proto__` rebuild behavior is real (reproduced, comment on line 203 keeps the details and now records it as an optional hardening nit), but it is not a merge gate for this change: the projection is display-only and non-fatal by construction, the failure over-*removes* a rare field rather than leaking anything, it neither throws nor pollutes the global prototype, tool execution uses the original object, and reaching it needs a structured argument/result that literally carries a `__proto__` key. I also corrected a wrong aside in that comment: the in-place `redactSecretKeyValues` does not share the issue, because assigning onto an existing own `__proto__` data property is an ordinary overwrite.

Everything this rewrite was asked to prove still holds and was independently verified: the recomposed `INLINE_SECRET_RE` is byte-identical to the `next` literal (500k randomized replaces, zero diffs); the test split is lossless (18 rule cases moved byte-identical, exemption cases rewritten as the operator-ordered new contract, 11 new walk cases); the pinned surface tests are extended by the book; layering is a net improvement (one capability in utils, core only calls it); CI is 9/9.

Remaining items are all optional and non-blocking: (1) `Object.defineProperty` hardening + one `__proto__` case if the author wants the exported `redactJson` defensive now rather than in a fast follow-up; (2) remove the unreachable `\`"` replacer branch so "`next`'s pattern untouched" is literal; (3) a follow-up record for the fourth drifting key-name list in `feishu-transport/src/transport/diagnostics.ts:167`.

ryanxiang7 (approved):
**Approved.** Re-review of the three commits pushed after my earlier approval on `109a358` (`0cfd0175`, `0008e023`, `2cd904ee`). Every claim was independently verified, not taken from the descriptions:

**`0cfd0175` — utils cut loose from dreamux-types (the boundary change, examined first):**
- Both replacement declarations are structurally identical to the canonical contracts, compared byte-for-byte at HEAD: the local `JsonValue` equals `packages/dreamux-types/src/json.ts`, and `SettledInvoke<T>` equals `JsonInvokeResult<T>` in `invoke.ts` (the dropped default parameter is irrelevant). Cross-seam assignability is compiler-verified: e.g. `feishu-session-mcp.ts` returns `settleJsonInvoke`'s result directly into a `JsonInvokeResult`-shaped handler with no cast, and full typecheck is green.
- No `@excitedjs/` import remains anywhere in `dreamux-utils/src` (grep; only comment references). The `package.json` dependencies field is gone and the lockfile moved with it.
- The rewritten guard is strictly stronger than what it replaces. The old assertion pinned the manifest dependency set, which is exactly why a type-only crossing was invisible; the new one pins (1) no dependencies field and (2) zero `@excitedjs/` import lines in every `src` module, so a future `import type` crossing fails loudly. The existing vocabulary neutrality pin still holds against the rewritten doc comments.
- The `.agents` delta records the real rule (three-way dependency statement in `provider-runtime.md`, architecture table corrected, sources indexed), and the minor change note carries no `BREAKING:` — correct under the current definition, since no persisted state or manual migration is involved.

**`0008e023` — dead replacer branch:** I mechanically extracted both regex constructions (no hand transcription): `source` and flags (`giu`) are equal to the `next` literal, and 500k randomized replacements produce zero behavioral differences. The replacer callback is now character-identical to `next`'s. The unreachable fourth key-name list is recorded as an open boundary question rather than silently left — that closes the cleanup-trail item from the previous round.

**`2cd904ee` — own `__proto__` member:** the diagnosis and fix are both right. I reproduced the old behavior directly: an object-valued `__proto__` off the wire (`JSON.parse` via CreateDataProperty, own enumerable member) silently became the rebuilt object's *prototype* and vanished from its JSON; a string-valued one vanished outright. `Object.fromEntries` uses CreateDataPropertyOrThrow, so it is the correct construction primitive — own member retained, prototype untouched — rather than a defensive guard or key filter. Both new tests fail against the old loop and lock the string- and object-valued shapes, including the prototype assertion.

CI is 9/9 at this head. No blocking or non-blocking items outstanding.

PR #393, opened 2026-09-08
Teach the TeamLeader that a TeamMate is not a subagent

YourWildDad:
## Summary

The bundled `teamwork` skill now tells a TeamLeader what a member has that a subagent does not, a judgment of its own, and how to use it.

The skill's 2026-09-06 version answered "context, not control" by taking the edit plan out of the brief. PR #389 showed the shape that did not cover: no edit plan, but a frozen premise handed down as settled, with reviewers asked to find holes inside it. Each found a more precise hole; none was asked whether the premise held. This carries the principle PR #392 wrote into this repository's review prompts to every Team through the skill itself.

## Changes to `teamwork`

- Choice section: a member's own judgment is the thing a subagent lacks; a fifth question asks whether the work needs a judgment that can contradict the leader's framing; a member briefed like a subagent returns a subagent's result.
- Hand-down: the leader's design is a guess too; a new bullet puts the question the work turns on before the leader's answer.
- `identity` carries standing to contradict the leader's framing.
- New section "When Every Round Finds Another Hole": stop patching, write the one-sentence premise, put it to members that have not seen the draft; agreement among members that share a brief proves a shared premise.
- "Keeping the Thread": the reviewer gets the question, not the developer's answer, nor the leader's.

Name, description, and load trigger are unchanged. One patch change file for `@excitedjs/dreamux`.

## Verification

- Skill validator, Rush build, lint, test, `typecheck:tests`, `.agents/scripts/check.sh`, `git diff --check`.

## Records

- `.agents/tasks/builtin-skills/teamwork-teammates-are-not-subagents/`

Refs #389 and #392.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01TUY56nDnNaV3NX5nT2vFkg

PR #397, opened 2026-09-08
Show native token and context usage before runtime turns end

YourWildDad:
## Summary
- Append a display-only assistant usage line immediately before each native Codex or Claude Code turn ends, without changing completion text, failure reasons, or interruption semantics.
- Read native cumulative snapshots only: Codex context is a percentage; Claude context is the latest main-model input token count. Use consistent decimal `k`/`m`/`b` formatting with no rollout scan, extra query, or history ledger.
- Add runtime regression coverage, product/runtime knowledge, and patch release notes. The optional Claude interrupted outcome preserves custom sessions that supply no usage.

Example output:
```text
Context usage 50% | Token usage: total=28.6k input=28.6k output=69
Context usage 14.5k | Token usage: total=28.6k input=28.5k output=69
```

Claude totals are cumulative within the resident query/process, not a backfilled cross-restart session total. Missing totals omit the line; missing context alone shows `n/a`. This is an additive display enhancement with no configuration/state migration or rebuild requirement.

## Test plan
- [x] `node common/scripts/install-run-rush.js build`
- [x] `node common/scripts/install-run-rush.js lint`
- [x] `node common/scripts/install-run-rush.js typecheck`
- [x] `node common/scripts/install-run-rush.js typecheck:tests`
- [x] `node common/scripts/install-run-rush.js test --to @excitedjs/agent-runtime-codex --to @excitedjs/agent-runtime-claude-code`
- [x] `node common/scripts/install-run-rush.js smoke-built-cli`
- [x] `node common/scripts/install-run-rush.js change --verify --target-branch origin/next --no-fetch` after commit
- [x] Task-record check, `.agents/scripts/check.sh`, and `git diff --check`
- [x] Six deterministic built-runtime → Core projection → Feishu COT adapter cases: completed, failed, and interrupted for both runtimes; usage precedes the wire terminal.
- [x] One independent read-only review: APPROVE, no findings. TeamLeader then restored the approved optional-outcome contract and reran the affected gates.
- [x] Repository pre-commit hooks executed, including gitleaks and the internal-content gate; no hook bypass used for delivery.
- [ ] Full `node common/scripts/install-run-rush.js test`: **not green**. Two full runs fail the same five real-model Codex cases (continuity/resume, structured output, mid-turn activity, unbound native end, and mid-turn inbound folding). Codex 0.153.4 initializes but receives no model response; standalone native requests also time out. No live skip flag or weakened assertion was used.
[...]

PR #413, opened 2026-09-11
Pin the state a task record may carry on the trunk

YourWildDad:
## The problem

Task records kept reaching `next` saying `State: review` — a state that stopped being true the moment the branch carrying it merged. The operator reported it from repeated experience rather than one incident.

Measured on `next` at `27428f85`: **7 of 35 records** carried a state that cannot be true on the trunk, three using values outside the supported set entirely (`implemented`, `in-progress`, `delivery`). Every one had in fact been delivered.

The rules were not missing. `init_task.py check` already rejected four of these records. Nothing ran it, because it only ran when someone remembered to run it for the one task they were holding.

## The principle

A record is committed inside the pull request that delivers it, so it can never contain the result of its own merge. One test governs every line:

> Can something outside this repository make this false while the file sits untouched?

If yes, it is a snapshot, and git and GitHub already record it authoritatively, for free, and without drifting.

## What enforces it

- `init_task.py check-all` walks `.agents/tasks/**` instead of taking one named task. A README is classified by the headings only one kind carries — never by depth, since `mcp/scheduler` is a domain nested exactly where a task record would sit. Headings match as **whole lines**, and a file carrying both kinds is refused rather than resolved by precedence.
- Only `intake`, `blocked`, and `done` may reach the trunk. The six mid-workflow states stay correct on a branch behind `--allow-in-flight`; `.agents/scripts/check.sh` never passes it.
- The `State:` line is matched anchored, so trailing prose fails.
- Domain index entries no longer repeat the state. The same fact lived in two files for all 35 records and **three** copies already disagreed.
- **Delivery-status bullet labels are rejected.** The closed set is `- Baseline:`, `- Branch:`, `- CI:`, `- CI / merge:`, `- Commit:`, `- Current PR:`, `- Current next action:`, `- Current repair baseline:`, `- Current repair branch:`, `- Merge:`, `- Pull request / CI / merge:`, `- Pull request / merge:`. The rejection message prints that set from the same constant the check reads, so the documentation cannot drift away from it.

**No new CI job**: the existing `kb` job already runs `check.sh`, which now calls the checker as check 5.

## How staleness is checked without reading prose

The first version checked only the state line, and its stated limit was that a record could still describe finished work in the present tense. The reasoning behind accepting that limit was wrong.

It held that catching stale prose needs a phrase list, and that such a list cannot distinguish a live status from the same words quoted inside an operator ruling. Both true. The conclusion does not follow, because **upstream does not check prose at all**: `verify-agent-note-format.ts` bans a closed set of proposal-era *headings* inside an implemented note. A heading cannot occur inside a quotation, and rejecting one says nothing about the sentences under it.

The equivalent closed token here is the bullet label. A record does not say "CI is still running" in prose — it writes `- CI:` and then says it. Two upstream exemptions come with the mechanism: fenced blocks are stripped before structure is read, and a section under a `## Historical …` heading is skipped, for the reason upstream's gates skip `archived/`.

Residual gap, stated rather than implied: a stale sentence inside an otherwise live field is still not caught. The labels cover where every recurring status actually lives, so the residue is small, not zero.

## What review found, four times

The first two rounds found the same defect in different clothes: **an exemption a passing phrase can claim by accident.** Each was reproduced end to end, not argued about.

- `is_domain_index` matched headings as substrings. A record whose body contained `## Tasks completed` was read as a domain index, skipped the state rule, and passed carrying `State: review` — exit code 0, record count silently 36 → 35.
- The `## Historical` exemption was a prefix test. `## Historically speaking …` claimed it, and every line after that heading went unchecked — again exit 0. It now matches on a word boundary.

The sweep was wrong too, and in the direction that matters. Hand enumeration found nineteen stale records. Running the finished gate against `27428f85` rejects **23 of the 35** — 7 on the `State:` line and 23 on a delivery-status label, the seven being a subset. Grepping prose for status phrases misses things; that is the argument for structural tokens, made against my own work. A third round then found two merged pull requests still described as "targeting `next`"; a scan of every live line found a record whose `Pending approval` hashes were byte-identical to its `Approved` hashes; and grepping that phrase across every file — not just the `README.md` the gate reads — turned up one more, #339 in a sibling `verification.md`, merged a month earlier. All three are the same shape and none is a gate bug: `- Pull request:` is durable and correctly unbanned, and what expired was the prose after the label, which the rule deliberately does not read. `- Coverage:` names that limit, and now also names the file scope.

A fourth round then caught the worst one, which was mine: the commit that fixed #339 claimed it had corrected "the last stale delivery phrase" — a completeness claim about prose, made after grepping a single phrase, twelve lines above another expired line in the same section about the same pull request. That is the failure this pull request exists to prevent, committed by the pull request. Nine live lines across four task directories are now past-tensed in place, per the `CONTRIBUTING.md` rule this branch added: the fact is unchanged and only the tense made it read live, so there is nothing to correct, only to date. `- Coverage:` states the method — grep `.agents` for present-tense delivery phrases, read all 74 hits, keep the nine a merge could falsify — and says plainly that the method cannot prove it found everything. That is the argument for a closed label set, and the reason no commit message here says "the last" again.

One record was damaged mid-branch and repaired: a `## Historical` snapshot in `refine-cot-tool-details-and-notifications` had its frozen `Blockers` line rewritten into the present tense — rewriting a record of the past, which is the mirror image of the rule this PR adds. It is back in past tense.

## Records corrected

**24 records and 7 domain indexes**, and both numbers are measured rather than asserted — `verification.md` carries the script, which names `27428f85` rather than `origin/next` so it keeps meaning the pre-change trunk after this merges. 23 records are the ones the gate rejects. The twenty-fourth is the honest limit of the check and is named rather than glossed: `teamwork-teammates-are-not-subagents` said "Land the pull request after CI and review" under `- Next action:` and cited #393 as "CI and merge pending" after #393 had merged — both labels ordinarily durable, so the closed token set does not and should not include them. Each referenced pull request was verified against GitHub before its record was touched. Two records were false rather than stale: `remove-cron-run-now` said PR #335 was open and unmerged a month after it merged; `adopt-lean-self-upgrade-sop` was `done` while its delivery said the PR had not started.

Preserved because it cannot expire: operator rulings in the operator's own words, approval boundaries, incident accounts, review adjudications, and links to requirement/design/verification. Where a durable finding sat inside a line that had to go, it moved rather than dropped — including `minimize-provider-boundaries`'s outstanding Fable audit, restated as a statement of record.

## Borrowed from upstream

`technical-design/final.md` lists what was adopted from `deepseek-ai/deepseek-harness`'s `.agents/notes/` and what was declined, with reasons. Declined: dissolving the index (our only discovery mechanism), path-encoded lifecycle (would break every inbound link), dates in filenames, a sealed archive with a hash manifest, and the required section skeleton — upstream's note vocabulary is closed because it was mandated from the start; ours has about 140 distinct bullet labels across 36 records, most of them genuine one-off facts, so a skeleton would redden nearly every record for cosmetics. That figure is rounded on purpose: the argument needs its order of magnitude, and an exact count goes stale the next time anyone edits a record.

## Verification

`.agents/tasks/architecture/pin-task-record-states/verification.md` holds the evidence and the probes as runnable code.

- `.agents/scripts/check.sh` reported `Task records OK: 36 checked` and `KB OK (238 files reachable from root.md)`.
- `rush build` / `lint` / `test` / `typecheck:tests` — all four invoked, zero failures. No package source changed.
- The record counts — 7 state-rejected, 23 label-rejected, 23 in the union — and the label figure come from two scripts in `verification.md`, run verbatim. The first names `27428f85` rather than `origin/next`, so it keeps measuring the pre-change trunk after this merges. The remaining figures are not script output and are not claimed to be: the 24th record and the 7 domain indexes were counted from the diff, `36 checked` and `238 files` are what `check.sh` reported on this head, and 14 is the probe tally. The label figure is deliberately rounded in the prose: it moved 141 → 139 → 137 across this task, each drop caused by correcting a record, so the document states the sequence and points at the block rather than quoting a reading.
- **14 behavior probes**, executed verbatim from `verification.md`, 14 PASS. Each mutates a real record, runs the checker, asserts an exit code, and restores the file: mid-workflow state, trailing prose, the `## Tasks completed` bypass, both-heading refusal, fenced `## Tasks` inert, four banned labels, fenced label allowed, `## Historical` label allowed, `## Historically speaking` refused, and index state repetition.
- **The one gap this leaves open is written down, with its measurement.** Nothing but a person runs those probes: `check.sh` runs the gate over the tree and never exercises the gate's own rules. Reverting `FROZEN_SECTION_RE` to the prefix test — the fail-open shipped and fixed twice here — leaves `check.sh` at exit 0 while the probes report 13 PASS / 1 FAIL, naming the right one. The clean home is an `init_task.py self-test` subcommand on `check.sh`'s fifth check, needing no new CI job; it is deliberately not in this pull request, which the operator scoped to pinning the rules. `verification.md` also records what is *not* in that gap: the expired prose corrected by hand carries no banned label, so no self-test of a structural rule could see it — that class is held by the closed label set plus the stated limits, and teaching the gate to parse prose to close it would add the mechanism this design exists to avoid.


.agents/skills/dev-workflow/scripts/init_task.py
```diff
@@ -310,10 +327,120 @@ def check(args: argparse.Namespace) -> int:
...
+
+
+# A domain index and a task record are told apart by the headings only one of
+# them can carry. Depth cannot do it: a domain may nest inside a domain, so
+# `mcp/scheduler` sits exactly where a task record would.
+DOMAIN_HEADINGS = ("## Child Scopes", "## Tasks", "## Active Tasks")
+
+
+def is_domain_index(text: str) -> bool:
+    return any(f"\n{heading}" in text for heading in DOMAIN_HEADINGS)
```
ryanxiang7: **Blocking — the whole-tree gate can be silently bypassed, which is exactly the failure `check_all`'s docstring says it exists to close.**

`f"\\n{heading}" in text` is a substring test, not a line match, and classification runs before any state check. A task record that contains the substring anywhere is routed to `check_index` and never state-checked; `records` is not incremented, so the count drops and the gate exits 0. Reproduced at this head in a throwaway worktree:

- append `## Tasks completed` to a record and set `State: review` → `Task records OK: 35 checked`, exit 0
- put `## Child Scopes` / `## Tasks` inside a ```` ```markdown ```` example fence and set `State: review` → same, exit 0
- (the single-task `check` still fails — only the new tree-wide gate is blind)

The reachable case is not contrived: a record documenting the task system is precisely the kind that quotes an index shape, and nothing downstream asserts the printed count, so 36 → 35 passes CI unnoticed.

Suggested fix, verified to reclassify nothing at head (all 8 indexes carry zero `## Current state`/`- State:`, all 36 records carry both):

1. match headings as whole lines (e.g. `re.search(rf"^#{{2}} (?:{'|'.join(h[3:] for h in DOMAIN_HEADINGS)})$", text, re.M)`);
2. make the record discriminator positive — `## Current state` marks a record;
3. a README carrying both kinds of heading is an **error naming the file**, not silently an index; neither kind keeps today's fail-loud record check.

That removes the asymmetric direction: a false "record" already fails loudly; after this a false "index" can no longer skip silently.

.agents/tasks/architecture/pin-task-record-states/README.md
```diff
@@ -0,0 +1,41 @@
...
+- Status: Granted by the operator on 2026-09-11 (Feishu group), verbatim: "可以，那边贡献的人多，你先把规则定死了，剩下的状态起一个 workflow 去刷一遍。"
+- Approved implementation boundary: the rule and its enforcement in
+  `init_task.py` and `.agents/scripts/check.sh`, the dev-workflow references
+  that own the task-record format, and the existing task records and domain
+  indexes brought into compliance. No new CI job, and no change to the set of
+  nine workflow states.
+
+## Delivery
+
+- Pull request: Not opened.
```
ryanxiang7: **Blocking — the record that ships the rule violates it on its first line.** PR #413 has been open since 2026-09-11 12:51Z; this commit was authored at 12:50Z. Under the test this PR writes in `task-records.md` ("can something outside this repository make this false while the file sits untouched?") the line is false 34 seconds after it is typed, and the reference itself requires "the pull request link once one is open". Replace with `- Pull request: [#413](https://github.com/excitedjs/dreamux/pull/413).`

.agents/tasks/architecture/README.md
```diff
@@ -13,15 +13,16 @@
...
+- [Minimize Core Provider Boundaries](/.agents/tasks/architecture/minimize-provider-boundaries/README.md) — Reduce Agent Runtime and Channel contracts to minimal Command-invocation and Core-event ports without reducing Channel's external-interaction responsibility. Merged as PR #350 (+#353, #356); the task archive (requirement, final design, verification) is the decision record.
+- [Split streaming display from the push-back mechanism](/.agents/tasks/architecture/split-streaming-display-from-pushback/README.md) — Separate the conversation-display surface every channel-facing TeamLeader and Dispatcher needs from the submission push-back mechanism it is currently built on, so a display consumer no longer requires a Dreamux submission to exist.
+- [Unify Team command projections](/.agents/tasks/architecture/unify-team-command-projections/README.md) — One flat Team summary for create and status with lifecycle status as the single status meaning; list keeps a compact row with the same names. PR #390 awaits the operator's re-review of the fix commit.
```
ryanxiang7: **Blocking — delivery status left in an index entry this commit rewrites.** "#390 awaits the operator's re-review of the fix commit" is false: #390 merged 2026-09-08 as `2b3d697`. §4 of the final design says an index entry carries title/link/goal and never delivery status; the state token was stripped from this line in this commit's own hunk but the expiring sentence was left behind. Please delete it while you are in this hunk.

.agents/tasks/channel/README.md
```diff
@@ -16,14 +16,14 @@
...
+- [Feishu conversation-of-thought cards](/.agents/tasks/channel/feishu-cot-conversation-cards/README.md) — Render dispatcher and TeamLeader conversations as conversation-anchored Feishu COT cards through a neutral, display-only core projection.
+- [Refine Feishu COT and binding cards](/.agents/tasks/channel/refine-feishu-cot-and-binding-cards/README.md) — Bound plain-text COT tool results by line count, restore Team runtime context on binding cards, distinguish route-removal notifications, and complete superseded COT cards successfully
+- [Feishu slash commands](/.agents/tasks/channel/add-feishu-slash-commands/README.md) — Handle /stop, /teams, and /dissolve deterministically in the Feishu channel through one extensible command table
+- [Align Codex command display parsing](/.agents/tasks/channel/align-codex-command-display/README.md) — Show the inner shell script in Codex tool rows instead of the shell launcher wrapper
+- [Display turn usage summaries](/.agents/tasks/channel/display-turn-usage-summary/README.md) — Implement native context and cumulative token usage as the final assistant activity; draft-only delivery pending live Codex verification.
```
ryanxiang7: **Blocking — delivery status in a goal field.** "draft-only delivery pending live Codex verification" describes #397, which merged 2026-09-09 as `bf090bc`. The durable half of this fact (real-model Codex requests time out, so full live verification could not be green) is a permanent coverage limitation and is correctly recorded in the record's `Blockers`/`verification.md`; the delivery-pending half is a snapshot. Index goals cannot carry either half — please shorten to the goal.

.agents/tasks/architecture/pin-task-record-states/requirement.md
```diff
@@ -0,0 +1,101 @@
...
+
+The recurring problem: a task record merges into `next` still saying
+`State: review`, a state that can never be true once the pull request carrying
+it has merged.
+
+## Confirmed current behavior and evidence
+
+Measured on `next` at `27428f85`, 2026-09-11.
+
+- 7 of 36 task records carry a state that cannot be true on the trunk. Three
```
ryanxiang7: Non-blocking evidence accuracy: "7 of 36 records" does not match the measurement it cites. At `27428f85` there are 43 README.md files under `.agents/tasks`, of which 8 are domain indexes → **35 records** — which also contradicts "all 35 records" twelve lines below. The seven non-compliant values themselves check out exactly (`review` ×4, `implemented`, `in-progress`, `delivery`).

.agents/tasks/architecture/pin-task-record-states/technical-design/final.md
```diff
@@ -0,0 +1,92 @@
...
+They are correct on a working branch, which is a working copy. `review` is the
+true state of a task under review; deleting it would remove a fact the
+TeamLeader needs mid-task. The rule is about what may be *committed to the
+trunk*, not about what may be written, so `--allow-in-flight` keeps them usable
+where they are true and `check.sh` refuses them where they are not.
+
+## 4. Why the index no longer carries a state
+
+The state appeared twice for every task: once in the record and once in the
+domain index entry. Two of the 35 copies already disagreed. Nothing can resolve
```
ryanxiang7: Non-blocking evidence accuracy: there were **three** disagreeing copies at `27428f85`, not two — (1) this one, `feishu-cot-conversation-cards` (index `done` / record `in-progress`), and `refine-feishu-cot-and-binding-cards` (`review` / `done`); (3) `split-streaming-display-from-pushback` (index `intake` / record `implemented`). The third hides from a strict parse because its `State:` line carried trailing prose, which is itself part of the problem statement. The correction strengthens the argument.

.agents/tasks/architecture/pin-task-record-states/technical-design/final.md
```diff
@@ -0,0 +1,92 @@
...
+
+The operator's instruction was "尽量把上游好的方式全都借鉴过来" (2026-09-11). The
+upstream reference is the `.agents/notes/` system in `deepseek-ai/deepseek-harness`,
+which this repository's workflow already borrows from.
+
+| Upstream practice | How it lands here |
+|---|---|
+| A status whose values cannot go stale (`proposed` / `implemented` / `rejected`) | The same property, different shape: nine states stay, but only three may be committed to the trunk. Upstream notes have no working-branch phase to describe; our task records drive one. |
+| "The status carries no dates and no parentheticals" | The `State:` line is matched anchored, so trailing prose fails the check rather than being tolerated. |
+| No note records a PR number, merge commit, CI result, or merge date | The Delivery section keeps the pull request *link* and drops its *status*, the merge commit, the gate results, and the dates. |
```
ryanxiang7: Non-blocking: this upstream characterization is literally false and should be reworded to the (accurate) rule you actually borrowed. Upstream notes cite PR numbers and CI runs as historical evidence throughout — e.g. `.agents/notes/implemented/process/2026-09-09-blacksmith-failover-leg.md` cites closed PRs and `github.com/.../actions/runs/...` URLs, and code search for `PR #` under `.agents/notes` returns ~33 hits. The true upstream rule is: there is no delivery-*status field*, and the `Status:` line carries no dates or parentheticals. Your Delivery-section practice matches that; the table row as written overstates it.

.agents/tasks/channel/add-feishu-slash-commands/README.md
```diff
@@ -38,15 +37,9 @@
 - Commit: rebased onto `origin/next` at `2575e056` (after #380 and #382 merged
```
ryanxiang7: Nit: this survived the sweep in one of the swept files. "rebased onto `origin/next` at `2575e056` … The branch carries nothing else" is a rebase note plus branch-shape fact about a squash-merged PR whose branch is deleted — the exact category the commit message says left the records. One-line cleanup; the merge itself is already recorded by #379's link.

ryanxiang7 (changes requested):
## Request changes — CI 9/9 green; the gate's core guarantee and the sweep both have holes

Two independent review seats plus my own verification. The design is right, the deletions are clean (details below), but three things need fixing before this lands.

### Blocking

**1. The new gate can be silently bypassed — `init_task.py:368`.** Domain-index classification is a substring test that runs before state validation; a record carrying `## Tasks` in a prose heading or inside a fenced example is skipped entirely (`35 checked`, exit 0) even with `State: review`. That is the exact "unchecked record hides" failure `check_all` says it closes. Repro and a reclassification-neutral fix are on the inline comment.

**2. Expiring facts left in the surfaces this PR itself edits.** The PR's own record still says "Pull request: Not opened" while #413 is open; two index lines this commit rewrote state false delivery status (#390 merged 2026-09-08; #397 merged 2026-09-09); and one of the four prose records the second commit explicitly claims to sweep (`remove-cron-run-now`) keeps a Next action awaiting merge of #335, which merged 2026-08-15. All four verified against GitHub; details inline.

**3. The prose sweep is incomplete by five records.** Beyond the four named in the PR body, five more `done` records still assert pre-submission status for PRs that have since merged: #392 (`strengthen-independent-review-prompts`), #393 (`teamwork-teammates-are-not-subagents`), #396 (`align-codex-command-display`), #397 (`display-turn-usage-summary`), #405 (`refine-cot-tool-details-and-notifications`). The PR's approved boundary is "existing task records brought into compliance", and these are one-line rewrites each. Where a clause is a durable coverage limitation rather than a status (the real-model Codex timeouts in #397/#396), keep it — the inline comments say so explicitly.

### Non-blocking

- Evidence counts in the decision record are off: "7 of **36**" should be 35 records at `27428f85`; "**Two** of the 35 copies already disagreed" was three. The seven stale values themselves are exact.
- `final.md` §5 upstream row ("No note records a PR number, merge commit, CI result, or merge date") is factually wrong about deepseek-harness; reword to "no delivery-status field; the Status line carries no dates or parentheticals".
- `CONTRIBUTING.md`: reconcile the append-only "Since this was recorded" doctrine against the new rewrite-in-place rule, and add check 5 to the Validation paragraph's enumeration.
- Gate robustness, optional: the index guard only catches in-set state tokens (restoring a backticked `delivery` passes — documented trade-off, worth a code comment); a task directory containing `requirement.md` but no README is invisible to `rglob`; a non-UTF-8 file raises `UnicodeDecodeError` past `main()`'s `except (TaskError, OSError)` as a traceback.

### Answers to the three questions in the review request

1. **Did anything unrecoverable get deleted? No.** Independently audited: all 11 links appearing on removed lines survive at head (scripted), the 2026-09-02 three-valued-card probe moved to its own bullet with the anchor link resolving, both operator approvals in the relocate record moved into Incident/Recovery, the #369 → #380 routing characterization matches #369's close comment and #380's body verbatim, and the split-streaming/relocate diffs remove only snapshots (the deleted SHAs and rebase notes were already falsified by the squash merge). One suggested follow-up nit from the deletion audit (the `Ask the Dispatcher to move it.` string surviving in `feishu-channel/src/routing/index.ts:195`) I checked and **reject as a finding**: #380's requirement explicitly inventories that string as shipped behavior, so its survival was adjudicated, not dropped.
2. **Heading-based classification with fail-loud fallback: the direction is correct, the discriminator is not.** Refusing directory depth is right (`mcp/scheduler` proves it), and unknown → treated as a record → specific error is the right asymmetry. But the substring implementation makes the *other* asymmetry fatal: a false record fails loudly, a false index skips silently. Fix the matcher (blocking 1) and the design stands.
3. **Not scanning prose: accepted, conditionally.** A phrase list cannot distinguish a live status from the same words inside a verbatim ruling quotation (these records are full of them), and a gate that fails on quotations is worse than the drift. The corollary is that the manual sweep must carry the full load — it is why blocking 2 and 3 have to be completed in this PR rather than deferred; once the sweep is actually total, the written rule plus state gate is a defensible steady state.

### Findings on lines outside this PR's diff (file:line links)

**Blocking 2 (cont.) — stale status in this PR's own surfaces**

- [.agents/tasks/mcp/scheduler/remove-cron-run-now/README.md](https://github.com/excitedjs/dreamux/blob/66ec8f37d67fcca64a1eae4ae44abd0895fe1b11/.agents/tasks/mcp/scheduler/remove-cron-run-now/README.md#L12) — this is one of the four prose files the second commit explicitly claims to have swept: the Delivery line was fixed, but the Next action still says "Await explicit operator direction before merging" #335, which merged **2026-08-15** (`06b5ceac`). It also contradicts `State: done` in the same file. Set `- Next action: None.`

**Blocking 3 (cont.) — five more `done` records the sweep missed** (each verified merged on GitHub)

- [.agents/tasks/architecture/strengthen-independent-review-prompts/README.md](https://github.com/excitedjs/dreamux/blob/66ec8f37d67fcca64a1eae4ae44abd0895fe1b11/.agents/tasks/architecture/strengthen-independent-review-prompts/README.md#L28) — #392 merged 2026-09-08; "CI and operator re-review are pending; merge remains operator-owned" is false.
- [.agents/tasks/builtin-skills/teamwork-teammates-are-not-subagents/README.md](https://github.com/excitedjs/dreamux/blob/66ec8f37d67fcca64a1eae4ae44abd0895fe1b11/.agents/tasks/builtin-skills/teamwork-teammates-are-not-subagents/README.md#L24) — #393 merged 2026-09-08; "opened on 2026-09-08 … CI and merge pending" is false.
- [.agents/tasks/channel/align-codex-command-display/README.md](https://github.com/excitedjs/dreamux/blob/66ec8f37d67fcca64a1eae4ae44abd0895fe1b11/.agents/tasks/channel/align-codex-command-display/README.md#L13) — #396 merged 2026-09-09; the awaiting-approval/merge-authorization and "CI is tracked on the PR" clauses are false. Keep "Feishu client visual verification remains unperformed" — that is a durable coverage gap, not a status.
- [.agents/tasks/channel/refine-cot-tool-details-and-notifications/README.md](https://github.com/excitedjs/dreamux/blob/66ec8f37d67fcca64a1eae4ae44abd0895fe1b11/.agents/tasks/channel/refine-cot-tool-details-and-notifications/README.md#L83) — #405 merged 2026-09-10 as `af96c9f`; line 82 ("the remote draft PR and published Alpha **still contain** the discarded implementation"), line 83 ("External final review remains pending"), and the Next action on lines 90-95 ("Update the existing draft PR, run CI … remains required") are false pre-submission status. Keep the dated reset/restart narrative — that is history.
- [.agents/tasks/channel/display-turn-usage-summary/README.md](https://github.com/excitedjs/dreamux/blob/66ec8f37d67fcca64a1eae4ae44abd0895fe1b11/.agents/tasks/channel/display-turn-usage-summary/README.md#L27) — #397 merged 2026-09-09 as `bf090bc`; line 27 ("draft … Not ready for merge") and line 28 ("Await normal PR CI; merge is not authorized") are false. Keep the `Blockers` finding that real-model Codex requests time out and full live verification could not be green — a permanent, verified limitation (the record already states `done` does not mean live-verification green).

**Non-blocking (cont.)**

- [.agents/CONTRIBUTING.md](https://github.com/excitedjs/dreamux/blob/66ec8f37d67fcca64a1eae4ae44abd0895fe1b11/.agents/CONTRIBUTING.md#L81) — reconcile the append-only "Since this was recorded" doctrine ("historical text … is never edited in place") with `task-records.md`'s new "Rewrite stale facts in place; do not append change history" — this PR even deletes a `## Since this was recorded` section. State the boundary: append-only for frozen/backfilled snapshots, rewrite-in-place for active records. Also add check 5 to the Validation paragraph at [.agents/CONTRIBUTING.md](https://github.com/excitedjs/dreamux/blob/66ec8f37d67fcca64a1eae4ae44abd0895fe1b11/.agents/CONTRIBUTING.md#L109), which still enumerates only checks 1–4.

.agents/skills/dev-workflow/scripts/init_task.py
```diff
@@ -310,10 +327,225 @@ def check(args: argparse.Namespace) -> int:
...
+    """The lines that make a claim about the present.
+
+    Frozen sections are dropped on top of the fenced ones: they describe a past
+    round, and a record of what was true then is not a claim about now.
+    """
+    lines: list[str] = []
+    frozen = False
+    for line in unfenced_lines(text):
+        if line.startswith("## "):
+            frozen = line.startswith(FROZEN_SECTION_PREFIX)
```
ryanxiang7: **The Historical-heading matcher is a prefix, and that is a fail-open bypass.** `line.startswith("## Historical")` also matches `## Historically speaking …` or `## Historicalness aside …`, which then exempts every following line from the label rule until the next `## ` heading. I reproduced this against the gate at this head: a done-state record carrying banned `- Merge:` / `- CI:` labels under such a heading passes `check-all` with exit 0 (a legitimate `## Historical …` section is rejected, and a plain live heading is rejected — only the look-alike passes).

This is the same failure shape as the round-1 substring classifier: an over-broad heading match silently exempts content from the checks. Suggest a token boundary, e.g.:

```python
frozen = line == FROZEN_SECTION_PREFIX or line.startswith(FROZEN_SECTION_PREFIX + " ")
```

Both real `## Historical …` headings currently in the tree (`refine-cot-tool-details-and-notifications`) still match.

.agents/tasks/workflow/harden-native-transcript-contracts/README.md
```diff
@@ -8,11 +8,10 @@
 - Final solution: [Direct hardening solution](/.agents/tasks/workflow/harden-native-transcript-contracts/technical-design/final.md)
 - Solution review Issue: Skipped by explicit operator instruction for this bounded follow-up.
 - Verification: [Current verification evidence](/.agents/tasks/workflow/harden-native-transcript-contracts/verification.md)
-- Blockers: No accepted finding or required CI gate remains unresolved.
+- Blockers: None.
   External Codex model-service gates remain outside this bounded hardening
   change; GitHub review approval remains required.
```
ryanxiang7: **False and self-contradicting, in a bullet this commit rewrote.** PR #339 merged on 2026-08-16, so `- Blockers: None.` followed by the continuation `GitHub review approval remains required` asserts no blockers and an outstanding approval in one breath. The edit in this commit deleted the old `- Next action: Obtain GitHub review approval…` line but orphaned the same claim in the continuation above.

The first clause is a durable scope boundary and should stay; suggest:

```
- Blockers: None.
  External Codex model-service gates remain outside this bounded hardening
  change.
```

This is exactly the class this sweep exists to fix: `#339`'s number is checked against GitHub everywhere else in the commit, and this one was walked past.

.agents/tasks/channel/refine-cot-tool-details-and-notifications/README.md
```diff
@@ -80,20 +77,18 @@ The following implementation and delivery notes predate the current repair.
   closed. At that reset only .agents records remained changed; the fresh Claude
   implementation passed final local pre-review. The remote draft PR and published Alpha
   still contain the discarded implementation.
-- Blockers: External final review remains pending. The test-fixture scan failure
-  was resolved under operator approval by changing only the synthetic commands;
-  staged gitleaks now passes with its rules unchanged. The operator rejected the
-  added shared START title/name budget; its withdrawal is verified while the
-  approved 80-byte cap removal remains.
-  Earlier implementation and gate results are historical.
+- Blockers: None. The test-fixture scan failure was resolved under operator
```
ryanxiang7: **This rewrites a frozen snapshot into the present tense — the edit the new rule in this PR forbids.** This line sits inside `## Historical feature snapshot before PR #401 merged`. At the base it recorded the round's actual blocker: `- Blockers: External final review remains pending. …`, and a later sentence said `External final review remains required.` This commit replaces the recorded blocker with `- Blockers: None.` and deletes the required-review sentence, erasing the fact the snapshot exists to keep. The `Next action → Next action at the time` edit two lines below is the correct treatment; the same line needs e.g. `- Blockers at the time: external final review remained pending.` with the fixture-scan and START-budget sentences kept in past tense.

One more accuracy problem with the commit message: it says this commit repairs a present-tense rewrite the manual sweep made in this `## Historical` section. Git does not show one — `git diff 27428f85 66ec8f3 -- <this file>` is empty; the sweep commit never touched this file, and this commit is the first to edit the frozen section.

.agents/skills/dev-workflow/references/task-records.md
```diff
@@ -56,7 +56,68 @@ Use one of these states:
...
+  itself, because the record moves on and the index keeps yesterday's answer.
+
+The check enforces this on the part of the record that can be checked without
+reading a claim: the bullet **label**. A label is a closed token — it cannot
+appear inside a quotation, and rejecting one says nothing about the sentences
+under it. These report where delivery stood when they were written, so a record
+being committed to the trunk may not carry them:
+
+```text
+- CI:          - Merge:        - Branch:    - Baseline:    - Commit:
```
ryanxiang7: **Doc/code closed-set mismatch: 10 labels documented, 12 enforced.** `DELIVERY_STATUS_LABELS` in `init_task.py` also contains `Current repair baseline` and `Current repair branch`; a writer hitting either rejection cannot find the rule here. The same 10-label block is duplicated in `technical-design/final.md` §8. Please enumerate all 12 in both places. The runtime error message in `init_task.py` ("branch, baseline, commit, CI result, and merge status") also under-enumerates the categories — worth completing or generalizing in the same edit.

.agents/tasks/architecture/pin-task-record-states/technical-design/final.md
```diff
@@ -0,0 +1,154 @@
...
+and `## Acceptance criteria` inside an implemented note. A heading is a closed
+structural token: it is either present or absent, it cannot appear inside a
+quotation, and rejecting one says nothing about the sentences under it. That is
+how the gate reaches staleness without ever parsing a claim.
+
+The same token here is the bullet label. A record does not say "the CI is still
+running" in free prose — it writes `- CI:` and then says it. So the label is
+what the check rejects, and only on a record being committed to the trunk:
+
+    - CI:            - Merge:       - Branch:     - Baseline:
```
ryanxiang7: Same closed-set mismatch as in `task-records.md`: this block lists 10 labels while the code enforces 12 — `Current repair baseline` and `Current repair branch` are missing here too.

.agents/tasks/architecture/pin-task-record-states/technical-design/final.md
```diff
@@ -0,0 +1,154 @@
...
+| The first-proposed date in the filename | Slugs here are action-prefixed capability names used as stable link targets. A date in the slug would put an expiring-looking token into every link. |
+| A separate `archived/` tree with a hash manifest and an append-only seal | The exemption it exists to grant is borrowed — a `## Historical …` section is skipped — without moving files into a second tree or maintaining a manifest. Nothing here is sealed, so nothing needs a hash to prove it unchanged. |
+| A required section skeleton (`## Problem`, `## Decision`, `## Consequences`) | Upstream's note vocabulary is closed because it was mandated from the start. Ours is not: 130 distinct bullet labels are in use across 36 records, most of them genuine one-off facts. Requiring a skeleton would redden nearly every record for cosmetics. The *banned* half of the same gate transfers without that cost, and it is the half that catches staleness. |
```
ryanxiang7: **The 130-label count is not reproducible using the gate's own primitives.** Running `LABEL_RE` over `live_lines` (fences stripped, `## Historical` sections dropped) gives 142 distinct live labels across 36 records at `6dd0eef`, 147 at `66ec8f3`, and 141 at this head. Please recompute and state the counting rule (whether labels in frozen sections / fenced blocks are included) — the same 130 appears in the PR body and in the a10aac9 commit message.

.agents/tasks/architecture/minimize-provider-boundaries/README.md
```diff
@@ -588,11 +588,11 @@ Pure code-edge hardening that has no demonstrated product consequence belongs in
 the final holistic code review and must not be promoted into product state or
 architecture merely because it is possible.
 
-- Next action: finish the active Stage 5 correction round, review each remaining
-  product issue with the operator one at a time, and do not advance the later
-  stages until the confirmed Stage 0-5 drift is removed. After implementation
-  and ordinary review are complete, run the independent line-by-line Fable
-  audit required by the failure ledger above.
+- Next action: None.
+- Final Fable audit: the failure ledger below requires an independent
```
ryanxiang7: Tiny pointer inversion introduced by the restatement: the failure ledger is **above** this line (`## TeamLeader failure ledger for the final Fable audit`, L401), not below. The restatement itself is truthful — I searched the whole task tree and found no evidence the independent audit was ever performed.

.agents/tasks/architecture/pin-task-record-states/README.md
```diff
@@ -0,0 +1,40 @@
...
+- Approved implementation boundary: the rule and its enforcement in
+  `init_task.py` and `.agents/scripts/check.sh`, the dev-workflow references
+  that own the task-record format, and the existing task records and domain
+  indexes brought into compliance. No new CI job, and no change to the set of
+  nine workflow states.
+
+## Delivery
+
+- Pull request: [#413](https://github.com/excitedjs/dreamux/pull/413).
+- Scope delivered: the rule, its enforcement, and every record brought into
```
ryanxiang7, line 27: "Every record brought into compliance" is not true at this head, and this review lists the two live exceptions: `harden-native-transcript-contracts:11-13` still says GitHub review of merged #339 remains required, and `refine-cot-tool-details-and-notifications`'s live `## Delivery` still presents merged #401 as a draft. I independently scanned every live bullet in all 36 records for status vocabulary and those were the only two stale surfaces, so this claim becomes accurate with those fixes — but it should not ship ahead of them.

.agents/tasks/architecture/pin-task-record-states/README.md
```diff
@@ -0,0 +1,40 @@
...
+  compliance. Several were false rather than merely stale: PR #335 had been
+  merged for a month while its record said no merge had been performed, and
+  `adopt-lean-self-upgrade-sop` was `done` while its delivery said the pull
+  request had not started.
+- Coverage: the check reads the `State:` line, the index entries, and the
```
ryanxiang7, line 32: The previous version of this record stated the residual honestly (`- Known limit: the check reads the State: line and the index entries, not the prose`). This `- Coverage:` rewrite drops it, while `final.md` §8 still admits "the residue is small, but it is not zero". The two stale surviving surfaces are instances of exactly that residue; suggest one line here saying what the label rule still cannot see, so the record does not read cleaner than the check is.

ryanxiang7 (changes requested):
## Round 2 — CHANGES_REQUESTED

The redesign is sound and the round-1 blockers are genuinely fixed, but the head still carries two live false delivery facts, the new frozen-section exemption (a) has a matcher bypass I reproduced end-to-end and (b) is used in the same commit to erase a frozen fact, and the documented label set does not match the code. All fixes are small; the gate design itself I would keep.

### What I re-verified at this head

- All five round-1 items are fixed: own record carries the #413 link; `architecture/` and `channel/` indexes carry goal-only entries; `remove-cron-run-now` says `Next action: None`; the five unswept records are corrected, with the #396/#397 real-model Codex and unobserved-Feishu limitations preserved as `- Coverage limit:`; the count is now 7 of 35 and all three disagreeing index copies are resolved.
- The structural classifier behaves: whole-line heading match, fence stripping before *every* structure read (including classification), both-kinds refusal, and the state moved solely into a `## Historical` section still fails. Probe battery against `init_task.py` at this SHA: the `## Tasks completed` bypass is closed, all 12 banned labels reject in live lines, labels inside fences and real `## Historical …` sections are allowed, and the section unfreezes at the next `## ` heading.
- CI is 9/9 green on a10aac9; locally `check.sh` reports `Task records OK: 36 checked`.
- I audited the other 141 distinct live bullet labels at this head for near-misses; the ones I sampled are durable (dated approvals, SHA-256 pins, verbatim rulings). The Fable-audit restatement is truthful. The public-repo red line is clean.

### Required

1. **Inline — `init_task.py:453`.** The `## Historical` matcher is a prefix; `## Historically speaking …` silently disables the label rule for the rest of the section. Reproduced bypass, one-line boundary fix.
2. **Inline — `harden-native-transcript-contracts/README.md:13`.** "GitHub review approval remains required" is false (#339 merged 2026-08-16) and contradicts the new `- Blockers: None.` in the same bullet; keep the durable Codex scope clause, drop the approval clause.
3. **Inline — `refine-cot-tool-details-and-notifications/README.md:80`.** A frozen `## Historical` snapshot's blocker was rewritten from "external final review remains pending" to `None.`, deleting the recorded fact — against the past-tense doctrine added by this same PR. The `Next action at the time` edit shows the intended treatment. The commit message's claim that this repairs the sweep's rewrite of that section is not supported by git (the sweep commit never touched this file).
4. **Live `## Delivery` section in the same file — not in the diff, so no inline thread:** [refine-cot README L151-159](https://github.com/excitedjs/dreamux/blob/a10aac9f4f7a6535c7c9310d82f2283a9eebd684/.agents/tasks/channel/refine-cot-tool-details-and-notifications/README.md#L151) still says `- PR: [Draft #401], targeting `next`` and "external review may require further updates". #401 merged 2026-09-10, the record's own `## Current state` already names #405 as the delivery (and #405 also merged that day, so its `targeting `next`` at L15-18 and "the following closeout changes" are stale too), and the discarded #401 round — including the Alpha — is already preserved verbatim in the frozen snapshot at L63-93. The live area should hold a single, current delivery statement; right now the section literally named `## Delivery` is the stale one.
5. **Inline (two threads) — 10 vs 12 labels.** `task-records.md` and `final.md` §8 enumerate 10; the code enforces 12. Sync the docs and the error message.

### Nits

- [`.agents/CONTRIBUTING.md` L84-87](https://github.com/excitedjs/dreamux/blob/a10aac9f4f7a6535c7c9310d82f2283a9eebd684/.agents/CONTRIBUTING.md#L81): "Historical text (backfilled records, frozen snapshots) is never edited in place" still contradicts the new "Status matters" bullet and `task-records.md`'s instruction to past-tense a `## Historical …` section inside a record. One boundary clause closes it.
- [`adopt-completion-token-routing/README.md:108`](https://github.com/excitedjs/dreamux/blob/a10aac9f4f7a6535c7c9310d82f2283a9eebd684/.agents/tasks/completion-routing/adopt-completion-token-routing/README.md#L108): "Its acceptance and evidence will be recorded separately" — #384 merged 2026-09-07; the dated section is live (it is not a `## Historical` heading), so past-tense it as the other lines were.
- `final.md` §5 "16 of 1948 notes": the direction is verified true (I read the PR-citing notes, e.g. the 2026-09-09 blacksmith failover note cites PR #3841 inline as evidence), but 1948 counts *files* under `.agents/notes/` — roughly half are `.zh.md` translations. Say "files" or give the English-note denominator.
- `final.md` §7 vs §8: §7 says the gate reds 7 and 12 more passed it; §8 says 19 were corrected "because the gate could not see them". Both are right only across different gate versions (state-only v1 vs the final gate, whose label rule catches 7 of the 12); one sentence would remove the apparent contradiction.
- Inline: ledger pointer "below"→"above" (`minimize-provider-boundaries:592`), the 130-label count, and restoring the residual-gap admission in this record's `- Coverage:`.

### Two things I looked at and am not asking for

- **Do not add `- PR:` / `- Delivery:` to the banned set** (a tempting corpus-fit after item 4). Plain `- Pull request:` is the mandated durable label — the rule explicitly says "keep the pull request link" — and the staleness in `Draft #401, targeting next` lives in the prose, which is the admitted residue. Banning the nearest synonym would forbid the link the policy tells writers to keep; the set would then grow one synonym per PR. Likewise `Current repair baseline`/`Current repair branch` occur today only inside exempt sections — they read as forward defense rather than corpus cleanup; one line saying so would be enough.
- **Verification claim in the PR body:** "12 behavioral tests" are not committed anywhere (the PR adds only `init_task.py`; no test or fixture file is in the diff). Given the first version of this gate shipped the classifier bypass in this same PR, I would prefer a tiny committed fixture test for the rejection paths; at minimum please call them manual probes so the Verification section describes the tree.

### Process note

The substantive bar this PR sets — structural, fail-loud, upstream-shaped — is the right one and the commit documents it unusually honestly. The remaining asks are the last two percent: two stale surfaces in files the commit already edited, one matcher boundary, and doc/code parity.

🤖 Devbox round-2 review.

.agents/tasks/architecture/strengthen-independent-review-prompts/README.md
```diff
@@ -25,7 +24,6 @@
```
ryanxiang7, line 24: **One surviving stale delivery fact, in a record this PR swept.** #392 merged into `next` on 2026-09-08, so present-tense "targeting `next`" went false at merge — the exact expiry this PR exists to eliminate. This is the same phrase this branch deleted from many sibling records (`unified-teammate-lifecycle`, `change-workspace-default`, `refine-feishu-cot-and-binding-cards`, the new `- Pull requests:` paragraph in `refine-cot`); it was left here because a10aac9 edited the lines directly adjacent (the old `- CI / merge:` bullet). One-line fix, matching the siblings: `- Pull request: [#392](https://github.com/excitedjs/dreamux/pull/392).`

.agents/tasks/architecture/pin-task-record-states/README.md
```diff
@@ -0,0 +1,41 @@
...
+- Status: Granted by the operator on 2026-09-11 (Feishu group), verbatim: "可以，那边贡献的人多，你先把规则定死了，剩下的状态起一个 workflow 去刷一遍。"
+- Approved implementation boundary: the rule and its enforcement in
+  `init_task.py` and `.agents/scripts/check.sh`, the dev-workflow references
+  that own the task-record format, and the existing task records and domain
+  indexes brought into compliance. No new CI job, and no change to the set of
+  nine workflow states.
+
+## Delivery
+
+- Pull request: [#413](https://github.com/excitedjs/dreamux/pull/413).
```
ryanxiang7, line 26: "Every record brought into compliance" becomes literally true only after the last two "targeting `next`" lines for merged PRs are dropped — `strengthen-independent-review-prompts:24` (#392 merged 09-08, inline thread) and `refine-cot:17` (#405 merged 09-10, in the review body because it sits outside the diff hunks). I scanned every live line of all 36 records; those two are the entire residue of the "targeting-next" class.

.agents/tasks/architecture/pin-task-record-states/technical-design/final.md
```diff
@@ -0,0 +1,158 @@
...
+request had not started.
+
+That gap is the honest limit of the check, and §8 states it rather than leaving
+a reader to infer that a green gate means a clean tree.
+
+## 8. Checking staleness without reading prose
+
+The first version of this check read only the `State:` line and the index
+entries, and its stated limit was that a record could still describe finished
+work in the present tense and pass. That limit was real: nineteen records were
```
ryanxiang7: Nit, now that the PR body says 24 records: §7 accounts for 7 + 12 = 19 corrections, and this line says 19 were invisible to "the gate" (true only of the v1 state-only gate; the final label gate caught 7). Both 19 and 24 can be honest with one defining clause: 19 state/prose-stale corrections vs 24 pre-existing records edited (the other 5 got durable fields relabeled, e.g. `- Blockers:` → `- Coverage limit:`). As written a reader cannot reconcile §7/§8 with the PR body — third occurrence of the count-class issue this design argues against.

ryanxiang7 (changes requested):
## Round 3 — CHANGES_REQUESTED (narrow: two one-line fixes)

The round-2 required work is genuinely done, and I executed the new probes myself rather than reading the claim. Two stale "targeting `next`" lines survive in records this branch itself swept, one of which round 2 already named. Everything else below is nit and does not need another review cycle.

### Independently verified at this head

- **All five round-2 required items fixed.** Historical matcher is now `^## Historical\b` — I reproduced the fail-open on a10aac9 (exit 0 with `- CI:` under `## Historically speaking …`) and confirmed exit 1 here, alongside bare/qualified/fenced/lowercase boundary cases; harden-native keeps the durable Codex boundary as `- Coverage limit:` with the false approval clause gone; the frozen `refine-cot` blocker is restored in past tense with the fact intact; live `## Delivery` now uses past-tense `- Pull requests:` and points at #401/#403/#405 (the discarded alpha is preserved, linked from verification.md); docs and error message match the 12-label frozenset programmatically.
- **14/14 committed probes executed verbatim from `verification.md`** — mutating a real record, asserting exit codes, restoring byte-identical (I diffed both files after the run). I also ran 6 extra boundary probes of my own; all behave.
- **Counts reproduced:** 24 corrected records (25 changed READMEs minus this new one), 7 indexes, 139 labels via the gate's own primitives ("about 140"), 238 KB files.
- **The self-correction is fully accurate:** `origin/next` (27428f8) L83 still reads `- Blockers: External final review remains pending.`; `git diff 27428f85 66ec8f3 -- <refine-cot>` is empty; a10aac9 authored the present-tense line itself. Fixing it forward without force-push was the right call, and no round-2 anchor was lost.
- CI 9/9; all 40 changed files under `.agents/`; red-line scan clean (the `sys.dont_write_bytecode` story in verification.md is a nice catch and correctly documented).

### Required

1. **Inline — `strengthen-independent-review-prompts:24` (#392).** `- Pull request: [#392], targeting `next`` — #392 merged 2026-09-08.
2. **[refine-cot README L17-18](https://github.com/excitedjs/dreamux/blob/bfb15ebc399627d77eb05867b2ecc0ed07917063/.agents/tasks/channel/refine-cot-tool-details-and-notifications/README.md#L17) (outside the diff hunks, so no inline thread):** `- Delivery: [PR #405], targeting `next`` — #405 merged 2026-09-10. Round 2 item 4 named this twin explicitly; the `## Delivery` section 130 lines below got rewritten, this one did not.

These are the only two hits of their class in all 36 records — the other "targeting next" occurrences are dated operator-authorization bullets (#396/#397 and the dispatch task) that describe what was authorized that day, which cannot expire.

### Nits (non-blocking; no re-review needed)

The Feishu note said all seven round-2 nits landed; three did not, plus two new ones:

- [CONTRIBUTING.md L81-83](https://github.com/excitedjs/dreamux/blob/bfb15ebc399627d77eb05867b2ecc0ed07917063/.agents/CONTRIBUTING.md#L81) still says frozen snapshots are "never edited in place"; this branch just past-tensed one in place per `task-records.md`, which is the contradiction round 2 flagged. One boundary clause.
- [`adopt-completion-token-routing:108`](https://github.com/excitedjs/dreamux/blob/bfb15ebc399627d77eb05867b2ecc0ed07917063/.agents/tasks/completion-routing/adopt-completion-token-routing/README.md#L108) still promises acceptance/evidence "will be recorded separately" (#384 merged 09-07; the section is dated, so this is small).
- [refine-cot L161-163](https://github.com/excitedjs/dreamux/blob/bfb15ebc399627d77eb05867b2ecc0ed07917063/.agents/tasks/channel/refine-cot-tool-details-and-notifications/README.md#L161) "external review may require further updates" — an open hedge rather than a status GitHub can falsify; past-tensing it matches the new `- Pull requests:` paragraph.
- This record's `- Coverage:` still omits the residual-gap sentence the r2 review asked to keep; `final.md` §8 carries it.
- Inline above: §7/§8 19-vs-24 needs one defining clause.

Worth recording as follow-up work, not blocking: the 14 probes are documentation — nothing in CI runs them. Reverting the `\b` boundary leaves check.sh green because no live record uses a `## Historically…` heading. A tiny `init_task.py self-test` wired into check 5 is the clean home; I agree it is out of scope for this PR.

### Verdict

The gate is correct on every boundary I could construct, the docs cannot silently drift from the code anymore, and the branch corrected its own record honestly. Drop the two "targeting `next`" phrases and this is an approval.

🤖 Devbox round-3 review.

.agents/tasks/architecture/pin-task-record-states/README.md
```diff
@@ -0,0 +1,47 @@
...
+  delivery-status bullet labels. The label rule is
+  [ported from upstream's format gate](/.agents/tasks/architecture/pin-task-record-states/technical-design/final.md),
+  which rejects a closed set of structural tokens rather than parsing prose —
+  the first version of this task checked only the state line and left the rest
+  to a hand sweep. What it still cannot see is stale text inside a label that is
+  not banned: `teamwork-teammates-are-not-subagents` said "Land the pull request
+  after CI and review" under `- Next action:` after #393 had merged, and only a
+  reader found it. It also reads `README.md` and nothing else, so the same prose
+  in a sibling `requirement.md` or `verification.md` is out of its reach — one
+  such line, `#339` still "targeting `next`" a month after it merged, was found
```
ryanxiang7: The grep this sentence documents also reached `harden-native-transcript-contracts/verification.md` — line 111 was corrected in this very commit — but stopped inside the same section: line 123 still carries a live `- PR remains unmerged` for #339 (merged 2026-08-16), and `unified-teammate-lifecycle/verification.md:252` says the same for #338 (merged 2026-08-16). Details and one-line fixes in the review body.

ryanxiang7 (changes requested):
## Round 4 — `b53e02d9`: one narrow class remains (2 one-line fixes + 1 bundled nit)

Everything round 3 named is fixed, and I re-ran the measurements rather than trusting the prose:

- Both r3 blockers: `strengthen-independent-review-prompts/README.md:25` and `refine-cot-tool-details-and-notifications/README.md:17` are plain links now.
- All five nits landed (CONTRIBUTING tense-vs-meaning boundary; `adopt-completion-token-routing/README.md:108`; the refine-cot hedge; pin `- Coverage:`; final.md §7/§8), plus the self-found `harden-native-transcript-contracts/verification.md:111`.
- The 14 committed probes execute 14/14 here, both mutated files restore byte-identical; the pinned snippet prints exactly `7 23 23` with `state ⊆ label`; the head label count is 137; `check.sh` reports 36 records / 238 KB files; CI 9/9; the two unified-teammate dedup lines are byte-identical by SHA-256.

**Correction of my round-3 review.** I wrote that five other records "got durable fields relabeled." That is wrong, twice. The structural partition gives **4** records that gained `- Coverage limit:`, and all 4 are inside the 23 label-rejected records — each also carried genuinely expired `- CI:` / `- Merge:` / `- CI / merge:` content at base. 24 changed records = the 23 rejected + `teamwork-teammates-are-not-subagents` (the only changed record carrying no banned label at base); no banned-label record was left untouched, and no record was edited for a relabel alone. The structural disproof was correct.

### Blocking

The gate reads `README.md` by design — I accept that boundary; what it cannot structurally see cannot fail *this gate*. But `b53e02d9` itself extends the cleanup past that boundary ("One had expired the same way … Correct the last stale delivery phrase, outside the gate's reach"), so that commit is rightly judged against its own claim of **the** last. It is not the last:

1. **[`workflow/harden-native-transcript-contracts/verification.md:123-124`](https://github.com/excitedjs/dreamux/blob/b53e02d986e8aa1b805730ac7f03f0c8c270f6bc/.agents/tasks/workflow/harden-native-transcript-contracts/verification.md#L123)** — twelve lines below the line `b53e02d9` corrected at :111, in the same `## PR Delivery` section, about the same #339 (merged 2026-08-16): `- PR remains unmerged; GitHub review approval and operator merge authority are separate gates.` This is the only `## PR Delivery` heading in any `verification.md` in the tree. The second clause is durable policy; the first is a flat present-tense status the merge falsified four weeks ago. One-line fix: drop the status clause, or past-tense it ("The PR was unmerged at delivery; …").
2. **[`workflow/unified-teammate-lifecycle/verification.md:252-253`](https://github.com/excitedjs/dreamux/blob/b53e02d986e8aa1b805730ac7f03f0c8c270f6bc/.agents/tasks/workflow/unified-teammate-lifecycle/verification.md#L252)** — under the live `## Current Native-Transcript Iteration` record: `The PR remains unmerged. GitHub reports the review requirement as the remaining merge gate; merge authority remains with the operator.` #338 merged 2026-08-16 — same sentence shape, same age, and this task's README was already edited in this PR. Same one-line treatment.

### Non-blocking nit (fold it into the same edit pass)

3. **[`channel/refine-cot-tool-details-and-notifications/verification.md:51`](https://github.com/excitedjs/dreamux/blob/b53e02d986e8aa1b805730ac7f03f0c8c270f6bc/.agents/tasks/channel/refine-cot-tool-details-and-notifications/verification.md#L51)** — trailing clause `…; final CI remains the last merge gate` right after "authorized squash-merge with “合入吧。”" #405 merged later that day (2026-09-10). It sits in a dated `## … (2026-09-10)` section, so it would not block on its own; deleting or past-tensing the trailing clause closes it.

I re-ran the full-tree sweep for unmerged / pending / last-gate status prose to make sure this list is complete: it is. Deliberately not listed, because durable: dated operator-authorization bullets (`align-codex-command-display/README.md:38`, `display-turn-usage-summary/README.md:21`, `strengthen-dispatch-and-compaction-text/README.md:21`), the permanent "#369 closed unmerged by operator ruling" family, git-worktree `unmerged` technical prose, design-phase rows such as `split-streaming-display-from-pushback/technical-design/final.md:28`, and dated proposal records such as `platform-run-roots-and-logging/requirement.md:151` (dated 2026-06-05).

On my round-3 "delete those two lines and I approve" — that commitment covered `bfb15eb`. The subsequent commit reopened the sibling-file surface itself and asserted completeness on it; :123 is a direct counterexample inside the very section that commit edited. Once these clauses are past-tensed (the first two are what I hold approval on), the gate design, the measurements, and the tree are in good shape.

On the bookkeeping question from the thread: I agree with keeping "24 records / 7 indexes" as a PR-body historical note. An executable `27428f85..HEAD` reconciliation script expires at squash time, and hardcoding the 24 names would encode a one-time census as permanent machinery; the prose note is the lower-entropy home.

.agents/tasks/architecture/pin-task-record-states/README.md
```diff
@@ -0,0 +1,49 @@
...
+  delivery-status bullet labels. The label rule is
+  [ported from upstream's format gate](/.agents/tasks/architecture/pin-task-record-states/technical-design/final.md),
+  which rejects a closed set of structural tokens rather than parsing prose —
+  the first version of this task checked only the state line and left the rest
+  to a hand sweep. What it still cannot see is stale text inside a label that is
+  not banned: `teamwork-teammates-are-not-subagents` said "Land the pull request
+  after CI and review" under `- Next action:` after #393 had merged, and only a
+  reader found it. It also reads `README.md` and nothing else, so the same prose
+  in a sibling `requirement.md` or `verification.md` is out of its reach. Nine
+  such lines were corrected during this task, found by grepping `.agents` for
```
ryanxiang7, line 42: Method named instead of a remainder count, with the explicit non-completeness caveat — this resolves the round-4 thread. I re-ran the sweep independently and confirmed the nine-line set is complete under the merge-falsifiable criterion; details in the APPROVE body.

ryanxiang7 (approved):
## Round 5 — `ff307f92`: APPROVE

The round-4 blockers are fixed, and I verified the wider resweep rather than the three lines I named.

**All nine lines are past-tensed in place, and I checked each fix in its section context** (not just the changed words):

- `workflow/harden-native-transcript-contracts/verification.md:123` and `workflow/unified-teammate-lifecycle/verification.md:252` — "The pull request was unmerged when this record was written …", with the durable policy clause ("merge authority rests with the operator") kept present-tense. #339/#338 both merged 2026-08-16.
- `channel/refine-cot-tool-details-and-notifications/verification.md`:51/246/316/391/402 — all five now read as dated narratives ("was pending at that point", "when this was written"), coherent under `Previous attempt`, `Final review handoff`, `Failure-order correction`, and `Remaining delivery and limitations` respectively; #405 merged 2026-09-10.
- `completion-routing/adopt-completion-token-routing/README.md:71` (under the dated 2026-09-07 heading) and the round-3 nit `…/verification.md:245` ("its results are recorded below" — the cleanup section does follow).

**I independently re-ran the sweep** (my own pattern set over all of `.agents`, not hers): every remaining hit is durable — dated operator-authorization bullets, the permanent "#369 closed unmerged by operator ruling" family, archived proposals, quoted findings in review logs (e.g. `adopt-completion-token-routing/verification.md:166`), code-state vocabulary (pending promises, open cards, unmerged worktrees), and validation-log result rows (e.g. `strengthen-dispatch-and-compaction-text/verification.md:22`, where "`smoke-built-cli`: pending" is a recorded gate result that no later event falsifies). The 9-set is complete under the stated criterion — lines a merge could falsify.

**The treatment of the failed completeness claim is right on both axes:** `- Coverage:` now names the method (grep → read all hits → keep the merge-falsifiable) and explicitly says it cannot prove completeness, instead of asserting a remainder count; and the `ff307f92` message drops "the last" while leaving `b53e02d9`'s message unrewritten so this record stays auditable. A method can be re-run and argued with; a completeness assertion can only be overturned. That is the correct entropy call, and it is the thesis of the gate itself.

**Re-run by me at this head, not taken on faith:** the 14 committed probes execute 14/14 against the live gate and both mutated files restore byte-identical (`git status` clean afterwards, no `.pyc`); the snippet pinned to `27428f85` prints exactly `7 23 23`; the head label count is 137; `check.sh` reports 36 records / 238 KB files; the gate code and `check.sh` are byte-identical to `bfb15eb`; CI is 9/9 including the public-repo content scan; PR is OPEN/MERGEABLE.

One reviewer note for the operator's scope decision on probes-in-CI: the round-5 miss does not strengthen the case for bundling the 14-probe self-test into this PR. The nine lines were prose that the gate is designed not to read — a self-test of the gate's structural rules would not have caught any of them; the failure that motivates the self-test is a gate-code bypass regression (the round-2 `startswith` class), which has not recurred. `init_task.py self-test` wired into check 5 remains the correct home and remains a clean fast follow-up; keeping this PR's gate code frozen since `bfb15eb` made the last two rounds cheap to verify. Either way, not a blocker.

On the round-4 exchange: the author was right to correct my "five relabels" (it was four, all inside the 23), and my round-4 sweep was right that the class exists but wrong that it was exhaustive — 3 of 9. Both errors are the same lesson written down in `- Coverage:`: prose completeness is not provable by hand, which is why the rule hangs on closed labels and the residual gap is stated, not implied away. Good to merge on the reviewer side; merge timing stays with the operator.

.agents/tasks/architecture/pin-task-record-states/verification.md
```diff
@@ -0,0 +1,176 @@
...
+| | `check.sh` | the fourteen probes |
+| --- | --- | --- |
+| `^## Historical` prefix test restored | exit 0, reports `Task records OK: 36 checked` | 13 `PASS`, 1 `FAIL` — the `## Historically speaking` probe |
+| the expired prose this task corrected by hand | exit 0 | all `PASS` |
+
+The first row is the gap: the fail-open this task shipped twice would return
+today without CI noticing. The clean home is a `self-test` subcommand on
+`init_task.py`, wired into `check.sh`'s fifth check, which needs no new CI job.
+
+The second row is not a gap and must not be treated as one. Those lines carry
```
ryanxiang7, line 107: This boundary sentence is the valuable part of the paragraph. I reproduced the table's first row independently (check.sh green, probes 13/1 with the prefix test restored); the explicit "second row is not a gap" ruling is what keeps a future self-test task from growing a prose parser. APPROVE re-stated on ac80c366 — see the review body.

ryanxiang7 (approved):
## Post-approval commit `ac80c366`: reviewed, APPROVE stands on this head

Thank you for declaring the new commit rather than letting the `ff307f92` approval cover an unseen head. I reviewed `ac80c366` ("Record that nothing but a person runs the probes") on its own merits and reproduced its evidence instead of accepting the table:

- The diff is one file, one documentation paragraph in this task's `verification.md`; the gate and `check.sh` are untouched.
- I reverted `FROZEN_SECTION_RE` to the round-2 prefix test myself and ran both surfaces: `check.sh` exits 0 printing `Task records OK: 36 checked`, while the probes report exactly 13 `PASS` / 1 `FAIL`, the failure being the `## Historically speaking` probe. Restored byte-identical (`git status` clean), after which the probes are 14/14 again.
- Re-run at this head: the pinned measurement still prints `7 23 23`, the head label count is 137, `check.sh` reports 36 records / 238 KB files, CI is 9/9 including the public-repo scan, and the PR is OPEN/MERGEABLE with review decision APPROVED.

The paragraph also records the boundary that matters most: the prose row "is not a gap and must not be treated as one … Teaching the gate to parse prose to close it would add the mechanism this design exists to avoid." That forecloses the well-meaning follow-up that would have turned a closed-token checker into a prose parser — worth more than the measurement itself.

This satisfies the repo's "leave the cleanup trail" rule in-repo, so the separately proposed tracking issue is now optional rather than load-bearing: the conclusion cannot be lost between an operator decision and a future task. A short issue would still help future routing discover the `self-test` follow-up, but none of the review criteria depend on it.

Approval is re-stated against `ac80c366ca28a0fccfc6cc61733d90e252fd2225`. Good to merge on the reviewer side; merge timing remains the operator's call.

PR #414, opened 2026-09-11
Use native Feishu replies and preserve reply context

### LINKED ISSUE #410: Simplify Feishu replies with native Markdown and inline mentions
# Simplify Feishu replies with native Markdown

## Problem and intended behavior

The reply tool carries mentions separately through multiple adapters, then
transforms Markdown into interactive-card elements. The operator requested inline
XML mentions and simpler reply logic, and selected native Feishu rich-text
presentation on 2026-09-11.

The reply contract becomes `reply(chat_id, message_id?, text)`. Text becomes
native `post` content containing `md` elements. Feishu renders headings, tables,
code, links, and `<at user_id="ou_example">Example</at>` at their authored
positions. A long reply can still produce several messages, in order.

This final implementation plan incorporates both external solution reviews and
the operator's explicit requirement to retain existing peer message formats.
Development authorization is pending the operator's discussion of this plan.
The selected presentation direction does not authorize deletion of unrelated
public APIs.

## Evidence

- Source baseline: `3cac2f7b00a7be159a52bd6aee0a9cdcd43214e1`.
- Tool and channel owners: `packages/channel/feishu-channel/src/tools/messaging-tools.ts`,
  `tools/types.ts`, `feishu-channel.ts`, `feishu-session-ops.ts`, and `bot.ts`.
- Transport owners: `packages/channel/feishu-transport/src/transport/feishu.ts`,
  `transport/outbound-card.ts`, `contract/outbound.ts`, and `render/`.
- [Official message content format](https://open.feishu.cn/document/server-docs/im-v1/message-content-description/create_json.md)
  documents native `post` Markdown, inline XML mentions, tables, and code blocks.
- [Official reply API](https://open.feishu.cn/document/server-docs/im-v1/message/reply.md)
  accepts `post` and documents the 30 KB content-size constraint and source-message
  addressing. A reply to an existing topic message stays in that topic.
- Five native-post examples were sent through direct Node HTTP calls, bypassing
  CLI/SDK/renderers. The operator reported that they looked fine. Readback and
  current-source replay separately exposed the inbound defects below.
- [Official inbound content](https://open.feishu.cn/document/server-docs/im-v1/message-content-description/message_content.md)
  recommends the nested `content_v2` projection for native Markdown fidelity.
- [Official message GET](https://open.feishu.cn/document/server-docs/im-v1/message/get.md)
  documents different bot-mention identities when `user_id_type` is supplied.

## Ownership and flow

The channel owns tool inputs and outbound address observation. Transport owns
Feishu message serialization, platform limits, and actual API calls. Core does
not learn a Feishu rendering mode or mention syntax.

The resulting flow is:

1. Parse the three reply fields and pass the body without mention rewriting.
2. Resolve the existing chat/source-message address at the channel boundary.
3. Encode native Markdown post content at transport; split only if the serialized
   content would exceed the existing safe platform budget.
4. Call the existing message create/reply endpoint and observe each successful
   creation before sending another part.
5. Return the created message IDs in order. Preserve normal platform errors.

## Remove the separate mention mechanism

Delete `mention_user_ids` from the tool schema, parsing, and handling; remove
`mentionUserIds`/`mentionUsers` from reply session and outbound-address contracts.
Remove `textWithLeadingMentions` and its count-only logging. Remove the array
validator if it has no remaining caller.

The pairing resend reminder currently passes `mentionUserIds` at
`feishu-session-inbound.ts:228`. Change that caller to prefix its existing body
with the literal `<at user_id="${inbound.sender_id}"></at>\n` instead. The sender ID
comes from the platform event, not arbitrary user text. This empty-name spelling
was accepted by the direct API and returned the intended person as a mention;
its supplementary client check remains pending. It needs no display-name lookup.
The tool's `text` description explains the same native syntax with a synthetic
example. No custom `<@...>` alias is processed by the new reply path. Remove the
helper at `transport/feishu.ts:652` and count-only field at
`feishu-session-ops.ts:177`.

Keep `channelOutboundToFeishuTarget` as the shared address adapter for all three
callers; remove its mention field mapping. Bypassing it for reply alone would
leave two target-construction mechanisms. Leave the unrelated public routing
hint outside this change: the review's no-producer finding is not evidence about
external callers and does not justify widening this task's public API removal.

## Native post construction and long messages

An ordinary reply produces the following content shape, serialized once for the
message API's `content` field:

```json
{"zh_cn":{"content":[[{"tag":"md","text":"The authored Markdown body"}]]}}
```

The common one-message path does not convert Markdown into card elements or
change heading/table/mention syntax. Do not retain a card fallback for unsupported
clients; native client rendering was the operator's selected behavior.

Long-message handling has a concrete existing scenario: reply currently sends
long reports as several messages. Reuse the existing 28 KB safe content budget
and trial-serialization technique; measure native post JSON with its escaping.
Do not introduce an independent full-HTTP-request budget. Preserve block boundaries where they fit;
reuse the existing Markdown tokenization/splitting primitives only where needed
for oversized input. A split code block must retain its fence context; a table
split across messages must remain readable with its header: retain the existing
long-table capability by repeating only that block's first two raw lines, without
token-to-Markdown serialization. Use the existing lexer's table classification,
then capture the source prefix with `/^(?:[^\r\n]*\r?\n){2}/`; do not introduce
a second GFM table recognizer that would reject optional outer pipes.
Unicode must not be truncated or corrupted.
Do not add a separate XML-aware byte splitter for an unobserved oversized single
line containing a mention.
Do not render card-native table elements or apply card element/column/cell limits
to native posts.

Keep the implementation limited to encoding and splitting. It must not grow a
second Markdown-to-platform renderer. The implementation review should compare
the complexity removed from the reply path with the new size-handling code.

## Reuse the actual send boundary

The current transport helper handles create/reply addressing twice depending on
whether an AbortSignal was supplied. The native post path and explicit card path
should use one platform message-send primitive, parameterized by the actual
Feishu message data rather than a new high-level rendering mode. Preserve the
existing AbortSignal behavior for explicit card sends. Do not change topic
semantics or add a chat-mode lookup to send a reply.

Retain `onMessageCreated` ordering and its non-authoritative observer behavior.
The source message and every successfully created part remain addressable even
when a later send fails. No retries, rollback, or durable delivery ledger are
introduced by this change.

## Inbound Markdown and mention identity

The operator explicitly requires compatibility with peer bots' existing sending
formats. The current card-based peer-bot mention/receipt path is an established
working baseline, not an unproven new capability. Sending native posts ourselves
does not narrow accepted input to native posts. Separate three facts: platform
message receipt, access-gate admission, and model-facing mention representation.
The formatting defects below do not establish failure of the first two.

Both `bot.normalizeInboundEvent` and message-read enrichment feed JSON-encoded
content to transport's `parseInbound`. Keep that one parsing boundary. For native
posts, actual message GET responses contain `content` and `content_v2` inside the
decoded post object, not on the outer event or body. The current `parse/post.ts:26`
selects only `content`; recognizing `md` at line 74 is insufficient. Replays of all
five captured responses demonstrate lost heading syntax, table alignment, inline
code delimiters in the captured examples, and extra blank lines in code. This is
not universal inline-code loss: legacy nodes retaining `style: ['code']` are
already re-fenced by the parser. Prefer the supplied `content_v2`
array inside the locale-aware parser; use legacy `content` when it is absent.
Never concatenate the projections or add an ordinary-post readback network hop.

Make ordered mention parts a transport parsing result shared by text, post,
and interactive cards. Interpret each source with its actual native semantics:

| Inbound source | Mention representation to parse |
| --- | --- |
| Text receive/readback | `@_user_N` resolved from the platform mention map |
| Structured rich post or simplified card | `at` nodes and their identifier/name fields |
| Native post Markdown | `<at user_id="ou_example">Example</at>` |
| Card `markdown` / `lark_md` | `<at id="ou_example"></at>`, including the unquoted ID form observed in live card readback |

These are native input formats, not outgoing aliases. Plain-text card fields,
escaped examples, and code-contained XML retain literal meaning; do not scan all
flattened text indiscriminately. The parser must retain whether a visible field
was plain text or Markdown until its inline meaning has been parsed.

Resolve real platform placeholder keys against the accompanying mention records;
native Markdown's actual inline tags also retain their identifier, display name,
and source position. A tag/node identifies a mention occurrence; resolve its ID
against the accompanying records first by `key`, then by supported identity.
The matching record supplies its display name, including for empty-name tags.
The metadata map never invents an occurrence absent from the source. Retain a
source-provided supported identity when no record matches; do not resolve names
over the network. Preserve all remaining Markdown source slices. Interpret
inline and fenced code boundaries before promoting a tag or image reference;
code examples remain code. Existing `marked` tokenization is available at this
layer; a new parser dependency or reconstructed Markdown document is unnecessary.
Transport owns these code boundaries and emits code parts. Delete the channel's
`promoteMarkdownCode` scanner as well as its raw-text mention reparsing; the
channel only serializes text, code, mention, and resource parts. Preserve the
existing plain-text message contract rather than interpreting its literal
Markdown as rich content.
Keep ordinary image/file/media nodes and native Markdown image resources in the
ordered resource projection, with existing download/deduplication behavior.

The actual reviewer card was read through both existing GET modes. Its
`user_card_content` response contains a Markdown `<at id=...></at>` and one open-ID
mention record, while the default response has no mention metadata and exposes
rendered resources. Current-source replay preserves the XML only as escaped text.
At `parse/card.ts:150-155` and `:262-269`, Markdown and plain-text fields currently
collapse into the same part; both structured `at` branches keep only the display
name. Pass the associated mention map into card parsing and resolve semantic
parts before that type distinction is lost. Do not add a card-only rewrite in
the channel renderer.

Retain the existing structured/default card read and merge owners. Each read is
parsed using its own metadata. Extend their visible-content comparison to account
for mention identity so the same mention in both views is not appended as new
content merely because it is no longer a text part. Preserve intentional repeated
mentions in the primary view and distinct identities sharing a name; do not add
global mention-ID deduplication. Hidden callback/control values remain excluded.

The channel serializes actual mention parts as
`<at user_id="ou_example">Example</at>`, matching native reply syntax. This
explicitly changes the model-facing mention attribute from `id` to `user_id`.
Remove the text-only raw-body reparsing in `feishu-message-render.ts:121-176`;
the channel should serialize parsed meaning, not independently recover it.
Keep normal text escaping and code CDATA. A selection-only replay shows why this
is needed: simply switching arrays leaves real `<at user_id>` tags as escaped
text. There is no outgoing `id`-to-`user_id` rewrite or shorthand converter.

For message GET alone, omit the explicit `user_id_type: 'open_id'` at
`transport/feishu.ts:616`. A live read of an incoming message mentioning this bot
returned `id_type: open_id` with the known bot ID when omitted, but `id_type:
app_id` when explicitly supplied. Current normalization mislabels the latter as
`open_id`. Also give received `app_id` an explicit normalizer case: retain the
record's key/name but do not place the application identifier into any supported
user identity slot. A regression assertion must show that this actual response
shape never becomes an open-ID mention or a model-facing `user_id="cli_..."`.
Retain `card_msg_content_type` when requested; do not add another bot
identity lookup or change other APIs' identity parameters.

The raw realtime event's `mentions` still drives admission before enrichment.
It is independent of the deleted reply argument, and body markup cannot replace
it. Interactive-card enrichment keeps its existing reads and lifecycle while its
parser/visible-content merge carries semantic mentions. Lazy merged forwards,
parent-reference behavior, resource handling, and slash-command entry-type
restrictions keep their existing contracts. Add end-to-end parser/body tests with the actual dual-array payload
shape instead of treating a synthetic `md` node in legacy `content` as sufficient.
Although admission does not consume parsed content, `normalizeInboundEvent`
does call the parser before reaching the gate. Parser termination and handling
of existing inputs remain necessary regression checks; raw-mention ownership
alone does not prove that every parser change is incapable of affecting receipt.

## Public API boundary and remaining renderer

The public package also exports `renderMarkdownToCards` and related functions,
and `FeishuTransport.editText` uses that renderer to edit existing cards. Dreamux
has no production `editText` caller, but this alone does not authorize removing
the package's public behavior.

This proposal preserves those explicit rendering/editing APIs and their existing
behavior. The renderer no longer participates in reply/text-send. Its own
shorthand conversion remains scoped to the public card-rendering API; the reply
tool accepts native XML only. This is a deliberate scope boundary, not a claim
that all Markdown-to-card code has been deleted. Eliminating those remaining
APIs would require a separate operator decision about external capability loss.

Keep the public `applyMentions` and `mentionName` helpers too. The former loses
its remaining internal caller when text parsing emits semantic parts; the latter
already has no production caller. Their exported behavior remains available to
external consumers. This is explicit public API accounting, not another internal
mention mechanism.

Explicit `sendCard`, `editCard`, question/binding cards, and COT remain card paths.
Their actual card size constraints remain where they are used. No state or
configuration migration is needed.

`FeishuTransport.send` is a public in-process API also used by an external proxy
consumer, as documented by the package README. Its card-to-post presentation
change and removed mention option need an explicit package release note. Keeping
the explicit rendering/editing APIs does not hide that send-contract change.
Adding `mention` to the exported `InboundContentPart` union must appear in the
same note because external exhaustive consumers must handle the new variant.

## Validation and delivery

- Exercise the actual tool session and transport with native post payloads,
  inline mentions, create/reply addressing, Markdown fidelity, and long content.
- Preserve existing assertions for per-message receipts, partial success,
  observer failures, topic addressing, explicit card sends, cancellation, and
  legacy rendering/editing APIs. Update only the reply presentation assertions
  that the operator's selected native-post behavior supersedes.
- Verify JSON escaping, Unicode, long fenced code/tables, and inline XML at
  ordinary block boundaries. Distinguish platform payload tests from real client
  rendering/notification evidence.
- Cover both post projections, legacy posts, mention placeholders and native
  tags, literal tags/images inside code, and ordinary image/file resources in
  source order. Verify omitted GET identity type and unchanged incoming mention
  admission. The external reviewer confirmed native-post peer receipt using a
  dispatcher submission log and the verified raw-mention gate; preserve this live
  evidence separately from the implementation's automated checks.
- Cover existing card `markdown`/`lark_md`, simplified `at` nodes, nested visible
  fields, plain-text lookalikes, and quoted/unquoted native card IDs. Test the
  structured/default merge with repeated primary mentions and overlapping views.
  Existing card-based peer communication remains the admission regression baseline;
  the live receive-side uncertainty belongs only to the new native-post path.
- Run monorepo build, lint, test, and `typecheck:tests`; run `.agents/scripts/check.sh`.
- Generate ordinary Rush change notes for the changed channel/transport contracts.
  This is not upgrade-blocking and needs no `BREAKING:` or `Rebuild:` wording.
- Update the channel domain and package docs; add the previously missing reply
  presentation entry to the product catalog and record the inbound XML change.
- Run an independent implementation review in the originating discussion after
  local verification, then prepare the reviewable PR against `next`.

## Known limits

Some native Markdown styles require a recent Feishu client. Unit tests can verify
request construction but not the actual native appearance or notification of a
person/peer bot; those require a live check. Native Markdown does not support
Setext headings, indented code blocks, email autolinks, raw HTML, or nested tables
and images within tables. Ordinary replies use one `md` node per row, respecting
the platform rule against mixing it with other inline nodes in that row. Feishu
now owns outgoing code-literal mention behavior; the initial probe exercises
tags in inline/fenced code and the supplement exercises empty names and bots.
No local mention-rewriting mechanism is added to compensate for native styling.
The flag research is complete at the
document/source/dry-run level, and no flag write or user OAuth is part of this
implementation.

## External review disposition

[Round-one review](https://github.com/excitedjs/dreamux/issues/410#issuecomment-5629682287)
requested changes. B1 now states the actual current caller and concrete future
empty-name XML; B2 is addressed by the shared inbound semantic parser across
text/post/cards and unified model-facing `user_id`. The operator explicitly
required existing peer-card compatibility after round one; the earlier text/post
only proposal was too narrow. Accept the existing-budget, single-adapter, code-literal,
public-send-contract, and native-subset corrections. Preserve long-table headers
with raw lines because dropping existing long-table delivery would be capability
loss; no Markdown writer is introduced. Do not add unrelated public-field
deletions. Peer-bot event evidence is tracked separately from the already accepted
five visual examples.

Round two returned
[sound-with-nits, no blocking findings](https://github.com/excitedjs/dreamux/issues/410#issuecomment-5629862950).
All six clarifications are incorporated above. The scope is already grounded in
the operator's reply change and subsequent explicit text/post/card compatibility
instructions. Implement them together; the feature must preserve communication
with peers still using existing card senders. Native-post peer receipt is
confirmed by the reviewer's receiver-side submission log plus the raw-mention
gate trace. The original webhook body was not captured in that evidence; its
mention contents are a supported inference from admission, not a quoted payload.
The supplementary empty-name client's visual appearance remains unconfirmed.

## Additional operator-requested identity experiment

The operator did not approve development on the authorization card, instead asking
whether the new native post could directly mention a peer bot by application ID.
Two direct-HTTP variants were tried: native `md` XML with the application ID in
`user_id`, and a dedicated post `at` node with the same value. Both posts were
accepted, but both send responses and default GET readbacks contained zero mention
records. The existing open-ID comparison has a real mention and verified peer
admission. The receiver independently reported no inbound submission or gate-drop
record for either app-ID probe, with normal neighboring events and positive
ordinary-mention controls. Its readbacks also contain no mention metadata. This
supports failure to trigger delivery; the original webhook was not captured, so
do not claim a complete negative delivery trace from log absence.

The experiment does not support replacing replyable bot identities with
application IDs in these tested post forms. Preserve the already verified open-ID
path, obtaining replyable IDs directly from default message reads rather than
adding an application-to-open-ID lookup. Keep the proposal pending the operator's
remaining questions; neither probe completion nor review grants development
authorization.

YourWildDad:
Replies currently carry a separate mention list through several adapters and convert Markdown into interactive cards. This change sends authored Markdown and native inline mentions as Feishu rich-text posts, preserves the initial topic reply address in newly provisioned Team identity, and returns actionable platform errors to the model.

Closes #410.

- Remove the separate mention parameters, unused Markdown-to-card renderer, and unused transport APIs. Keep raw interactive cards and native COT.
- Retain the 28 KiB serialized-content budget with ordered splitting at Markdown blocks, repeated fences/table headers, lines, and Unicode grapheme clusters.
- Parse text, structured posts, native Markdown posts, and existing cards into one ordered mention/code/resource representation. Prefer `content_v2` and render model-facing mentions with `user_id`; leave raw-event admission unchanged.
- Keep identity strings and stored formats. Append the configured identity and initial reply guidance at automatic Team creation, with persistence and restore coverage.
- Include Feishu status, code, reason, and log ID in send-error logs and MCP failures. Preserve unrelated error identities and exclude request bodies and credentials.

Validation: all four local Rush gates passed on `87dda034`, together with concrete splitter/parser replays, the complete MCP error boundary, and a live Codex compaction probe using synthetic reply addresses. Prior direct-HTTP native-post probes cover client rendering and peer wakeup within the limits recorded in the task verification. No Claude compaction result is claimed.

The branch was rebased onto `next` at `7cfd6ce5`. Current head `49ddf2c7` changes only documentation relative to the reviewed implementation: resolve the task-index conflict and align task records with the updated repository rules. `packages/` and `common/` are byte-identical to `87dda034`. Local knowledge-base and change-declaration checks passed; [all nine fresh CI checks](https://github.com/excitedjs/dreamux/actions/runs/34768321061) passed, including full Rush gates on Linux and macOS.

[Devbox approved `87dda034`](https://github.com/excitedjs/dreamux/pull/414#pullrequestreview-5180205413) and [approved current head `49ddf2c7`](https://github.com/excitedjs/dreamux/pull/414#pullrequestreview-5191359471) after independently verifying the documentation-only rebase, unchanged product files, and all nine fresh CI checks. Its one [non-blocking finding](https://github.com/excitedjs/dreamux/pull/414#discussion_r3990491736), overlapping text-mention placeholder keys, was reproduced on the baseline and recorded as a deferred follow-up; this PR does not fix it. No merge is claimed.

The operator-requested alpha is `@excitedjs/dreamux@0.26.0-alpha.g87dda034b746`, published by [release run 34610506606](https://github.com/excitedjs/dreamux/actions/runs/34610506606). After npm finished processing the dependency, isolated installation, service-entrypoint loading, and the installed native-post serializer check passed. The installed Dreamux, types, channel, and transport package versions all share `alpha.g87dda034b746`. The CLI exits successfully for `--version` but prints `unknown`; version verification used package manifests. No global installation or service restart was performed. The documentation-only rebase does not change this alpha's product code.

The task README, approved requirement, final solution, and verification are under `.agents/tasks/channel/simplify-feishu-replies/`.


packages/channel/feishu-transport/src/parse/mention.ts
```diff
@@ -0,0 +1,109 @@
...
+  }
+  return parts
+}
+
+function splitOnKey(
+  text: string,
+  key: string,
+  resolved: InboundContentPart,
+): InboundContentPart[] {
+  const pieces = text.split(key)
```
ryanxiang7, line 72: Nit (pre-existing, non-blocking — I am approving anyway): the iterative `text.split(key)` walk is sensitive to placeholder *prefix overlap*. Feishu keys are `@_user_N` in the records' delivery order, so a single text message that mentions ten or more users contains both `@_user_1` and `@_user_10`. When the `@_user_1` record is processed, its key also matches inside `@_user_10`: that token splits into this resolved mention part plus a literal `0`, and when the `@_user_10` record is later processed, `splitOnKey` only scans `text` parts, so it substitutes nothing. The tenth @-position is then projected as a structured mention of user 1 followed by `0` (and `@_user_11` similarly becomes mention #1 + `1`). In the old renderer this came out as wrong plain text; here the projection is structured, so it becomes a wrong-identity `<at>` in the channel serialization — same root defect, slightly sharper consequence.

The baseline `applyMentions` did the identical substring walk, so this is ported behavior rather than a regression, and the trigger (10+ distinct @-mentions in one plain text message) is rare; hence non-blocking. A minimal fix that also deletes the iterative pass: scan once with one alternation of all present keys, longest first and regex-escaped — `new RegExp(keys.sort((a, b) => b.length - a.length).map(escapeRegExp).join('|'), 'g')` — and resolve each hit through `findMentionRecord`. Happy for this to land as a follow-up.

ryanxiang7 (approved):
Independent review of the full diff against `next` (merge base 27428f8), focused on complexity, deletion completeness, and capability/requirement parity. I read the task inputs under `.agents/tasks/channel/simplify-feishu-replies/` and reviewed the product change file by file. One non-blocking nit is attached inline; nothing here blocks merge.

**Deletions are complete.** Repo-wide greps find no remaining production references to the mention-list argument, `conversationKey`, `editText`/card fallback, the Markdown-to-card renderer exports, `textWithLeadingMentions`, or `optionalStringArray`. The outbound contract is down to `chatId` + optional `replyToMessageId`, and `package-boundary-guards` was tightened in lockstep with `index.ts` rather than loosened. COT and explicit `sendCard`/`editCard` keep their real callers and share the same 28 KiB serialized-byte guard; no retained capability lost its path.

**Outbound rewrite matches the approved design.** `nativePostContents` tries the whole body first and only splits an oversized one, on marked block `raw` seams (whole blocks → fenced code with repeated fences → table rows with the exact two source header lines → lines → grapheme clusters). I empirically verified against marked 15.0.12 that block tokens tile the source exactly via `.raw`, and exercised the splitter: pieces re-concatenate to the authored content modulo the designed fence/header repetition (output is intentionally longer, never lossy), whitespace-only pieces are dropped, and a header-plus-one-oversized-row table fails with an actionable error. Byte accounting uses JSON-escaped length, which is what the platform actually receives.

**One send boundary, honest errors.** `sendFeishuMessage` branches only on create vs reply; a rejection is treated as a Feishu envelope only when `err.response.data.code` is numeric (gateway HTML keeps its identity), a resolved nonzero business code fails the send, and the description carries operation/HTTP/code/log_id/msg with the cause preserved and no body or credentials.

**Inbound parsing covers the four shapes without inventing meaning.** text placeholders, v1 `at` nodes, native `content_v2` (preferred, never concatenated with legacy `content`), and card markdown/lark_md/plain fields all resolve occurrence position against the same mention records; post uses `user_id`, card uses `id`, and the wrong dialect's attribute stays literal. Escapes, code spans (matched-length backticks), and fenced code (tilde/backtick, info-string guard, unterminated-to-EOF) are settled before mentions, so `<at>` in examples cannot resolve; message-resource `img_*` keys become attachments while web images stay literal; app_id records keep key/name with no identity slot; the dual-view card merge is line-scoped with per-line identity sets. The raw-event mention gate is untouched — parsing happens before it but does not feed it.

**Tests were strengthened, not weakened.** The CDATA truncation tests moved onto real code parts rather than being deleted; new tests pin every parser dialect/code-literal/identity/merge scenario, hostile id/name XML escaping in the channel serializer, the three-field tool contract (unknown args dropped, verbatim body, thread target), partial-send observer ordering, cancellation, non-envelope rethrow, and the creation-time identity guidance (absent/custom identity, chat vs anchor, per-topic, concurrency, replay, restore-without-reappend). Changelog types match the version-line policy (channel minor on 6.x, transport minor on the 0.x line with no `BREAKING:` since this is not upgrade-blocking, dreamux none). The `.agents` delta and product README describe the shipped behavior accurately; committed probes take all identities from runtime args and use only synthetic placeholders in docs.

All 9 CI checks are green at this commit.

ryanxiang7 (approved):
Re-review after the rebase onto current `next` (`7cfd6ce`, which includes #413). I independently verified rather than taking the byte-identical claim on faith.

**Product change is byte-identical to the approved commit.** `git diff 87dda034 49ddf2c -- packages/ common/` is empty; every product source, test, package manifest, and change note has the same blob as the commit I approved. The only differences between the two heads live under `.agents/`, and I checked each:

- The parent of this commit is exactly the current tip of `next` (`7cfd6ce`), and the PR is MERGEABLE again (the prior CONFLICTING state is resolved).
- The only files this commit itself changes beyond the reviewed product change are the task's own `README.md`/`verification.md` reconciliation plus its one-line index entry. The diffs against the new parent for `task-records.md`, `SKILL.md`, `development-approval.md`, `implementation.md`, `domains/channel.md`, and `product/README.md` are the same content I already approved at 87dda034; what differs there versus the old head is entirely #413's baseline.
- The reconciliation follows #413's trunk record rule correctly: `State: done` (a merge-proof state), the index line states the goal with no mutable status, no PR/CI merge result is asserted, and the review/alpha sections record dated evidence with links instead of mirroring live status. The N1 placeholder-prefix finding is reproduced and recorded honestly as inherited/non-blocking with its inline-comment link; nothing claims it was fixed.
- The other task READMEs that differ between heads were changed by #413 on the base, not by this commit.

**CI at the new head: all 9 checks green** (rush ubuntu/macos, gitleaks, internal-content tree scan, rush change, knowledge-base check, shellcheck x2, author email). The cited alpha release run 34610506606 checks out as a successful run at the reviewed commit; the verification note correctly treats the CLI `--version` output as non-evidence rather than dressing it up.

My earlier approval stands and is reconfirmed at this commit. The non-blocking N1 nit (≥10 mentions, `@_user_N` prefix overlap) remains the only open item and is appropriately tracked as a follow-up.

YourWildDad:
Superseded by #424, which carries this PR's outbound commit unchanged as its first commit and redoes the inbound side as one text body per message (see the task record under .agents/tasks/channel/read-feishu-inbound-as-text/).

PR #422, opened 2026-09-13
Use native Feishu replies with a parts-only inbound body

YourWildDad:
## Summary

- Carry the original native Feishu reply work from #414 onto `next`: native post replies, inline XML mentions, actionable send errors, inbound reply context, and creation-time reply-address guidance. This also carries the original removal of the unused card renderer and outbound edit API, plus its workflow and knowledge records.
- Make required ordered parts the only inbound body representation. Remove parallel `text` / `parsedText` / `compatibilityText` and resource projections, conversion wrappers, and body-reconstruction fallbacks without removing supported inbound formats, attachments, mentions, reply splitting, or reply guidance.
- Publish the preserved implementation for renewed discussion, with its requirement, design, review adjudication, and Rush change declarations. This draft does not close or update #414 and is not a merge-ready replacement.

Related solution discussion: #410. Original work: #414.

## Draft limitations

- **Wrong mention identity:** `parse/mention.ts` matches `@_user_1` inside `@_user_10` when processed first. A synthetic ten-mention replay reproduced the first person's identity plus literal `0` instead of the tenth person. This is present in the original PR and remains unfixed here.
- **R2:** an unreachable attachment map-miss fallback remains. The TeamLeader accepted the cleanup finding; no correction has been authorized.
- **R5:** card tests inspect only text for hidden values and allow additional attachments, reducing coverage. No current leak was observed. The operator declined this test correction for now, so the assertions remain unchanged.
- **R8:** three slash-command fixtures still supply flat mention text instead of real mention-plus-text parts. A fixture correction has not been authorized.
- **Full tests are not green:** the latest root run had five Codex-live failures. A prior baseline control reproduced four failing case names, not identical causes, and does not explain the fifth continuity timeout. These failures are not established as solely environmental.
- Independent implementation review completed, but its findings and verification limits have not been cleared for acceptance. Further corrections, merge, deployment, and closing #414 are outside this publication request.

## Test plan

- [x] Rebase onto `next` at `7cfd6ce5`; resolve only the channel task-index conflict and verify preserved package contents against the saved work.
- [x] Fresh root Rush build (eight up-to-date operations), lint, and `typecheck:tests` (seven successful operations each).
- [x] Previous transport and channel suites: 205 and 488 passing tests; 17 direct built-module semantic scenarios passed.
- [x] Final `.agents/scripts/check.sh` and `git diff --check`.
- [x] Mandatory repository pre-commit gates, staged secret scan, full indexed-tree internal-content scan, and secret scan of the rebased original commit. Ordinary commit hooks ran without a bypass.
- [ ] Resolve the five live-Codex failures and obtain a fully green root Rush test run. The full suite was not rerun after the documentation-only rebase; the latest run had 1,083 Dreamux passes, all 73 non-live Dreamux files passing, and the six other package operations passing.
- [ ] Resolve the named review items under operator direction and complete review acceptance.
- [ ] Check the new PR's own CI; results from #414 do not validate this head.

Dependency preparation succeeded in the developer's environment. The TeamLeader's later Rush update encountered a managed-hook installation conflict; no Git configuration or safeguards were changed. Both published commits were uploaded through the GitHub API after the Git transport endpoint was unreachable, with tree and commit hashes verified byte-for-byte against the locally checked commits.


YourWildDad:
Superseded by #424: the inbound side is redone as one text body per message instead of an ordered parts list (see the task record under .agents/tasks/channel/read-feishu-inbound-as-text/).

PR #423, opened 2026-09-13
Filter list_bindings by target and Team

YourWildDad:
## What

`list_bindings` answered exactly one question — the whole routing table. An
operator asking where one Team answers, or what is bound in one chat, read every
row and picked; and a collaboration space installs a binding per provisioned
topic, so that table is as long as the space is busy.

Four optional filters now narrow the same read:

| Parameter | Question it answers |
| --- | --- |
| `team_name` | Which conversations does this Team answer in? |
| `chat_id` | What is bound in this chat — its own row and every topic row under it? |
| `thread_id` | Which Team answers in exactly this topic? |
| `target_kind` | Whole-chat bindings only, or topic bindings only? |

They are matched exactly, combine with AND, and carry no authority of their own:
they select rows the caller could already see. No filter still returns
everything, matching nothing is an ordinary empty answer, and the result shape
is unchanged.

`target_kind` earns its place because exact equality on `thread_id` cannot say
"is null" — without it, "only the whole-chat bindings" is unreachable. It offers
`group` and `topic` only: a bind takes a chat id and an optional thread id, so
`selectorTarget` stores nothing else, and an offered `p2p` would be an option
that always answers with nothing.

The filter lives in the tool definition, over the rows the session already
returns. Pushing it into `FeishuRouting.listBindings` would add a second query
vocabulary plus a translation to the snake_case view rows, for no gain while no
in-process caller wants one.

Declined as out of scope by the operator: an `origin` filter, a `space_name`
filter, pagination, fuzzy `display` matching, and resolving which Team
*effectively* answers in a topic (the parent-chat fallback is route resolution,
not a table read — a separate tool if it is ever wanted).

## Also here

A second commit strengthens the `dev-workflow` skill: every request for the
operator's *authority* — task creation, solution path, development approval,
merge, dissolution — must be an interactive question card, not chat text.
Clarification questions already had that rule; the decisions that actually move
the workflow did not.

## Verification

- New `list_bindings` unit block over a four-row fixture, covering every
  acceptance criterion in the task record. Two of them invoke the tool through
  the registered Dispatcher catalog, so what is exercised is the schema a
  Dispatcher is actually advertised.
- `rush build`, `rush lint`, `rush test`, `rush typecheck:tests` all green.
- Rush change file for `@excitedjs/feishu-channel`, ordinary note, no upgrade
  impact.

Task record: `.agents/tasks/channel/add-list-bindings-filters/`


ryanxiang7 (approved):
Reviewed both commits against source at f78f9785.

**Code**
- The filters live at the right layer: they narrow rows the Dispatcher could already read, so they add no authority and touch no route resolution. Keeping them in the tool definition rather than adding a second query vocabulary to `FeishuRouting.listBindings` is the correct entropy call while no in-process caller wants one.
- `optionalLiteral` is a proper capability beside the existing validators: the JSON-schema enum and the parse-time refusal agree, and the `PublicInvokeFailure` sentence names the accepted values.
- AND intersection via a single predicate over `null` (absent) members is minimal and correct; the closed object schema, empty-string-as-absent, and the defensive row copy all match the file's existing conventions.

**The p2p claim was independently traced**
- Bind tools: `selectorTarget` (feishu-session-bindings.ts:220) can only return `chatTarget(...,'group')` / `topicTarget` — the `isBindableTarget` guard at line 68 is genuinely unreachable on this path.
- Provisioning: its only caller is feishu-channel.ts:417, gated by `FeishuRouting.plan`, which returns `{kind:'dispatcher'}` for a non-bindable target before a provision plan exists (routing/index.ts:124).
- Conclusion holds: excluding `p2p` from the enum rests on the product rule both writers keep, and `FeishuTargetRecord.kind` still typing it is stated accurately in the comment and task record. Recording the unreachable guard as a cleanup-trail non-goal instead of quietly deleting it is the right discipline.

**Skill change (commit 2)**: every authority-ask site (task select/create, solution path, dev approval, merge, dissolution) now points at the card rule; task-discovery's single card sentence covers all three branches, and SKILL.md step 10 is covered by the new global rule plus team-dissolution.md. Consistent with the existing requirement-clarification convention.

**Verification**
- Independent local run: rush build (feishu-channel + deps) green; the target test file passes 29/29; `.agents/scripts/check.sh` green; CI 9/9 green on f78f9785.
- Change file is an ordinary minor note for a post-1.0 package — additive tool input, no upgrade blocker, no BREAKING needed. Knowledge-closeout claim checked: the three pages that name `list_bindings` state no input shape, so no page went stale.

No blocking findings. Approving.

ryanxiang7 (changes requested):
## Rebase needed after #424 landed on `next`

The reviewed code is unchanged (HEAD `f78f9785`, approved 2026-09-13 with CI 9/9 green), but `next` advanced via #424 (`d1bd1f1`) and this PR now has three content conflicts. They are mechanical: I resolved them against current `next` locally, and the resulting tree builds, lints, and passes the full `@excitedjs/feishu-channel` test suite. Please rebase onto `next`:

1. `packages/channel/feishu-channel/src/tools/schema.ts` — keep `optionalLiteral`; **do not** re-add `optionalStringArray`. Its only consumer (`send_messages` in `messaging-tools.ts`) was removed by #424, so re-adding it would leave dead code.
2. `.agents/tasks/channel/README.md` — keep both sides of the index-bullet hunk (#424's two entries plus this PR's `add-list-bindings-filters` entry).
3. `.agents/skills/dev-workflow/references/development-approval.md` — take the `next` side. #424 renamed the interactive-card tool to `ask_user_question` and strengthened the authorization-card wording; this branch's `ask_human_question` text is the older version.

`.agents/skills/dev-workflow/SKILL.md` auto-merges cleanly.

After the rebase is pushed, the nine CI checks must run green against the new head, and I will re-approve. The `list_bindings` implementation and its tests need no changes.

ryanxiang7 (changes requested):
Rebase needed after #424 landed on next

The reviewed code is unchanged (HEAD f78f9785, approved 2026-09-13 with CI 9/9 green), but next advanced via #424 (d1bd1f1) and this PR now has three content conflicts. They are mechanical: I resolved them against current next locally, and the resulting tree builds, lints, and passes the full @excitedjs/feishu-channel test suite. Please rebase onto next:

1. packages/channel/feishu-channel/src/tools/schema.ts — keep optionalLiteral; do NOT re-add optionalStringArray. Its only consumer (send_messages in messaging-tools.ts) was removed by #424, so re-adding it would leave dead code.
2. .agents/tasks/channel/README.md — keep both sides of the index-bullet hunk (#424's two entries plus this PR's add-list-bindings-filters entry).
3. .agents/skills/dev-workflow/references/development-approval.md — take the next side. #424 renamed the interactive-card tool to ask_user_question and strengthened the authorization-card wording; this branch's ask_human_question text is the older version.

.skills/dev-workflow/SKILL.md auto-merges cleanly.

After the rebase is pushed, the nine CI checks must run green against the new head, and I will re-approve. The list_bindings implementation and its tests need no changes.

ryanxiang7 (approved):
Re-reviewed the rebased head **57df9039** against current `next` (`d1bd1f1`, post-#424). The three conflict resolutions were verified individually rather than taken on description:

1. **schema.ts** — only `optionalLiteral` is added; `optionalStringArray` is correctly left dead (I grepped the full rebased tree: zero references; #424 removed its sole consumer `send_messages`). Re-adding it would have been zombie code; the resolution is right.
2. **tasks/channel/README.md** — pure end-append, both #424 bullets and this PR's bullet present.
3. **development-approval.md** — #424's stronger hard-gate text kept, plus the one genuinely missing sentence (playback may be prose; the request itself must be the card). Read in full, it is complementary, not contradictory: #424's section enforces the dev-approval point, this PR's new SKILL.md rule fixes the scope across all five authority asks.

**The p2p conclusion was re-traced on the new inbound chain #424 reshaped.** `FeishuTargetRouter.project` builds `chatTarget(place.chatId, chatType)` for a direct chat, so the p2p kind survives into routing; `FeishuRouting.plan` still returns `{kind:'dispatcher', reason:'not_bindable'}` before any provision plan (routing/index.ts:124), and `provisionForInbound` still has exactly one caller (feishu-channel.ts:417). The bind-tool path is unchanged (`selectorTarget` still spells only group/topic). Conclusion stands on the new baseline.

**Verification**: local build, lint, typecheck:tests all green on 57df9039; target test file 29/29; `.agents/scripts/check.sh` green; CI 9/9 green.

One non-blocking observation, no action required in this PR: the merged skill uses `ask_human_question` as the generic name (with the inline bridge "the Feishu channel exposes it as `ask_user_question`") while #424's dev-approval hard gate uses the Feishu-specific name. This is coherent under the existing glossary in requirement-clarification.md, but a future name-unification pass over the skill could pick one; folding it into the dated-quote cleanup already planned for operator decision would be natural.

Approving.

PR #425, opened 2026-09-15
Name the card tool once and drop the dated quotes

YourWildDad:
Two operator rulings of 2026-09-15, both on the `dev-workflow` skill's prose.

## One name for the card tool

The skill named the interactive question tool `ask_human_question` in five
places and `ask_user_question` in two, bridged by a sentence in
`requirement-clarification.md` explaining that the first is the operator's name
and the second is what the Feishu channel exposes. Following a rule about one
tool meant carrying two names and the mapping between them.

Every site now uses `ask_user_question`, the name the tool actually has, and
the bridging sentence is gone. The trade is stated rather than hidden: the
skill is now written against the Feishu surface, so a channel exposing this
tool under a different name will have to rewrite these five lines. The operator
chose that trade knowing it.

## No dated quotes in a current-state page

`requirement-clarification.md` carried a dated operator ruling under the
clarification rule, and `teammate-briefs.md` carried a dated operator reaction
as the label on a good-case handoff. Neither added anything its surrounding
prose did not already state — the clarification rule is written out in full
directly above where its quote sat — while both made a reader judge whether a
dated line was a live instruction or preserved material. A settled decision's
attribution belongs in the evidence layer and in git history, not in the page
that states the current rule.

## Verification

- `.agents/scripts/check.sh` green.
- `rush build`, `rush lint`, `rush test`, `rush typecheck:tests` green.
- `git grep ask_human_question` returns nothing repo-wide; a scan of
  `.agents/skills/` for any `YYYY-MM-DD` text returns nothing. Both methods
  cover the form they name and not others.

No package source changed, so there is no Rush change file.

Note for sequencing: PR #423 is open and on hold at the operator's direction,
and it adds four more `ask_human_question` mentions. Whichever of the two lands
second has to convert the other's lines; if #423 goes second, its rebase does
it.


ryanxiang7 (approved):
Review complete. The change is docs-only and consistent:

- All seven skill sites now use `ask_user_question`, matching the actual Feishu MCP tool name (`packages/channel/feishu-channel/src/tools/ask-user-question.ts`); `git grep ask_human_question` on the branch is empty, including the four sites #423 added after this branch opened (the second commit converts them).
- The naming bridge is gone and the trade-off is explicit: the skill is bound to the Feishu surface. The other KB references (channel.md, task records) already use the single name, so the repo is uniform.
- The two dated quotes are removed from current-state pages whose surrounding prose already states the rules in full; dated operator evidence remains correctly preserved under `.agents/tasks/`, and no `YYYY-MM-DD` strings remain in `.agents/skills/`.
- `.agents/scripts/check.sh` is green; no package source changed, so no Rush change file is needed.

No blocking findings.

Issue #426, opened 2026-09-15
Solution review: Feishu /bind and /help slash commands


YourWildDad:
Technical-solution review requested. **No code has been written yet** — this reviews the design, before the development gate.

Task record (branch `dreamux/team-space-dreamux-zhc1wl`):

- [requirement.md](https://github.com/excitedjs/dreamux/blob/dreamux/team-space-dreamux-zhc1wl/.agents/tasks/channel/add-bind-and-help-slash-commands/requirement.md)
- [technical-design/draft.md](https://github.com/excitedjs/dreamux/blob/dreamux/team-space-dreamux-zhc1wl/.agents/tasks/channel/add-bind-and-help-slash-commands/technical-design/draft.md)

## Scope

Add `/bind <team_name>` and `/help` to the Feishu slash-command table.

Settled by the operator, **not open for re-litigation** in this review:

- `/bind` always binds the chat-level target, never the topic it was typed in;
- a chat registered as a Collaboration Space container refuses the command;
- rebinding an already-bound chat is ordinary, not an error;
- `yargs-parser` is taken as the argument parser. This overruled a recommendation to add no library at all; the recommendation and its data are kept in the draft's last section for the record. Please review *how* it is used, not *whether* to use it.

## Please challenge these

1. **Changing the binding operation's parameter type** (draft 4a). `bindChannel`/`unbindChannel` would take a `FeishuTarget` instead of a `FeishuBindTargetSelector`, moving `selectorTarget` down into `tools/routing-tools.ts` where the selector is an MCP wire shape. The motive: `selectorTarget` maps a bare `chatId` to a `group` target, so `isBindableTarget`'s p2p guard is unreachable today — the repository already documents this in the header of `tests/feishu-session-bindings.test.ts`. Is the signature change worth it, are there call sites the draft missed, and does it make the tool layer worse? The rejected alternative is a p2p check inside the command itself.
2. **`containingChat()` as a new export in `routing/target.ts`** — or should this reuse `resolutionChain` rather than adding a function next to it?
3. **`parse-positional-numbers: false`** (draft 2). Measured on `yargs-parser@21.1.1`: `1e5` parses to `100000` and `0x1f` to `31` by default, and both are legal team names under `validateTeamId`. Is that reading right, and is there any other yargs-parser default that damages a legal team name (dots, underscores, dashes, the 64-character ceiling)?
[...]

PR #428, opened 2026-09-15
Add Feishu /bind and /help slash commands

YourWildDad:
Closes the solution review in #426.

An operator can now point a Feishu group at a Team by typing `/bind <team_name>`, instead of asking an agent to call a tool, and `/help` says what the Channel answers.

## What changed

**The command table stops being a name-only table.** A row carries its own `usage` and `summary`, so `/help` renders the table rather than a second list that can drift from it — a row the Channel can run is a row `/help` names. Recognition hands the row a parsed invocation instead of discarding everything after the token, and the per-row failure label is gone: dispatch builds `Command /<name> failed` from the key it looked the row up by.

**`yargs-parser` does the parse**, with `parse-positional-numbers` off. Left on, it destroys legal Team names — `1e5` → `100000`, `2.50` → `2.5`, `1.` → `1`, all of which `validateTeamId` accepts and `String()` cannot restore. The remainder is sliced from the original text rather than the lowercased copy used for matching, so `/bind MyTeam` binds `MyTeam`. This is the package's first non-workspace runtime dependency, so its `CLAUDE.md` dependency boundary and `package.json` description are updated in the same change.

**`/bind` binds the containing chat**, so typing it inside a topic routes the group. It validates nothing itself: a malformed name, a missing or closed Team, and a direct message are each refused in the owning layer's own words. A chat registered as a Collaboration Space is refused outright, because a Space already gives each of its topics a Team.

**The binding operations take a `FeishuTarget`.** They took a `{ chatId, threadId? }` selector and converted it, which made every bare chat id a `group` target and left `isBindableTarget`'s direct-message guard unreachable — a fact the test file's own header used to document. The guard is now live, and the selector type is deleted rather than relocated, since the tools can build a target in one hop. An MCP caller still supplies a bare `chat_id` with no kind, so `bind_channel` still reads one as a group; only inbound projection knows a chat's kind, and that limitation is recorded rather than papered over.

**A bind answers where it was asked.** `bindChannel` gained an optional `announceIn`; a slash bind supplies the conversation the command was typed in, so the card reaches the operator's own topic instead of opening a topic of its own. An MCP-initiated bind supplies nothing and is unchanged.

**A card sent away from the target it announces carries no anchor Team.** A topic can hold a binding row of its own in an ordinary chat, so without this the newly bound Team's COT fallback anchor would be seated inside another Team's conversation. `LeaderLifecycleFence.blocksAnchor` does not catch it: it fences a closed leader, or a target that leader released, and a fence one Team raised says nothing about another.

**A rebind names the Team it displaced** — a `Previous Team` line on the binding card, on every path that binds, absent on a first bind and on rebinding to the Team that already holds the chat.

## Verification

`rush update`, then `build`, `lint`, `test`, `typecheck:tests` — all green, re-run independently rather than taken from the implementer's report. Feishu channel: 550 tests. The three `SUCCESS WITH WARNINGS` operations under `test` are pre-existing stderr logging in the Codex runtime, Claude Code runtime, and Dreamux packages; `feishu-channel` is not among them. No live Feishu chat was exercised.

Every acceptance criterion in the task record has a test, including the anchor case above, the numeric-looking names, and that adding a row to the table makes it appear in `/help` with no second edit.

## Record

Requirement, final solution, and the finding-by-finding review adjudication are in `.agents/tasks/channel/add-bind-and-help-slash-commands/`. Two corrections are recorded there rather than only in conversation: the solution claimed `bindChannel` already showed the previous Team on its card (it did not — the sentence came from misreading `routing-tools.ts`'s "the previous owner is reported back", which describes the tool's result field), and it claimed dotted Team names were safe under either parser setting (they are not).


ryanxiang7 (approved):
## Implementation review — approved

Re-checked every finding from #426 against the code on this PR, and re-ran the gates independently in a clean worktree at `cb8e709`: `vitest run` for `@excitedjs/feishu-channel` — **550/550 pass (28 files)**; `tsc` (src and tests), `eslint .`, and `.agents/scripts/check.sh` all clean.

### The anchor adjudication (the one to re-check) — you are right, and I am not

I re-read my proposed gate against your fix, and your rejection holds. Gating `announceIn` on the topic holding a binding row sends the card back to the container chat, which opens a new topic — precisely the behavior this PR removes — so it cures the anchor bug by resurrecting the placement bug in exactly the sub-case that motivated the change. Splitting the two halves is the correct decomposition, and tracing the code confirms it is implemented as designed:

- `feishu-session-bindings.ts`: `const announce = input.announceIn ?? target; notify(announce, card, sameTarget(announce, target) ? input.teamName : null)` — the destination moves, the anchor does not.
- COT release/claim still take the bound `target`; nothing about fence ownership moves.
- `onNotificationSent` early-returns on a null anchor Team while still `observe`-ing the card into the target it physically landed in — which is also correct: that message genuinely is the newest one seen in that topic, and future topic-routed cards may legitimately reply under it.
- An omitted `announceIn` is byte-identical old behavior, pinned by the MCP-path test (`keeps MCP notifications in the bound target and supplies a fallback anchor there`, including the card's `originMessageId`).

One more point in favor of your version that neither of us wrote down: a finer-grained gate ("seat the anchor when the requester topic resolves to the new Team after the bind") would carry a *later* hazard — if that topic is subsequently bound to another Team, the first Team's anchor stays parked in a conversation that no longer routes to it, and `blocksAnchor` would not retire it (the Team released the group route, never the topic). Seating no cross-target anchor avoids that whole class. The residual cost is bounded: in the benign case (unbound topic, chat-level bind), the newly bound Team gets no fallback anchor, so a card it emits before the next human message in that topic is not shown; the next inbound establishes the anchor through `beginInboundSubmission`, and the pre-PR behavior in that case opened a junk topic anyway. The `it.each([false, true])` topic test pins this trade-off deliberately for both branches. Good call.

The product sub-question (the "group bound to B" receipt appearing inside A's separately bound topic): acceptable as-is. The card names the bound *group* and its target explicitly, so it reads as the fact the operator asked for, and it carries no anchor — a receipt, as the doc says.

### The five requested spots

1. **Tests were not weakened.** I diffed both files line by line. The existing assertions changed only in shape (`{ chatId }` → `chatTarget(..., 'group')`), and the load-bearing expectations are intact or stronger: the old "guard is unreachable" header is replaced by a test that asserts the p2p refusal happens **before Core is queried, with zero rows and zero COT calls** (`expect(h.calls).toEqual([])` is stronger than the old comment ever was); rebind ordering keeps the full `cotCalls` sequence assertion including the same-team re-bind's extra `claimed`; the wire-conversion test newly locks `null`/`''`/real thread id through both bind and unbind, which matters because `bindTarget` dropped the old explicit `''` branch — parity now rests on `optionalString` coercing `'' → null`, which I verified in `schema.ts:52`. The missing/closed-Team tests still assert rows unchanged and zero cards.
2. **`displaced` narrowing and conditional spread.** Fine in this house's taste. TS 5.7 narrows `previousTeamName` to `string` inside `if (displaced)` via const aliased conditions (the typecheck proves it), and conditional spread is already the established local idiom — the same method uses `...(input.requireOwner !== undefined ? { requireOwner } : {})`, and the card uses the array form right next to it. The derived boolean names the product fact ("a different Team was displaced") and is reused consistently for fence release *and* the card line; first bind and same-team rebind are both absent-line, matching the operator ruling.
3. **The `input` shadow in `command()`.** Leave it. The arrow is a one-line pass-through that never references the outer `input`, and the identical wrapper already exists in `toolSession()`; renaming only the new one would make the two look different when they are the same. Lint agrees (no shadow rule fires).
4. **The rewritten reconcile comment.** Accurate. The old sentence ("a command … does not also correct this Channel's routing document") was falsified by `/bind`; the replacement scopes the claim to what survives — a *rejected Core command* is not proof a route is finished — and the `TEAM_CLOSED`-still-pending explanation that follows is unchanged and still the reason neither `/dissolve` failure nor `/bind`'s pre-validation refusal may touch rows. It also stays correctly placed above the `invoke` callback it explains.
5. **Knowledge vs code.** Consistent. I checked the three load-bearing statements in `channel.md` against the implementation: (a) "a bind may say otherwise" + `announceIn` defaults — matches the code; (b) route release/claim always use the bound target — matches; (c) the cross-target card "is observed … in the target it was actually sent to" — matches `onNotificationSent(announce, …)`. The routing-tools paragraph now distinguishes the tool's `previous_team_name` result field from the card line, which is exactly the misreading the task record owns up to. `product/README.md`, the package `CLAUDE.md` boundary (both the prose sentence and the allowed-upstream list), `package.json` description, and the rush change file (`minor`, correct for a 6.x package) all agree.

### Nits (non-blocking, no change requested to merge)

- `final.md` §1 still writes the type as `yargsParser.Arguments` while the code imports the binding as `parser` and uses `parser.Arguments`. Doc-only; code is self-consistent.
- `/bind`'s expected refusals (bad/missing/closed Team, DM) surface as `Command /bind failed: …` through the dispatch catch, so an operator-facing refusal wears a "failed" label unlike the in-command Space refusal. This is the deliberate layering outcome — the binding layer owns the sentence — and it is tested; noting it only so it stays a conscious choice.

Nothing else in the diff is out of place: lockfile entries point at the already-present `21.1.1`/`21.0.3`, `FeishuBindTargetSelector` is fully gone (no dangling references, including `index.ts`), `containingChat` is the one-line structural capability from the design, and the parser is only ever fed positionals (`args._`), with the no-throw behavior locked for flag-shaped and backslash input outside the dispatch catch.

Approving.

ryanxiang7 (approved):
## Implementation review — approved

Re-checked every finding from #426 against the code on this PR, and re-ran the gates independently in a clean worktree at `cb8e709`: `vitest run` for `@excitedjs/feishu-channel` — **550/550 pass (28 files)**; `tsc` (src and tests), `eslint .`, and `.agents/scripts/check.sh` all clean.

### The anchor adjudication (the one to re-check) — you are right, and I am not

I re-read my proposed gate against your fix, and your rejection holds. Gating `announceIn` on the topic holding a binding row sends the card back to the container chat, which opens a new topic — precisely the behavior this PR removes — so it cures the anchor bug by resurrecting the placement bug in exactly the sub-case that motivated the change. Splitting the two halves is the correct decomposition, and tracing the code confirms it is implemented as designed:

- `feishu-session-bindings.ts`: `const announce = input.announceIn ?? target; notify(announce, card, sameTarget(announce, target) ? input.teamName : null)` — the destination moves, the anchor does not.
- COT release/claim still take the bound `target`; nothing about fence ownership moves.
- `onNotificationSent` early-returns on a null anchor Team while still `observe`-ing the card into the target it physically landed in — which is also correct: that message genuinely is the newest one seen in that topic, and future topic-routed cards may legitimately reply under it.
- An omitted `announceIn` is byte-identical old behavior, pinned by the MCP-path test (`keeps MCP notifications in the bound target and supplies a fallback anchor there`, including the card's `originMessageId`).

One more point in favor of your version that neither of us wrote down: a finer-grained gate ("seat the anchor when the requester topic resolves to the new Team after the bind") would carry a *later* hazard — if that topic is subsequently bound to another Team, the first Team's anchor stays parked in a conversation that no longer routes to it, and `blocksAnchor` would not retire it (the Team released the group route, never the topic). Seating no cross-target anchor avoids that whole class. The residual cost is bounded: in the benign case (unbound topic, chat-level bind), the newly bound Team gets no fallback anchor, so a card it emits before the next human message in that topic is not shown; the next inbound establishes the anchor through `beginInboundSubmission`, and the pre-PR behavior in that case opened a junk topic anyway. The `it.each([false, true])` topic test pins this trade-off deliberately for both branches. Good call.

The product sub-question (the "group bound to B" receipt appearing inside A's separately bound topic): acceptable as-is. The card names the bound *group* and its target explicitly, so it reads as the fact the operator asked for, and it carries no anchor — a receipt, as the doc says.

### The five requested spots

1. **Tests were not weakened.** I diffed both files line by line. The existing assertions changed only in shape (`{ chatId }` → `chatTarget(..., 'group')`), and the load-bearing expectations are intact or stronger: the old "guard is unreachable" header is replaced by a test that asserts the p2p refusal happens **before Core is queried, with zero rows and zero COT calls** (`expect(h.calls).toEqual([])` is stronger than the old comment ever was); rebind ordering keeps the full `cotCalls` sequence assertion including the same-team re-bind's extra `claimed`; the wire-conversion test newly locks `null`/`''`/real thread id through both bind and unbind, which matters because `bindTarget` dropped the old explicit `''` branch — parity now rests on `optionalString` coercing `'' → null`, which I verified in `schema.ts:52`. The missing/closed-Team tests still assert rows unchanged and zero cards.
2. **`displaced` narrowing and conditional spread.** Fine in this house's taste. TS 5.7 narrows `previousTeamName` to `string` inside `if (displaced)` via const aliased conditions (the typecheck proves it), and conditional spread is already the established local idiom — the same method uses `...(input.requireOwner !== undefined ? { requireOwner } : {})`, and the card uses the array form right next to it. The derived boolean names the product fact ("a different Team was displaced") and is reused consistently for fence release *and* the card line; first bind and same-team rebind are both absent-line, matching the operator ruling.
3. **The `input` shadow in `command()`.** Leave it. The arrow is a one-line pass-through that never references the outer `input`, and the identical wrapper already exists in `toolSession()`; renaming only the new one would make the two look different when they are the same. Lint agrees (no shadow rule fires).
4. **The rewritten reconcile comment.** Accurate. The old sentence ("a command … does not also correct this Channel's routing document") was falsified by `/bind`; the replacement scopes the claim to what survives — a *rejected Core command* is not proof a route is finished — and the `TEAM_CLOSED`-still-pending explanation that follows is unchanged and still the reason neither `/dissolve` failure nor `/bind`'s pre-validation refusal may touch rows. It also stays correctly placed above the `invoke` callback it explains.
5. **Knowledge vs code.** Consistent. I checked the three load-bearing statements in `channel.md` against the implementation: (a) "a bind may say otherwise" + `announceIn` defaults — matches the code; (b) route release/claim always use the bound target — matches; (c) the cross-target card "is observed … in the target it was actually sent to" — matches `onNotificationSent(announce, …)`. The routing-tools paragraph now distinguishes the tool's `previous_team_name` result field from the card line, which is exactly the misreading the task record owns up to. `product/README.md`, the package `CLAUDE.md` boundary (both the prose sentence and the allowed-upstream list), `package.json` description, and the rush change file (`minor`, correct for a 6.x package) all agree.

### Nits (non-blocking, no change requested to merge)

- `final.md` §1 still writes the type as `yargsParser.Arguments` while the code imports the binding as `parser` and uses `parser.Arguments`. Doc-only; code is self-consistent.
- `/bind`'s expected refusals (bad/missing/closed Team, DM) surface as `Command /bind failed: …` through the dispatch catch, so an operator-facing refusal wears a "failed" label unlike the in-command Space refusal. This is the deliberate layering outcome — the binding layer owns the sentence — and it is tested; noting it only so it stays a conscious choice.

Nothing else in the diff is out of place: lockfile entries point at the already-present `21.1.1`/`21.0.3`, `FeishuBindTargetSelector` is fully gone (no dangling references, including `index.ts`), `containingChat` is the one-line structural capability from the design, and the parser is only ever fed positionals (`args._`), with the no-throw behavior locked for flag-shaped and backslash input outside the dispatch catch.

Approving.

.agents/tasks/channel/add-bind-and-help-slash-commands/technical-design/final.md
```diff
@@ -0,0 +1,509 @@
...
+```ts
+bind: {
+  usage: '/bind <team_name>',
+  summary: 'Route this group to a Team.',
+  async execute(context) {
+    const [teamName] = context.args;
+    if (teamName === undefined) {
+      return { kind: 'text', text: `Usage: ${COMMANDS.bind.usage}` };
+    }
+    if (context.inSpaceContainer) {
```
ryanxiang7: Nit: this `inSpaceContainer` branch no longer exists — `20331c0` moved the refusal into `FeishuRouting.bind` and deleted the field, and §6 below already removed it from the context table. Please delete the branch from this pseudocode so §4b matches shipped code and §11.

ryanxiang7 (approved):
## Re-review after the new commits — approved

Re-reviewed `36883df`, `e1efdaa`, and `20331c0` on top of the previously approved `cb8e709`, and re-ran the gates independently in a clean worktree at `20331c0`:

- `@excitedjs/feishu-channel` `vitest run` — **552/552 pass (28 files)** (the two new `FeishuRouting` cases included);
- `tsc -p tsconfig.json --noEmit`, `tsc -p tsconfig.tests.json`, `eslint .` — clean;
- downstream `@excitedjs/dreamux` `package-boundary-guards.test.ts` — 24/24;
- `.agents/scripts/check.sh` — clean.

### The Space-container refusal moved into the routing document — correct placement

I checked the three things that placement had to get right, and they hold:

1. **It is genuinely in the commit.** The check is the first precondition inside `FeishuRouting.bind`'s `store.update` mutator, next to `requireOwner`, with the same stated reason. The mutator runs on `structuredClone(current)`; the throw rejects before any atomic write, so `listBindings()` stays empty — pinned by the new test asserting exactly that plus that a fresh topic still plans `provision`.
2. **Every caller meets it.** `routing.bind` has exactly two callers: provisioning and `bindChannel`, and both `/bind` and MCP `bind_channel` converge on the latter. Nothing between the MCP tool and the document reads `spaces`, so the old slash-only refusal was a real hole; the entry-point-free placement is what closes it for present and future callers.
3. **It cannot break the Space mechanism.** Provisioning binds only a `kind: 'topic'` target (`feishu-provisioning.ts`, `origin: 'space'`), and the rule refuses only `group`. The sibling-topic test pins the narrowness: one provisioned topic binds while a fresh sibling topic still plans `provision`.

The two behavior deltas are both honest consequences of the move, and both are now tested rather than merely documented: the refusal wears the `Command /bind failed: …` wrapper (uniform with every other `/bind` refusal — the prefix is true, the command did fail), and a missing Team in a Space chat reports the Team error because `team.status` runs first. The rewritten end-to-end test seeding a live provisioned topic and asserting `team.status` *was* called is good hygiene — a green run there cannot be the wrong refusal wearing the right name.

### Seam cleanup (`e1efdaa`)

`FeishuSlashCommandInvocation.args: readonly string[]` is the right seam. I checked the emitted declarations in a fresh `dist/`: `yargs-parser` appears in no `.d.ts`, and the pinned `FeishuInboundDelivery` export no longer carries the parser's `any` index signature. The narrowing is honest rather than asserted: every positional is a string precisely because `parse-positional-numbers` is off, which the shared `NUMERIC_LOOKING_TEAM_NAMES` fixture keeps locked at both the recognition and end-to-end levels. The dead `FeishuTargetSelectorFields` deletion and `resolutionChain` deriving from `containingChat` remove two genuine duplications; `spaceForContainer` is not orphaned (`plan` still uses it).

### Recorded reverse hole

I independently confirmed the documented asymmetry in code rather than taking the record's word: `bindSpace` inspects only `document.spaces` (duplicate name) and never `document.bindings`, so group-bind-then-register-space still shadows. Recording it as the operator's call is the right call — closing it changes `bind_collaboration_space` behavior — and it is stated identically in `channel.md`, `final.md` §11, and the task README.

### One non-blocking nit

The §4b pseudocode in `final.md` still shows the deleted `if (context.inSpaceContainer) { … }` branch inside the command row, while §6 has already removed the field from the context table and §11 documents the rule's move to `FeishuRouting.bind`. Worth deleting those lines so the final design has no internal contradiction; it does not affect anything shipped.

Approving on `20331c0`.

packages/channel/feishu-channel/tests/feishu-routing.test.ts
```diff
@@ -347,6 +347,140 @@ describe('FeishuRouting — Collaboration Space policy', () => {
...
+      teamName: 'team-a',
+    });
+    // The sibling topic is untouched: one provisioned topic never speaks for
+    // the rest of the Space, which is what binding the chat itself would do.
+    expect(routing.plan(topicTarget('oc_space', 'omt_two'), 'oc_space')).toMatchObject({
+      kind: 'provision',
+    });
+  });
+
+  /**
```
ryanxiang7, line 412: The legacy coexistence case is untested: a document that already carries BOTH the space and a `group` row (predating this rule), and then a **re-registration/rename** of that space, must also refuse and commit nothing. The check currently sits above the `existing === undefined` fork, which is exactly the fail-loud-on-update ruling — worth one test so a future "exempt updates" relaxation has to change an assertion first.

.agents/tasks/channel/add-bind-and-help-slash-commands/technical-design/final.md
```diff
@@ -0,0 +1,511 @@
...
```
ryanxiang7: Still here on the second push: §4b's pseudocode shows `if (context.inSpaceContainer)` even though the field was deleted, §6 dropped it from the context table, and §11 says the rule moved into `FeishuRouting.bind`. Please delete this branch so the final design stops contradicting itself.

.agents/tasks/channel/add-bind-and-help-slash-commands/README.md
```diff
@@ -0,0 +1,162 @@
...
+| `.agents/domains/channel.md` | Slash-command section: the five-row table, `usage`/`summary`, the single `yargs-parser` recognition seam and why `parse-positional-numbers` is off, `/bind`'s behavior, and the correction that a command may now change routing. Card placement: `announceIn`, and why a card sent away from its target carries no anchor Team. Routing tools: the binding operations take a `FeishuTarget`, the selector type is gone, and the MCP wire input still cannot tell a direct message from a group. |
+| `packages/channel/feishu-channel/CLAUDE.md` | Dependency boundary admits `yargs-parser`; slash-command responsibility restated. |
+| `package.json` | `yargs-parser` runtime dependency, `@types/yargs-parser` dev dependency, and the description's dependency claim. |
+| Rush change file | Three, all `@excitedjs/feishu-channel` type `minor` with ordinary notes: the command surface, the Collaboration Space container refusal, and the reverse-order refusal in `bind_collaboration_space`. Each is a separate added file rather than an edit of the last, because `rush change --verify` counts only added files. No persisted file shape changed, so no `BREAKING:` and no `Rebuild:`. |
```
ryanxiang7: Four change files land with this PR, not three: `document-space-binding-exclusion_2026-09-15-22-20.json` is `@excitedjs/dreamux` type `none` — the file commit `20604ba` exists to add. The `dreamux-maintenance` row below should name it.

ryanxiang7 (approved):
## Re-review of `36883df..20604ba` — approved

Scoped to the four new commits as asked (`e1efdaa` and `20331c0` were already covered in my previous review; I re-verified them only where the new commits touch the same lines). Gates re-run independently in a clean worktree at `20604ba`: `vitest run` **554/554 (28 files)**, both `tsc` configs, `eslint .`, `rush change --verify`, downstream `package-boundary-guards` 24/24, and `.agents/scripts/check.sh` all green.

### The group-only narrowing is complete — I went looking for another shadowing path and could not build one

Every writer of the two collections, and every way a target reaches provisioning:

- **The two collections have exactly four mutators, and only two add state.** `bind` pushes/updates `bindings` (guarded, space-side predicate first); `bindSpace` pushes/updates `spaces` (guarded, binding-side predicate first); `unbind`, `forgetTeam`, and `unbindSpace` only ever remove rows, so they can only restore the invariant, never violate it. A rebind-in-place cannot morph a row's kind — the existing row is keyed by `targetKey(input.target)`, only `team_name`/`display`/`origin`/`space_id` are overwritten.
- **Provisioning physically cannot install a `group` row.** `project()` returns `containerChatId: null` for anything that is not a message carrying a real thread id in a chat whose mode resolves to `topic`; `plan` only consults spaces when `containerChatId !== null`, and it returns `provision` carrying that topic target. So `routing.bind({ origin: 'space' })` is reached with a topic target on every path, not by convention but by construction. The new refusal cannot reject a provisioning bind.
- **The MCP side is caught by the same predicate.** A bare wire `chat_id` converts to `group`, `leader_bind_channel` converges on the same `routing.bind` with `requireOwner` — there is no second code path that could carry a group target around the check.
- **The space-side predicate is exact.** It matches `target.kind === 'group' && target.chat_id === containerChatId`; the `topic` rows provisioning installs carry the same `chat_id` and are deliberately excluded — pinned by the "re-registers a space whose topics are already provisioned" test, including a rename and an unchanged plan for the provisioned topic. p2p rows cannot exist (`isBindableTarget`), and at most one space can sit on a container (`deriveSpaceId` includes the container, so repeat registration is the update branch, not a second row).

The merged two-sided rule really is one invariant: the commit ordering argument that justified `requireOwner` inside the commit applies symmetrically, and a single `FeishuRoutingDocument` header is the right home — that header already declares the two sections one consistency domain. Rejecting the shared predicate was also right; a direction parameter would be a mechanism invented to deduplicate two three-line collection scans.

### Historical documents: fail-loud on the update path is the right call — don't exempt it

You asked specifically, so my position rather than a shrug: refusing the update/rename path too is correct, for a reason beyond the repo's general fail-loud posture. A document with the coexistence already on disk is **currently malfunctioning** — fresh topics plan `bound` to the group Team and silently never get their own. A create-only check would let an operator keep mutating policy on top of a broken state and go on not noticing; failing the one write that names this exact entity surfaces the defect at the moment the operator is touching it, and the message names the Team to release. It also keeps the rule genuinely one invariant instead of "both writes refuse, except this branch of this write." The remediation is verified end-to-end by the new test: unbind the group → re-register succeeds → a fresh topic plans `provision`. Legacy documents still load (`validated` is deliberately shape-only) and `plan`'s behavior for them is unchanged, so this cannot block a start or an upgrade.

Naming the owning Team in this refusal while the `requireOwner` refusal next door withholds it is also coherent: `bind_collaboration_space` is registered with `callers: ['dispatcher']`, so route ownership is a read that caller is entitled to; the tools that withhold it serve TeamLeaders.

### Docs vs code — wording is honest at all three sites

- `builtin-feishu.md`: "each names the conflict and writes nothing" — accurate: the bind side names *that it is* a Space, the space side names the holding Team; both throws are pre-write inside the mutator. "Unbind the one you do not want first" matches both remediation orders.
- `product/README.md`: "Both ways of reaching that state are refused" and the reconfiguration sentence are both bounded to the topic-rows case, which is the normal state of a live Space.
- `channel.md` + the `FeishuRoutingDocument` header: the "one consistency domain" sentence it points at exists verbatim in the pre-existing header. One tiny phrasing snag: "Neither write can be reached without passing the other's check" reads as if each write passes the *other method's* check; each actually passes the check over the other *collection*. Cosmetic.

### Nits (non-blocking)

1. **The fail-loud-vs-exempt decision you asked me to adjudicate has no test pinning it.** The first new test covers the create path (group row → first registration refuses); the second covers topic-rows reconfiguration. What is missing is the legacy case: space already registered **and** a group row present → a *re-registration / rename* also refuses and commits nothing. That is precisely the conscious choice in your message, and the current placement (the check is above the `existing === undefined` fork) is exactly what a future "helpful" change would relax. One test there locks the ruling the same way the topic-rename test locks the narrowness.
2. `final.md` §4b's pseudocode still shows the deleted `if (context.inSpaceContainer)` branch — I flagged this on the previous push and it survives in this one, while §6 removed the field and §11 documents the move. The final design currently contradicts itself on the page a reader trusts most.
3. The task record's knowledge-closeout table says "Three, all `@excitedjs/feishu-channel`" but four change files land with this PR — the fourth is `@excitedjs/dreamux` type `none`, which commit `20604ba` exists entirely to add; the `dreamux-maintenance` row should name it.
4. Changelog types are right as written (`minor` for a 6.x package; no `BREAKING:` — no shape/version/startup contract changed, and the now-refused call only ever produced a silently broken Space; the `none` dreamux file matches the doc-only precedent). One optional half-sentence for the `refuse-space-over-bound-chat` note: a pre-existing coexistence document keeps loading and heals by unbinding the chat first — that is the one upgrade-time question the note does not answer today.

Approving on `20604ba`.

PR #429, opened 2026-09-15
Subscribe a recipient to one Feishu document's comments

### LINKED ISSUE #427: Solution review: subscribe a recipient to one Feishu document's comments
## What this is

Solution review for a new Feishu channel capability: a caller subscribes one
document, and the comment events Feishu pushes for that document are delivered
into that caller's own model context.

Task record: `.agents/tasks/channel/subscribe-document-comments/`
(requirement, probe evidence, and the draft technical design).

## The premise correction that shaped it

The original request assumed the channel could ask Feishu for a document's
comments. It cannot. `drive.notice.comment_add_v1` is a *notification* event
keyed to a recipient, not a document feed — the official event page says so
outright ("该事件通知逻辑和文档中评论通知逻辑一致"), and a live nine-case probe
against a real app agrees:

| # | Setup | Comment action | Delivered to the app? |
| --- | --- | --- | --- |
| A | app has no permission on the document | plain comment | no |
| B | app has no permission on the document | comment @-mentions the bot | no |
| C | app is a collaborator (`full_access`) | plain comment | no |
| D | app is a collaborator (`full_access`) | comment @-mentions the bot | **yes** |
| E | app is a collaborator (`full_access`) | anchored (block) comment @-mentions the bot | **yes** |
| F1 | the bot has not replied in that thread | reply in the thread, no mention | no |
| F2 | the bot has replied in that thread | reply in the thread, no mention | **yes** |
| G | `POST /drive/v1/files/:token/subscribe` returned `is_subscribe: true` | plain comment | no |
| J | app is a collaborator with only `view` | comment @-mentions the bot | **yes** |

So an app receives a comment only when the bot itself would be notified: the
comment @-mentions the bot and the app can see the document (`view` is enough),
or the reply lands in a thread the bot already replied in. The file-level
subscribe API does not widen this (row G). The user-identity path
(`POST /drive/v1/user/subscription`) would, but it needs user OAuth and is out
of scope.

`subscribe` is therefore **routing and filtering of events Feishu already
pushes**, not a request for more events. The tool description says this in the
model's own words so it cannot promise a user something the platform will not do.

## The design

One new fact enters the channel: *a recipient follows a document*. Everything
else is an existing mechanism extended to carry it.

- **State.** A third section in the channel's own routing document, keyed by
  `(file_token, team_name)`, with `team_name: null` meaning the Dispatcher
  Agent. An absent section reads as `[]` — no version bump, no `Rebuild:`.
- **Removal.** `forgetTeam` drops a Team's subscriptions in the same commit that
  drops its bindings, so one commit still holds one consistency domain. The
  existing two proofs stay the only two: Core's final closed `team.state`, and a
  delivery rejected pre-admission with `TEAM_CLOSED` / `TEAM_NOT_FOUND`. No
  third authority. Dropped subscriptions are returned separately from removed
  routes, because the route-removal announcement addresses a chat target and a
  subscription has none.
- **Tools.** `subscribe_document` / `unsubscribe_document`, one definition each,
  offered to both callers. Neither takes a recipient argument: it is derived
  from the caller, so a TeamLeader structurally cannot subscribe on another
  Team's behalf. Input is a document URL or a bare token; a wiki URL is resolved
  to the underlying object token first, because the node token is not the
  `file_token` an event carries.
- **A subscribe refuses what can never fire.** It reads the document's metadata
  first; a failed read means this app cannot see the document, and the refusal
  says to add the bot as a collaborator.
- **Delivery.** A fourth typed event route feeds one new module. Each subscriber
  gets one submission whose body is a single ref element in the shape
  `<merged-forward>` and `<reply-to>` already established:
  `<doc-comment file_token=… file_type=… comment_id=… reply_id=… notice_type=…
  mentioned=… note="pull the comment thread with lark-cli" />`, plus a
  doc-comment-specific reminder (the standing chat reminder would be a lie —
  there is no chat to reply into). An event no subscription claims is dropped
  and logged.
- **Deleted in the same change:** `FeishuTransport.fetchDocComment` and its
  helper and result types. They have no caller, the body they fetch is exactly
  what `lark-cli` pulls on demand, and a platform method nobody calls is a
  mechanism nobody is paying for.

## Two open questions

1. **Who may unsubscribe.** The caller-is-the-recipient rule was stated for
   *subscribe*. If it also governs unsubscribe, the Dispatcher can only remove
   its own rows and a Team's subscription cannot be cleaned up short of
   dissolving the Team — whereas `unbind_channel` gives the Dispatcher authority
   over every route. Follow `unbind_channel` instead?
2. **Document title in the body.** The event carries no title. Carrying one
   costs an extra platform call per event; omitting it costs the agent a
   `lark-cli` call to say which document it is reading.

## Not yet authorized

No product code has been written. This issue is the solution review surface;
implementation waits on the development-authorization card.

YourWildDad:
Closes #427.

## What this does

A recipient — a Team, or the Dispatcher Agent — follows one Feishu document, and
the comment events Feishu pushes for it arrive in that recipient's own model
context.

## The premise this was built on, and how it was established

The original request assumed the channel could ask Feishu for a document's
comments. It cannot. `drive.notice.comment_add_v1` is a *notification* event
keyed to a recipient, and a live probe against a real app settled exactly what
that means for a bot:

| # | Setup | Comment action | Delivered? |
| --- | --- | --- | --- |
| A | app has no permission on the document | plain comment | no |
| B | app has no permission on the document | comment @-mentions the bot | no |
| C | app is a collaborator (`full_access`) | plain comment | no |
| D | app is a collaborator (`full_access`) | comment @-mentions the bot | **yes** |
| E | app is a collaborator (`full_access`) | anchored comment @-mentions the bot | **yes** |
| F1 | the bot has not replied in that thread | reply, no mention | no |
| F2 | the bot has replied in that thread | reply, no mention | **yes** |
| G | `POST /drive/v1/files/:token/subscribe` returned `is_subscribe: true` | plain comment | no |
| J | app is a collaborator with only `view` | comment @-mentions the bot | **yes** |

So `subscribe_document` is **routing and filtering of events Feishu already
pushes**, not a request for more of them, and the tool description says so in
the model's own words. The file-level subscribe API does not widen delivery
(row G) and is not called. The user-identity subscription API would, but needs a
user token this channel does not hold — and `subscription_status` answers
`is_subscribe: false` for both identities of this app, so every observed event
was addressed to the bot itself.

## The shape

One new fact enters the Channel: *a recipient follows a document*.

- **State.** A third section in the routing document the Channel already owns,
  keyed by `(file_token, team_name)`, so a recipient can only ever write its own
  row. An absent section reads as `[]` — no version bump, no `Rebuild:`.
- **Removal.** `forgetTeam` drops a Team's subscriptions in the same commit that
  drops its bindings, and reports them *apart* from its routes: a removed route
  is announced back into the conversation it named, and a subscription has none.
  Route removal keeps exactly its two existing proofs and gains no third
  authority.
- **Tools.** `subscribe_document`, `unsubscribe_document`, `list_subscriptions`
  — one definition each, offered to both callers. None takes a recipient
  argument; the caller *is* the recipient. That is why they need no second
  definition the way `bind_channel` does, and it covers the read and the removal
  as well as the write.
- **Delivery.** A fourth typed event route feeds one new module. Subscribers are
  submitted concurrently and independently, and a pre-admission rejection
  removes that one subscriber's row while the others are untouched.
- **The unclaimed event.** Dropped and logged unless it @-mentions the bot, in
  which case it reaches the Dispatcher Agent — but only from a commenter in the
  gate's `allow_users`. The reason a subscribed comment needs no access gate is
  that the subscription is the recipient's own authorization, and that reason
  does not exist when nobody subscribed.
- **The body** carries the comment text and the document text it is anchored to,
  so a cold-open mention is readable without a tool call first; the ids still
  point at lark-cli for the rest of the thread. A failed read delivers the event
  without the text rather than dropping it.

## Corrections this change makes along the way

- `FeishuSubmission` becomes a discriminated union rather than growing a nullable
  anchor, so a chat submission cannot lose its anchor by omission.
- `fetchDocMeta` returned `null` for three unrelated situations — invisible,
  request error, unsupported type. A subscribe refusal quoting that would have
  told an operator to fix permissions that were fine after a timeout. It is
  discriminated now.
- `fetchDocComment` and its helper had no caller and were deleted; the
  comment-text requirement then created one, so the narrower
  `fetchDocCommentText` replaces them.
- The event-route seam's own note asked for promotion to a generic
  `eventType -> handler` map at a third event type, and this is the fourth. The
  note is answered rather than followed: every route carries a different payload
  type and its own normalizer, so one map value type would force each handler to
  re-narrow exactly what the seam exists to have narrowed.
- The `feishu-channel.ts` doc-comment wiring had no test crossing it; an
  end-to-end session test now does.

## Verification

All four gates green from the repository root after the rebase onto `next`:
`build` `SKIPPED: 6`/`SUCCESS: 2`, `lint` `SUCCESS: 7`, `test` `SUCCESS: 4` +
`SUCCESS WITH WARNINGS: 3`, `typecheck:tests` `SUCCESS: 7`. No `FAILURE` in any.
The three warning operations are pre-existing runtime suites writing to stderr
while asserting error paths.

Two platform facts were closed by live probe beyond the delivery matrix: a
thread's own comment is `reply_list.replies[0]` (load-bearing, because an event
for a top-level comment carries an empty `reply_id` while the API gives that
comment its own non-empty one), and a whole-document comment carries no `quote`.

Two things cannot be verified before deployment and are recorded as such: that a
comment event for a **wiki-hosted** document carries the object token rather than
the node token — derived from how every comment API behaves, never seen in a
payload — and that a `--as bot` reply actually arms later delivery end to end.

Requirement, probe evidence, design, operator rulings, and verification are in
`.agents/tasks/channel/subscribe-document-comments/`.


packages/dreamux/skills/dispatcher/dreamux-maintenance/references/builtin-feishu.md
```diff
@@ -9,20 +9,24 @@ There are no credential defaults and no other built-in Feishu config fields.
 
 ## Feishu-Owned Routing State
 
-The built-in Feishu Channel owns which conversation reaches which Team, and its
-collaboration-space provisioning policy. Both live in one server-owned document
-per configured channel at
+The built-in Feishu Channel owns which conversation reaches which Team, its
```
ryanxiang7, line 12: **Blocking — CI's `rush change — declaration check` is red because of this file.** This maintenance reference lives under `packages/dreamux/`, and the PR also changes `packages/dreamux/tests/helpers/fake-feishu-bot.ts` and `packages/dreamux/tests/package-boundary-guards.test.ts`, so `@excitedjs/dreamux` is a changed project with no change description:

```
ERROR: The following projects have been changed and require change descriptions,
- @excitedjs/dreamux
```

Identical situation to PR #428, fixed in commit 20604ba: add a change file for `@excitedjs/dreamux`, type `none` (this package's own runtime behavior is unchanged — a bundled reference doc and test helpers). The 0.x line forbids `major`.

common/changes/@excitedjs/feishu-transport/dreamux-team-space-dreamux-xchyo3_2026-09-15-14-37.json
```diff
@@ -0,0 +1,11 @@
...
...u reports the token in `failed_list`, or `unsupported_type` — and a failed request rejects instead of resolving. It used to return `null` for all three, so a caller using it as a permission proof could report a timeout as missing permission. `resolveWikiNode(token)` is new: it answers the document a wiki node holds, or `null` when this app cannot see the node, and throws on a non-zero Feishu business code rather than reading it as invisibility. `parseFeishuDocumentRef` reads a document share URL or a bare token into `{ token, type }`, mapping Feishu's URL words (`docs`, `sheets`, `base`, `mindnotes`) to the types its APIs take and passing anything else through. `FeishuCommentEvent` now carries the event's `timestamp`. `fetchDocCommentText(request)` replaces `fetchDocComment`: it reads the text of the single comment an event names — the reply `reply_id` names, or the head of the thread when it is empty — plus the document text that comment is anchored to, and answers `null` whenever the thread does not hold it. The old call returned a whole thread of raw Feishu content elements; `commentFromBatchQuery` and the `FeishuDocComment` / `FeishuDocCommentReply` types are removed with it.",
```
ryanxiang7: **This entry removes a 0.x package's exported surface and needs the `BREAKING:` marker per repo rule.** `commentFromBatchQuery`, `FeishuDocComment`, and `FeishuDocCommentReply` are deleted and `fetchDocMeta`'s contract changes (`null` for everything → a discriminated result) — and this transport is the shared SDK boundary claudemux depends on. The rule for 0.x packages: breaking changes ship as type `minor` with the comment leading `BREAKING:`, and a same-shape change needing no rebuild uses the `BREAKING: Review: … No rebuild is needed.` form. Precedent: the `removeReaction` removal in PR #347 used exactly that (change file consumed by release commit 8ca6008). The text itself is complete; it just needs to lead with the marker so the break is visible in the changelog rather than discoverable only at a consumer's compile.

packages/channel/feishu-channel/src/feishu-document-comments.ts
```diff
@@ -0,0 +1,469 @@
...
+
+  /**
+   * What happens to an event nobody follows.
+   *
+   * The two drops carry different reasons on purpose: one says the bot was not
+   * addressed, the other says it was addressed by someone this Dispatcher does
+   * not trust, and only the second is something an operator can act on.
+   */
+  private async deliverUnclaimed(event: FeishuCommentEvent): Promise<void> {
+    if (!event.mentionedBot) {
```
ryanxiang7, line 276: **Design question on the new unclaimed-mention increment: this drop contradicts `DOC_COMMENT_REMINDER` on the path this same method opens.**

When a trusted @-mention on an unsubscribed document is delivered to the Dispatcher below, the submission's reminder tells the model to answer in the thread with lark-cli `--as bot` precisely because “Feishu keeps delivering later replies in a thread only once this bot has replied in it”. Those later replies arrive in the exact shape dropped here: no subscription row, and Feishu requires **no** @-mention for a reply inside a thread the bot has already replied in. So on the happy path the reminder steers toward — Dispatcher answers in the thread, human replies again — the follow-up is silently dropped with `no_subscriber`. The KB's claim that this shape “can only be the tail of a removed subscription” therefore has a second reachable source: the tail of a cold-open thread the Dispatcher just replied in.

Suggested minimal fixes, pick one consciously:

1. Give the cold-open delivery its own reminder sentence: the thread keeps reaching the Dispatcher only if it subscribes the document first (it can — `subscribe_document`, recipient derived as null).
2. Accept the cold open as one turn and say that in the reminder and the KB, instead of promising continuation.

I would **not** route every non-mention unclaimed event to the Dispatcher without added state — it would resurrect tails of genuinely removed Team subscriptions, which the current split exists to prevent.

packages/channel/feishu-channel/src/feishu-document-comments.ts
```diff
@@ -0,0 +1,469 @@
...
+   * A wiki URL is resolved here too, because the string a recipient subscribed
+   * with is the string it will unsubscribe with, and the row holds the object
+   * token. No metadata read: removing a row proves nothing and needs nothing
+   * proven.
+   */
+  async unsubscribe(input: {
+    document: string;
+    teamName: string | null;
+  }): Promise<{ file_token: string; unsubscribed: boolean }> {
+    const resolved = await this.resolveDocument(input.document, null);
```
ryanxiang7, line 188: Minor — please confirm this trade-off is deliberate. Unsubscribing a wiki URL still resolves the node, so a wiki page whose node the bot can no longer see cannot be unsubscribed (the `null`/business-code failure above), despite the doc comment that removing a row “proves nothing and needs nothing proven”. It is structurally unavoidable while the row stores only the object token (the node token is not persisted), and the refusal does give a workable remediation — re-grant view, then unsubscribe — so I would not add schema for it. One sentence here or in the KB noting the limitation would stop a future reader reading the refusal as a bug.

ryanxiang7 (changes requested):
## Re-review of the two new increments

Read the full diff at `6b897e6` against `origin/next`, including the transport implementation against the installed `@larksuiteoapi/node-sdk@1.73.0` types, the routing document/store changes, the tools, the discriminated-union narrowing, the tests, the maintenance reference, and the `.agents/` delta.

### What lands well

- **Unclaimed @-mention → Dispatcher.** The trust authority is right: `isTrustedDispatcherUser` is a read-only view of the gate's own `allow_users`, injected as a port, so this module learns no access-file layout and the gate keeps one owner. No chat window exists for a pairing card, so dropping an untrusted commenter is the only honest answer; missing `access.json` reads as the secure default. Subscribed events deliberately run no gate, the fan-out is concurrent with `Promise.all`, enrichment happens only after the drop checks (a dropped event spends no platform call), and the proof acts per subscriber — a rejected Team loses exactly its own row, the Dispatcher's submission is untouched, and nothing is re-routed.
- **Comment text + anchor quote.** The half-reverted fetch is narrower in all the right ways: one named item instead of a raw thread, `reply_id === ''` selecting `reply_list.replies[0]`, a `null` for “no text to show” kept distinct from a thrown business code/request failure, one shared 2s enrichment deadline with the sender-name lookup, and ids-only delivery when the read fails. The body's escaping reuses the chat body's exact primitives, and `sourceId = fileToken:commentId:replyId` correctly admits every reply in a thread.
- **Everything from the first-round review that was pinned is in the code:** the discriminated `fetchDocMeta` that throws on a no-verdict response, wiki-first resolution with throw (not “invisible”) on a non-zero business code, the optional `subscriptions` section materialized to `[]` with present-but-malformed failing like the other two, `FeishuChatSubmission` narrowing so an anchorless submission cannot open a COT card, the typed fourth route with the map-promotion note answered, and the caller-scoped tool catalog that structurally cannot name another recipient. Test coverage matches the claims, including the end-to-end session test asserting no card is opened, and the boundary-guard export inventory is updated.

### Why this is “request changes”

1. **Blocking: the `rush change — declaration check` CI job is red** — no change file for changed project `@excitedjs/dreamux` (this bundled maintenance reference + two test files). Same fix as commit 20604ba in #428: a type-`none` entry. The other seven checks, including both rush build/test/lint gates, are green.
2. **The `@excitedjs/feishu-transport` change entry needs the `BREAKING: Review: …` marker** — it deletes exported surface from a 0.x shared package; repo rule and PR #347 precedent inline.
3. **One design contradiction to resolve consciously** (inline at `deliverUnclaimed`): after a cold-open Dispatcher reply posted `--as bot`, the human's non-mention follow-up in that thread is the exact event this branch drops, so the standing reminder promises a continuation the delivery path discards.
4. One nit on the wiki-unsubscribe-when-invisible edge, inline.

Once 1–2 are in and 3 has a decision, this is approvable from my side.

ryanxiang7 (approved):
## Re-review at `5f8ebef` — approved

All four items from the previous review are resolved, and the two new increments are clean.

**Previous findings**

1. **Blocking change file** — fixed: `@excitedjs/dreamux` now has a type-`none` entry; the declaration check is green and all nine CI jobs pass on this commit.
2. **Breaking-change marker** — fixed: the transport entry leads with `BREAKING: Review:` and closes with `No rebuild is needed.`, type stays `minor` on the 0.x line, matching the #347 precedent. The replacement-file detail (rush verify only recognizes Added) is handled correctly.
3. **The cold-open contradiction** — resolved at a better level than my suggestion. The ownership probe (bot-owned doc → every comment pushed regardless of mention; `full_access` collaborator on a user-owned doc → nothing) falsifies the closed enumeration rather than just extending it, and the code now states a rule with examples in all four places instead of enumerating a set. The operator's ruling to keep dropping unclaimed non-mention events is carried through consistently — code comment, `channel.md` (which now explicitly names all three sources that reach that drop, including the post-cold-open human reply and bot-owned-document comments), `final.md`, and `verification.md` record the accepted cost. Reminders now state what happened, with the cold-open variant adding only the fact the envelope cannot carry; the promised continuation that did not exist is gone.
4. **Wiki unsubscribe edge** — documented in both `unsubscribe()` and the maintenance reference, with the working remediation (`file_token` from `list_subscriptions`) and the reason not to persist a node token.

**New increments**

- **Envelope flattening.** Correct shape: the locating facts are attrs on the same `<channel>` start tag Core renders for chat, next to the shared `sender_id` / `sender_name` / `create_time`; every name passes Core's safe-name rule and Core owns value escaping. Empty `reply_id` is dropped exactly as the chat path drops an empty attr. The body is now only the blocks a chat body already uses, and always writing `<content />` is the right call for both reasons given — one shape, and `team.submit`'s non-empty-text validation.
- **Mention normalization.** The layering is right in both directions: the transport answers ordered segments and stays out of the agent-facing XML business, and the channel renders a mention through one function shared with the chat path, so the two cannot drift. `docs_link`-as-text, empty payloads contributing nothing, and the exhaustive switch over the SDK's element kinds all check out; the empty-label mention is deliberately the same output the chat renderer already produces for a nameless record.

Tests pin the new contracts (attr inventory and order, dropped `reply_id`, segment rendering, the two reminders, ids-only `<content />` on a failed read), the export guard includes `FeishuCommentSegment`, and both full rush gates are green on ubuntu and macOS.

ryanxiang7 (approved):
Reviewed only `e99b3476` (the two earlier commits were reviewed line by line in prior rounds). CI is 9/9 green at this commit.

**Focus point 1 — change-file wording.** Compliant for the 0.x line: type `minor`, note leads `BREAKING: Review:`, closes with `No rebuild is needed.` The renamed file (17-24 → 02-10) folds in the newly exported-surface deletion (`FeishuDocMeta`) and states why the payload shrank: the only question a caller asks is visibility, `failed_list` already answers the invisible case, and the title/URL had no reader. I grepped the tree: no consumer of `FeishuDocMeta` or of a visible payload remains (`fetchDocMeta` callers branch on the discriminator only), the boundary-guard pin is gone, and `with_url: true` is no longer requested. Nothing in the note claims more than the code does.

**Focus point 2 — the rewritten comments.** Both checked against the code and neither overclaims:

- The `deliverTo` reaping comment: `status: 'rejected'` is produced only for `TEAM_NOT_FOUND`/`TEAM_CLOSED` thrown before admission, so it is the same proof the chat path acts on; `routing.unsubscribe` and `routing.forgetTeam` commit through the same `store.update` queue; the width difference is real (`forgetTeamRoutes` drops every row for the Team and announces into each removed route's chat; the document path removes exactly the one `(file_token, team_name)` row and announces nowhere), and both named costs are real — other subscriptions are reaped one delivered comment at a time, and with no final-close event a document nobody comments on keeps its row. Nothing in the comment implies the document event re-routes or reaps siblings.
- The `isTrustedDispatcherUser` comment: the state is in fact written tmpfile→rename (`writeAtomic`), so the whole-version-or-whole-version race statement holds; and `allow_users` is tested inline at multiple existing sites against state those callers already loaded (`feishu-gate.ts` x4, the inbound path, `introduce`), so the old "second reader = second owner" framing was indeed false and the new rationale (a caller outside the gate would gain an access state for no other reason) is the truthful one.

**Rest of the diff.** `subscribed: boolean` → `reminder: string` removes a hidden two-value callee domain — each caller already owns the exact constant — and the end-to-end test still asserts both reminder strings, so coverage is behavior-level and intact. Moving `FeishuDocumentSubscriptionView` and its projection beside `FeishuBindingView`/`listBindings` puts the row→view projection in the module that owns the rows; `feishu-document-comments` loses a pure delegation method rather than gaining one. The builder becoming module-private is correct: nothing outside the module calls it.

Net direction is entropy-down: fewer requested fields, one fewer exported surface, narrower module seams, and comments that say exactly what the code does. Approving.

PR #431, opened 2026-09-16
State where a document comment is anchored

YourWildDad:
Follow-up to #429, opened separately at the operator's instruction.

## The defect

The delivered envelope described `<quote>` as *the document text this comment is
anchored to*. Reading real comments back shows it is not: Feishu derives that
string from the anchored content and cuts it at a length of its own, with no
truncation marker and no statement of the original length. A recipient reasoning
over it was reasoning over a prefix it had been told was the whole thing.

An empty quote was worse. It meant three different things at once — a comment on
the whole document, a comment Feishu returned no quote for, and a comment we
could not read at all — with nothing to tell them apart.

## What Feishu actually gives

Four rounds of live reads, including one whole-document comment made in the
client, established what the payload holds. The old read used one field of it:

| field | what it says | when it arrives |
| --- | --- | --- |
| `is_whole` | whether the comment is on the whole document | always |
| `quote` | a shortened preview of the anchored content | only when not whole |
| `extra.content_anchor_id` | Feishu's id for the anchored content | always, unasked |
| `relation.content_deleted` | whether that content still exists | only with `need_relation` |

A selection spanning several blocks records exactly one anchor; the rest appear
nowhere in the item, and no read recovers them.

## The change

`FeishuDocCommentText.quote` becomes `anchor`: either `{ kind:
'whole_document' }` or `{ kind: 'content', anchorId, preview, deleted }`. The
read passes `need_relation` so the last fact is answered — same call, same
response, no extra round trip.

The envelope then states all three:

- **`anchor`** — `whole_document` or `content`, absent *exactly* when the
  comment could not be read. That absence is the distinction the change exists
  to make.
- **`anchor_id`** — Feishu's id for the anchored content. For a docx it is the
  block id, which lark-cli reads back in full, so the recipient decides whether
  it wants the passage. The channel fetches none of it: one anchored part can be
  far larger than every comment on it.
- **`anchor_deleted`** — written only when Feishu says the content is gone. A
  response that stays silent is not a denial.

`<quote>` keeps its shape and loses its claim: the note now says it is a preview
Feishu shortens, and points at `anchor_id` for the content itself.

Reading one comment moves out of `transport/feishu.ts` into
`transport/doc-comment.ts`. The anchor pushed that file past the 700-line source
limit, and the comment reader was the one whole concept inside it that nothing
else there touches.

## Not in scope

- Fetching the anchored content. That is `lark-cli`'s, and the size is the
  reason.
- Recovering the other blocks of a multi-block selection. Feishu keeps one.
- The chat path's character-vs-byte body cap — real, unrelated, its own task.
- `DOC_COMMENT_REMINDER` is untouched: the `<quote>` note is written exactly
  when there is a preview to warn about, while the reminder is sent for every
  comment including those with none.

## Gates

`build`, `lint`, `test`, `typecheck:tests`, and `.agents/scripts/check.sh` all
green locally. `lint` failed once first, on `max-lines` — that is what the file
split answers. `typecheck:tests` is the gate that proves no test still builds a
`FeishuDocCommentText` with the removed field.

Task record: `.agents/tasks/channel/state-the-comment-anchor/`.


common/changes/@excitedjs/feishu-transport/dreamux-team-space-dreamux-xchyo3_2026-09-16-06-40.json
```diff
@@ -0,0 +1,11 @@
+{
+  "changes": [
+    {
+      "comment": "A document comment now answers where it is anchored instead of a single quote string. `FeishuDocCommentText.quote` is replaced by `anchor`, which is either `{ kind: 'whole_document' }` or `{ kind: 'content', anchorId, preview, deleted }`: Feishu's own `is_whole` decides which, `anchorId` is the id Feishu gives the anchored content (for a docx, its block id), `preview` is the text Feishu shortens from it, and `deleted` is what Feishu says about the content still existing. The read asks for `need_relation` so that last fact is answered, in the same call and the same response as before. The old `quote` field was described as the document text the comment is anchored to and is not: Feishu derives it and cuts it at a length of its own, without saying that it cut anything or how long the original was. Reading one comment moves out of `transport/feishu.ts` into its own `transport/doc-comment.ts`, which owns the comment types, the segment normalization, and the anchor; `FeishuCommentAnchor` joins the package's exported types and every other export keeps its name and path.",
```
ryanxiang7, line 4: This removes and replaces a field on an **exported** transport type (`FeishuDocCommentText.quote` → `anchor`), so it is a compile-time breaking change for every consumer of `@excitedjs/feishu-transport` — `feishu-channel` itself moves in this same PR. The house precedent is #429, one commit back: for replacing the transport's document-comment surface its change file leads with `BREAKING: Review:` and states that no rebuild is needed. Two consecutive changes to this same exported surface should be marked the same way. Suggest leading this comment with `BREAKING: Review:` (still type `minor` on the 0.x line) and stating explicitly that no rebuild and no state migration are needed — only the exported type shape changed.

packages/channel/feishu-transport/src/transport/doc-comment.ts
```diff
@@ -0,0 +1,196 @@
...
+ * The comment fields Feishu returns and the SDK's response type omits. `extra`
+ * arrives on every item; `relation` only when the read asks for it, and only
+ * for a document type that has one.
+ */
+interface UndeclaredCommentFields {
+  extra?: { content_anchor_id?: string }
+  relation?: { content_deleted?: boolean }
+}
+
+function undeclaredCommentFields(item: unknown): UndeclaredCommentFields {
```
ryanxiang7: Nit on the narrowing you asked about: substantively this is fine — every read goes through optional chaining and `content_deleted` is taken only on `=== true`, so a payload that differs from the type degrades to empty instead of throwing, and a structural cast of an authenticated API response matches `transport/feishu.ts` (`feishuChatClient`, `contactUserApi`); runtime guards stay at the parse boundary, which faces untrusted input. The one awkwardness is that `undeclaredCommentFields` is an identity cast but reads like an extraction — opening it, a reader finds a zero-op that hands the whole item back. Inlining `item as unknown as UndeclaredCommentFields` at the read site (keeping the interface and its doc comment as the reason), or renaming this to `asItemWithUndeclaredFields` so it reads as a view onto the item, would state what it does. Not blocking.

packages/channel/feishu-channel/src/feishu-document-comments.ts
```diff
@@ -339,6 +341,49 @@ export class FeishuDocumentComments {
...
+        );
+        return;
+      case 'duplicate':
+      case 'stopped':
+        this.opts.log.info(
+          { ...scope, status: outcome.status },
+          'feishu document comment was not admitted',
+        );
+        return;
+      default:
```
ryanxiang7: The new reporting fixes the lie for the outcomes this commit names, but the `default` arm re-asserts a smaller one for `ambiguous`, and this package already owns the honest classification elsewhere.

**1. `ambiguous` is not "failed to deliver".** The union in `feishu-submit.ts` defines it as "proves nothing about whether a turn exists" — a turn may already be answering the comment. This branch logs it at error as a delivery that did not happen, which asserts the negative the type deliberately refuses. The chat path's own `reportDelivery` (`feishu-session-inbound.ts`, ~line 530) already has the template: its own error line, `feishu inbound admission was ambiguous; not replaying` — error level (attention is earned) without claiming failure. This method deserves the same line for doc comments.

**2. `rejected` on the cold-open path lands here too.** The subscribed path intercepts it above, but `deliverUnclaimed` calls `reportDelivery` unconditionally, and `submit(null, ...)` is typed able to return `rejected`. The sibling maps that to **warn** — a proven pre-admission refusal is not a delivery fault — not error. Worth matching, plus it keeps the structured `code`.

**3. This is now the second switch over the same union in one package, and they disagree.** Both `reportDelivery` functions map `FeishuSubmitOutcome` → level/reason, and `outcomeFailure` is a third re-derivation of the `error?.message ?? 'unknown'` / `message` reading the sibling already does inline. When the union grows a status, two switches have to be found and they can silently classify it differently — which is exactly the gap here. Consider one classifier next to the type it classifies, in `feishu-submit.ts` — e.g. `describeSubmitOutcome(outcome) -> { level: 'info' | 'warn' | 'error', fields, kind }` — with each surface keeping only its own prose and scope. The mapping is the shared fact; "inbound" vs "document comment" wording is the only thing that should differ.

Noting what I checked and is fine: `duplicate`/`stopped` as info is right in both; the fence-closed case genuinely is `error` status and is now honest; the four tests drive the real switch through a stubbed submitter (the logger rewrite weakens nothing); the null-team cold-open case is covered.

ryanxiang7 (approved):
Approved after three rounds.

**Anchor change (`doc-comment.ts` + envelope)**
- The four response shapes measured on the endpoint this code actually calls (`batch_query` with `need_relation`) map to `commentAnchor()` without gaps: anchored → content id/preview/deleted; whole document → `whole_document`; deleted block → `deleted: true` with the id kept; short passage → full preview. The unobserved `is_whole`-absent shape falls honestly into `content` instead of guessing `whole`, and stays distinct from an unread comment.
- `anchor` absence has exactly one source: the comment could not be read (`null`). The attribute is required on every successful read, the channel's empty-value drop is the only filter, and Core's attr path is open names with no allowlist.
- Reading the SDK's undeclared `extra` / `relation` via a structural cast matches this package's trust boundary (authenticated API response; runtime guards stay at the parse edge), and the inline cast now reads honestly.

**Delivery logging**
- The final shape is the right root fix: one classifier (`describeSubmitOutcome`) beside the union it classifies, exhaustive over all seven statuses, each delivery path reduced to a wording table. The chat path's five messages are byte-identical; `rejected` is warn and `ambiguous` is a non-presupposing error on both paths; the one new field (`status` on chat failure lines) is additive and noted.
- The cold-open `rejected` entry is unreachable today (no `team_name` means Core never resolves a Team), but keeping the mapping exhaustive is the type's requirement and now costs a single table row.
- Five outcome kinds are pinned by tests driving the real class; change files follow the changelog policy (no `BREAKING:` for an API-contract change — verified against `.agents/domains/state-config-and-files.md`).

All nine CI gates green on `a06128c`.

PR #433, opened 2026-09-16
Open a topic binding's /teams link on its topic

YourWildDad:
## Summary

The `/teams` card linked every binding with `client/chat/open?openChatId=…`, so a Team bound to a topic opened the group the topic lives in rather than the topic.

`feishu-running-teams-card.ts` now builds each entry's link from the binding:

- **topic binding** → `https://applink.feishu.cn/client/thread/open?open_chat_id=<chat>&open_thread_id=<thread>&openchatid=<chat>&openthreadid=<thread>&thread_position=-1`. This copies, parameter for parameter, the `message_app_link` Feishu returns for a message in a topic.
- **group binding** → unchanged: `https://applink.feishu.cn/client/chat/open?openChatId=<chat>`.
- Both query strings are encoded by `URLSearchParams`; for real chat ids the group link is byte-identical to before.
- A Team with no bindings still renders "No Feishu bindings"; the grey `· <thread_id>` suffix on a topic entry is unchanged.

Both ids come from the Channel's own routing rows (`FeishuBindingView.chat_id` / `thread_id`); nothing new is stored.

## Tests

- New `/teams` dispatch test: one Team bound to a group and to a topic in the same chat; asserts the exact markdown line of each, so the group link is pinned unchanged and the topic link is pinned to the full thread form. Run against the previous source, the test fails on the topic line only.
- On 713d6670: `rush build`, `rush lint`, `rush typecheck:tests`, `rush test` all SUCCESS (three packages "with warnings", which is test stderr output).
- On 41461a98 (encoder unification only): `@excitedjs/feishu-channel` tsc, eslint, `typecheck:tests`, and its full vitest suite (612 tests) pass.
- `rush change --verify --target-branch origin/next` and `.agents/scripts/check.sh`: OK.

## Knowledge delta

- `.agents/product/README.md` and `.agents/domains/channel.md`: a topic binding's link opens that topic, a group binding's its chat.
- Change file: `@excitedjs/feishu-channel`, patch.

## Caveat

The link has not been clicked in a Feishu client from a rendered card. The shape rests on the `message_app_link` Feishu returned for a topic message; whether the client honors it from a card markdown link, including the `&`-separated query (the old group link had a single parameter), is what a `/teams` click after merge verifies.


ryanxiang7 (approved):
Reviewed the change end to end:

- `bindingLink` is a clean, pure presentation-layer helper in the card module; it uses the Channel's own routing facts (`target_kind`/`thread_id`) and adds no new state, no glue, no layer violation.
- The topic-with-null-`thread_id` row safely falls back to the chat link, so a malformed binding can never produce a broken thread URL.
- `URLSearchParams` percent-encodes ids, which is actually safer inside the markdown link than the raw string interpolation on the group branch; verified the rendered string matches the pinned form.
- The new test exercises the real dispatch → card-build path and pins both exact lines, so the group link is locked unchanged and the topic link is locked to the full thread form.
- KB delta and the patch change file are present and appropriate; the behavior is link-shape only with no persisted-state impact.

The only residual risk is unclicked-in-client validation, which the description discloses honestly; the shape mirrors Feishu's own `message_app_link` and the fallback (and revert) is trivial. Approving.

PR #436, opened 2026-09-16
test(feishu-channel): make resource deadline tests clock-deterministic

YourWildDad:
## What

Make the two "hanging resource stream" budget tests in
`packages/channel/feishu-channel/tests/feishu-message-budget.test.ts`
clock-deterministic.

## Why

`bounds a hanging resource stream by the per-message deadline` and
`uses the operation deadline for a hanging resource stream` ran against
20–25 ms **real timers** but asserted `stream.destroyed === true`
synchronously after the result settled.

The resource is only destroyed once it has been acquired:

- via `onLateValue` when the fetcher resolves after the bounded operation
  already timed out, or
- in the `readStreamWithLimit` abort / catch path.

On a congested CI runner, the cold-cache `mkdir`/`stat`
(`ensureOwnerOnlyDir`) can consume the whole 20–25 ms budget before the
fetcher runs. `resolveAttachment` then returns the deadline/timeout result
without ever acquiring the stream, so the stream cannot be destroyed yet
and the synchronous assertion flakes (`expected false to be true`). The
`reason` assertion still passes; only the `destroyed` assertion fails. This
was observed repeatedly on CI under load while passing consistently on
fast developer machines.

The sibling test
`destroys an acquired stream when its operation deadline is already expired`
already models this deterministically with fake timers — this PR applies
the same approach to the two flaky tests.

## How

- Use `vi.useFakeTimers()` + `vi.setSystemTime()` and wait until the
  fetcher has actually returned the hanging stream (via the existing
  `deferred` helper) before `vi.advanceTimersByTimeAsync(...)`.
- This guarantees the deadline/timeout fires against an already acquired
  hanging stream, which is the behavior under test.
- Test names and both assertions (`reason` + `stream.destroyed`) are
  unchanged. Test-only change; no production code touched.

## Verification

- `vitest run tests/feishu-message-budget.test.ts` → 27/27 pass,
  deterministically across repeated runs.
- No production code changed.


ryanxiang7 (approved):
Test-only flake fix; verified the change against the production path:

- The deadline fires from a faked `setTimeout` (`feishu-inbound-work.ts`), and the clock is advanced only after the fetcher has entered; `advanceTimersByTimeAsync` drains microtasks, so the stream is acquired before the 25ms/20ms timer fires — the synchronous `stream.destroyed` assertion is then valid (abort path in `resolveAttachment` destroys the acquired stream before the result settles).
- File-level `afterEach` already restores real timers, so no timer leakage.
- Test names and assertions unchanged; matches the existing deterministic sibling tests.
- Rush change file is `type: none`, appropriate for a test-only change.
- All CI checks green on ubuntu and macos.

PR #438, opened 2026-09-16
Report no Claude Code subagent activity

YourWildDad:
## Why

A Claude Code agent that delegates to a native subagent put the subagent's tool calls, their results, and — for a background subagent — its words on the agent's live conversation card, as if the agent had produced them. The expected display is the main agent's activity only, and the subagent's activity should not be reported at the provider layer at all.

## What changed

- `packages/agent-runtime/claude-code/src/runtime-activity.ts`: `emitStreamActivity` drops every `assistant` / `user` envelope whose `parent_tool_use_id` is non-null, before reading any block. A subagent's text, `tool_use`, and `tool_result` are not reported and never enter the tool correlation map. The main agent's own `Agent` call and its result carry `parent_tool_use_id: null` and still display.
- One test fires the probed wire shape (main `Agent` call → subagent prompt, tool call, tool result, text → main `Agent` result) and asserts only the two main rows.
- Knowledge: the display-line section of `provider-runtime.md`, a new "Observing agents" entry in the product catalog, and row 4 of the stream-json research record (previously deferred, now fixed).
- Patch change file for `@excitedjs/agent-runtime-claude-code`.

Not changed: `TurnAggregator`'s fallback `lastAssistantText` (it still reads a background subagent's text; recorded in the research document), the recent-activity reader, Codex, Core, Channel, contracts, config, and persisted state.

## Evidence

Claude Code 2.1.272 probes, foreground and background subagent: every subagent `assistant` / `user` envelope carried the spawning `Agent` call's id in `parent_tool_use_id`; every main-agent envelope carried `null`. A background subagent's text reached stdout without `--forward-subagent-text`, which the documentation says it needs.

## Validation

- `rush build`, `rush lint`, `rush test`, `rush typecheck:tests`: pass.
- `.agents/scripts/check.sh`, `git diff --check`, `rush change --verify`: pass.
- Mutation check: removing the filter line fails the new test; restoring it passes all 31 tests in the file.
- Independent review of the branch: approved, no findings.

Limits: the probe ran the raw CLI rather than a Dreamux-launched resident session, and the Feishu card has not been observed with the change.

Task record: `.agents/tasks/channel/hide-claude-subagent-activity/`.


ryanxiang7 (approved):
Independent implementation review passed (baseline `next` @ `bc7d769d`).

**Implementation**
- The filter is a single envelope-level check in `emitStreamActivity` (`runtime-activity.ts`): after the `compact_boundary` branch and before any block is read, every `assistant`/`user` envelope with a non-null `parent_tool_use_id` is dropped. A subagent's text, tool calls, and tool results never reach the sink and never enter the correlation `tools` map; the main agent's own `Agent` call and its result are untouched (probe confirmed main envelopes carry explicit `null`).
- Placement is correct: envelope meaning already lives in `emitStreamActivity` (it owns hiding all `user` text), so no glue leaked into the RPC or parser. The RPC still feeds every line to `TurnAggregator`, keeping the documented `lastAssistantText` non-goal behavior unchanged. Compaction, interruption, usage, and `turn.ended` never cross the check.
- `!= null` matches the existing main-context read in `stream.ts` and tolerates builds that omit the field.

**Independently verified (not taken from the task docs)**
- `rush build`, full `rush test` (4 success + 3 with the expected failure-path warnings), `eslint`, and `tsc -p tsconfig.tests.json` all pass locally.
- Mutation check: with the filter line removed the new test fails (30 passed / 1 failed); restored, all 31 pass.
- The 3 post-review commits (`b6d376e`, `ae55aab`, `c266331`) touch only task-record files (state `review` → `done`, PR link, review record); the code/KB/change-file diff is byte-identical to the reviewed `e50dfe7`. `.agents/scripts/check.sh` passes on the PR head.
- Change file is `patch` — correct, since no config/state/path contract changed.

No findings. Approving; merge still awaits the operator's authorization and the pending rush CI gates.

PR #439, opened 2026-09-17
Remove the repo-root bin/dreamux shim

YourWildDad:
## What

Delete the repo-root `bin/dreamux`. It was a source-checkout convenience shim that only forwarded to `packages/dreamux/bin/dreamux`, was never published, and was kept by the npm-package-split CLI decision purely for convenience. The operator ruled to remove it; `packages/dreamux/bin/dreamux` is now the single launcher, and a source checkout runs it directly.

## Changes

- `bin/dreamux`: deleted.
- `.github/workflows/ci.yml`: drop the shim from the shellcheck file list.
- `packages/dreamux/tests/bin-launcher.test.ts`: drop the shim's cases, including its forwarding assertion. With one launcher left, the per-launcher case table and loop collapse into a single `describe`; the package launcher keeps all six of its cases (15 tests -> 9).
- `README.md`: drop the paragraph describing the shim.
- `packages/dreamux/README.md`: the "Run the server" example ran from the repo root through the shim, so it now reads `./packages/dreamux/bin/dreamux serve`; the test-list entry no longer mentions the shim.
- `.agents/domains/repository-operations-and-release.md`: drop the shim from the shellcheck row, the published-surface paragraph, and its Source list.
- `.agents/tasks/architecture/npm-package-split/requirement.md`: the backfilled CLI decision record gains a dated "Since this was recorded" note with the ruling; the preserved body is unchanged.
- Rush change file for `@excitedjs/dreamux`, type `none`: docs and tests only, no runtime change.

## Verification

- `rush build`, `rush lint`, `rush typecheck:tests`, `rush test`: all SUCCESS locally (`@excitedjs/dreamux`: 74 files, 1083 tests passed).
- `.agents/scripts/check.sh`: OK.
- `rush change --verify --target-branch origin/next`: finds the change file.
- shellcheck was not run locally (not installed); CI runs it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01T9u7QAifAmzj7pi3bbNsb1

PR #440, opened 2026-09-17
Reject commit hashes in task records

YourWildDad:
## Why

Task records cite work by commit hash, and almost none of those hashes stay resolvable. The repository merges by squash or rebase, so a branch commit never reaches `next`; the merge commit does not exist while the record merging with it is written; and a commit that was never pushed resolves in one clone only — the same wall #413 met with states.

On `next` after #439 merged, `.agents/tasks/` held 188 commit citations on 167 lines in 77 files. A clone of branches and tags resolved about half of the distinct commits cited repository-wide; two resolved nowhere.

Upstream `deepseek-ai/deepseek-harness` added a gate for exactly this after the snapshot #413 borrowed from. This ports it.

## What changes

- **Rule** (`init_task.py check-all`): every 7–40 digit hexadecimal word in any file under `.agents/tasks/` is resolved in one `git cat-file --batch-check` call with `GIT_NO_LAZY_FETCH=1`. A line fails only when a token names a commit object, so digests, blob hashes, and unrelated hex pass. `## Historical …` sections stay exempt; fenced blocks do not, because a script pinned to a commit is still a citation. `--allow-in-flight` skips it, as it skips the label rule. Both rules now read one `classified_lines` pass.
- **CI** (`kb` job): `fetch-depth: 0` plus `git fetch origin '+refs/pull/*/head:refs/pull/*/head'`, because an object the checkout lacks cannot match.
- **Docs**: `task-records.md` states the rule and which exemptions each rule has; the `kb` row of the CI gate table and the `check.sh` header describe what the job now fetches and checks.
- **Rewrite**: every existing citation now names the pull request, review, or Actions run that keeps it. Trunk baselines read "`next` after #N merged"; #413's counting script resolves its baseline from the pull request at run time and still prints `7 23 23`.

Task record: [`architecture/gate-commit-references-in-task-records`](.agents/tasks/architecture/gate-commit-references-in-task-records/README.md), with the design, the operator's rulings verbatim, and the verification.

## Verification

- `.agents/scripts/check.sh` passes in a shallow clone and with a full clone's objects added.
- 16 behavior probes pass; each of 7 mutations of the gate (skip fences, prefix-match the frozen heading, accept any object type, ignore `--allow-in-flight`, README only, drop the word boundary, unwire from `check-all`) fails at least one probe.
- Before/after measured with the rule itself: 167 lines / 188 citations / 77 files → 0.
- The rewritten lines carry 25 review and Actions run identifiers; each one's recorded head is a commit the same file's removed lines cited.
- Rush build, lint, test, and typecheck:tests pass. No package source changed, so no change file.

The rewrite ran as #413's did: a Sonnet node per file rewrote, a Sonnet node per file re-checked (9 of 77 files corrected), and an Opus node reviewed the whole diff (11 more corrections).

## Left as written, on purpose

- Six npm prerelease versions of the form `…-alpha.g<prefix>` in four records. The word boundary, copied from upstream, does not treat them as candidates, and each is the version npm published the package under.
- Review-input SHA-256 digests in `channel/simplify-feishu-replies` describe what reviewers read, not the current tree; 41 of 57 already differed from `next` before this change.

## After merge

Open pull requests whose own task files cite commits — #377 (1 line) and #361 (5 lines) — will fail the `kb` job until those lines are rewritten.

https://claude.ai/code/session_01T9u7QAifAmzj7pi3bbNsb1


ryanxiang7 (approved):
Reviewed the gate, CI wiring, docs, and the full 77-file citation rewrite.

- Gate logic verified with behavior probes: full/short commit hashes rejected; tree/blob/64-hex digest/non-existent hex/pinned prerelease suffix pass; fenced blocks are read for this rule while `## Historical …` stays exempt; `--allow-in-flight` skips it as documented.
- All removed commit citations were cross-checked: PR-number replacements match GitHub's authoritative commit→PR association, and every replacement review/Actions-run link records the exact head SHA that was removed (spot-verified via the API), including the two hashes that resolve in no clone.
- The two line classifiers are now one `classified_lines` pass — a net simplification, not new glue. `check.sh` passes locally; all CI jobs green.

PR #442, opened 2026-09-17
Emit a neutral token.usage activity instead of a synthetic usage message

YourWildDad:
## Why

Token usage was the one runtime fact that reached channels as an `assistant.message` assembled inside the provider: the text the card showed (`Context usage 50% | Token usage: total=...`) was never something the model said, and the counters were embedded in display prose, so any structured consumer had to parse them back out. The numbers were already the native runtime's cumulative session totals passed through verbatim (the runtimes held no baseline and did no differencing); this PR makes that contract explicit on a structured activity and deletes the prose assembly.

## What changed

- `packages/dreamux-types`: new live-only `token.usage` member of both `RuntimeActivity` (camelCase) and `TeammateActivity` (snake_case), carrying the native runtime's **cumulative session totals** `inputTokens`/`outputTokens` plus optional context `{usedTokens, windowTokens}`. Documented as session-total, never turn deltas, never cold-replayed.
- `@excitedjs/agent-runtime-codex` (`turn-manager.ts`): at the existing terminal point, emit one `token.usage` per native turn from the latest `thread/tokenUsage/updated` snapshot (`total` + `last.totalTokens` + `modelContextWindow`); the id convention is unchanged (`\${turnId}:usage`). The snapshot is consumed (cleared) at the terminal, so a same-thread turn with no usage update stays silent instead of repeating the previous totals. `context` is included only when the used footprint and a *positive* window are both present, preserving the historical `Context usage n/a` line when codex has no window; the `usageSummary`/`formatTokenCount` prose assembly is deleted.
- `@excitedjs/agent-runtime-claude-code` (`runtime-activity.ts`): on `result`/`interrupted`, emit one `token.usage` from the native result envelope's `modelUsage` totals (claude accumulates these per session in the resident CLI), id `stream-\${seq}:usage`, with the last main-context footprint as `usedTokens` and a structurally-null window; no provider-side accumulation.
- `dreamux` conversation projection: projects `token.usage` camelCase → snake_case with `redacted: false` (numeric counters carry nothing to redact); the projection and channel-adapter switches are compile-time exhaustive (`never` defaults), so a new union member without an arm is a build error. There is deliberately no run-time unknown-kind path: Core pins the runtimes and the built-in channel through `workspace:*` and they ship in one release, so a deployed process cannot mix activity-union versions.
- `feishu-channel`: the CoT layer now owns the rendering — `acceptTokenUsage` appends the exact historical one-line summary, and the compact-number formatter moves into this package with its format-table tests.
- Rush change files for all five packages (minor, additive activity kind); knowledge pages updated to the new carrier.

Not changed: the `thread/tokenUsage/updated` subscription plumbing, id conventions, terminal ordering (interrupt marker → usage → `turn.ended`), the recent-activity cold reader (usage stays live-only), persisted state, and every other activity kind.

## Semantics

- Zero adapter state beyond "latest snapshot": native CLI/app-server owns accumulation; consumers that need a turn delta difference consecutive snapshots per agent; a dropped snapshot only widens the next delta.
- Additive union member: older edges/server that don't know the kind ignore it, no wire fields removed or retyped.

## Validation

- `rush build --to` all five packages: pass.
- `rush test --to @excitedjs/agent-runtime-codex @excitedjs/agent-runtime-claude-code @excitedjs/feishu-channel`: pass (usage tests rewritten to structured assertions; new `feishu-cot-token-usage.test.ts` owns the rendering table).
- `@excitedjs/dreamux` test suite: non-live suites pass in the sandbox — 1,075 dreamux tests (73 files), 51 codex, 238 claude-code, 636 feishu-channel — including projection exact-object, null/zero-window, stale-snapshot, and activity-to-card delivery tests. The 6 `codex-live.test.ts` real-model tests time out identically on the clean base in this sandbox (no model API reachability); verified by stashing this branch and rerunning.
- eslint on every changed file: clean.

Follow-ups live in downstream integrations (other runtime adapters, consumer-side snapshot differencing); this PR is deliberately limited to the upstream-neutral contract.


packages/agent-runtime/codex/src/turn-manager.ts
```diff
@@ -585,34 +578,33 @@ function restoreCollectedTurn(turn: CollectedTurn, codec: CodexOutputSchemaCodec
...
+  const contextUsed = usage?.last?.totalTokens;
+  const contextWindow = usage?.modelContextWindow;
+  return {
+    kind: 'token.usage',
+    occurredAt: Date.now(),
+    id: `${turnId}:usage`,
+    inputTokens: input,
+    outputTokens: output,
+    context: isTokenCount(contextUsed)
+      ? { usedTokens: contextUsed, windowTokens: isTokenCount(contextWindow) ? contextWindow : null }
```
ryanxiang7: **Major — this shape cannot preserve the codex historical line when no window is known.**

On `next`, codex's `usageSummary()` rendered `Context usage n/a` whenever `modelContextWindow` was missing/non-positive (even with `last.totalTokens` present). Here codex instead emits `{usedTokens: 14_500, windowTokens: null}` — which is byte-identical on the wire to a claude activity, whose window is *always* `null` and whose historical rendering of the same shape is the compact count `14.5k`. The single Feishu renderer therefore now prints `14.5k` for a codex turn that previously printed `n/a`. Verified against both `origin/next` and this branch with `modelContextWindow: null` and `0`.

This fails the task record's own acceptance criterion ("unchanged … percentage, compact count, and `n/a` cases") and the "preserved verbatim" non-goal, and the old codex test that pinned `n/a` was rewritten into the shape assertion above, so CI is green despite the change.

Suggested fix (minimal, restores identical output, adds no new concept):

```ts
context: isTokenCount(contextUsed) && isTokenCount(contextWindow) && contextWindow > 0
  ? { usedTokens: contextUsed, windowTokens: contextWindow }
  : null,
```

The Feishu branch `window_tokens === null → formatTokenCount(used)` then remains the claude-only path, and codex restores `n/a`. Restore a codex end-to-end assertion for the null-window case as part of this.

Alternative: if the new display is actually desired for codex, it needs an explicit operator ruling and updates to `requirement.md`, `technical-design/final.md` §3, `.agents/product/README.md`, and the feishu-channel change file wording — the PR cannot claim byte-identical rendering as written.

packages/channel/feishu-channel/src/feishu-cot-activity.ts
```diff
@@ -121,6 +122,46 @@ export function acceptAssistantMessage(
...
+}
+
+/** Render cumulative counters in the runtime's historical one-line shape. */
+export function tokenUsageSummary(event: CotTokenUsage): string {
+  const context = event.context;
+  let contextUsage = 'n/a';
+  if (context !== null) {
+    contextUsage = context.window_tokens !== null && context.window_tokens > 0
+      ? `${Math.round(context.used_tokens / context.window_tokens * 100)}%`
+      : formatTokenCount(context.used_tokens);
```
ryanxiang7, line 147: This compact-count fallback is correct for claude (window is structurally always `null`, and its old formatter rendered the count here) but is also reached by codex's no-window case, which historically rendered `n/a`. The wire shape gives the renderer no way to tell them apart — the distinction has to be made (or deliberately abandoned with an operator ruling) upstream in the codex mapping. See the comment on `turn-manager.ts`.

packages/agent-runtime/codex/src/turn-manager.ts
```diff
@@ -222,15 +222,8 @@ export class TurnManager {
...
-    const summary = usageSummary(this.tokenUsage);
-    if (summary !== null) {
-      this.emitActivity({
-        kind: 'assistant.message',
-        occurredAt: Date.now(),
-        id: `${turnId}:usage`,
-        text: summary,
-      });
-    }
+    const usage = tokenUsageActivity(turnId, this.tokenUsage);
```
ryanxiang7, line 225: **Minor — a same-thread turn with no usage update re-emits the previous turn's snapshot.** `this.tokenUsage` is reset only on collector/thread change (line 184); it is never cleared after being consumed here. Sequence: turn A completes with snapshot S; turn B on the same thread fails/is interrupted before any `thread/tokenUsage/updated`; this emits S with id `turnB:usage` — counters that were never observed for B.

This predates the PR (the old prose path had the same carryover), but this PR's `verification.md` ("No counters, no activity … failed status / absent envelope") newly claims silence, which is not true for this sequence. Either clear `this.tokenUsage = null` right after emitting here (makes "snapshot for this turn" exact; updates always precede the terminal, so no live ordering is lost), or correct the verification record and add a same-thread missing-update test documenting the intended behavior. Do not add history or differencing.

packages/dreamux/src/channel/conversation-projection.ts
```diff
@@ -201,6 +201,22 @@ function projectedActivity(
...
+        output_tokens: activity.outputTokens,
+        context: activity.context === null
+          ? null
+          : {
+              used_tokens: activity.context.usedTokens,
+              window_tokens: activity.context.windowTokens,
+            },
+        redacted: false,
+      };
+    }
```
ryanxiang7, line 219: **Minor — no exhaustive default means an unknown activity kind publishes `activity: undefined`.** If an independently upgraded provider runtime emits a kind this Core build doesn't know, `projectedActivity()` returns `undefined` and the event is published anyway; the Feishu listener then throws on `event.activity.kind` (caught and logged, so degraded to a warning rather than a turn failure). In a monorepo release this is moot, but the task's design doc claims "a deployment that does not know the kind ignores it" end to end, which only holds for older *channels behind an upgraded Core*, not older Core. Suggest an explicit `default:` that drops/guards the projection (and/or a compile-time `never` assertion), and narrowing the compatibility wording in `technical-design/final.md` §5.

packages/channel/feishu-channel/src/feishu-cot-adapter.ts
```diff
@@ -250,6 +251,9 @@ export class FeishuCotAdapter {
       case 'tool.call':
         acceptToolCallActivity(this.activity, found.key, found.state, event.activity);
         return;
+      case 'token.usage':
```
ryanxiang7, line 254: **Minor — this arm has no test and is not enforced by exhaustiveness.** Deleting `case 'token.usage':` still compiles (the switch has no `never` default) and every test stays green, while the usage line silently stops reaching cards — `feishu-cot-token-usage.test.ts` only calls the pure `tokenUsageSummary`. Please add one adapter-level test through `onActivity(...)` (assert the line lands as an assistant row keyed by `event_id`, opens a card when none is anchored, and is ignored without an anchor), mirroring how the old assistant message behaved. A `default: assertNever(event.activity.kind)` (or equivalent) would also lock future members.

packages/dreamux/tests/cot-projection-privacy.test.ts
```diff
@@ -470,6 +470,43 @@ describe('conversation projection: content visible after redaction is unchanged,
...
+      kind: 'token.usage',
+      occurredAt: Date.now(),
+      id: 'stream-0:usage',
+      inputTokens: 10,
+      outputTokens: 5,
+      context: null,
+    });
+    const usage = activityOf(publisher);
+    expect(usage.kind === 'token.usage' && usage.context).toBeNull();
+    expect(usage.kind === 'token.usage' && usage.redacted).toBe(false);
```
ryanxiang7: **Nit — this assertion is satisfied even when `kind` is wrong**: if `kind !== 'token.usage'`, the `&&` short-circuits to `false`, and `toBe(false)` passes. The first new test already uses a whole-object `toEqual`; assert the full projected object here too (kind, snake_case fields, `context: null`, `redacted: false`).

ryanxiang7 (changes requested):
## Review: `token.usage` activity

The refactor direction is right — getting token counters off a synthetic `assistant.message` and onto a neutral, structured, live-only activity is a good entropy reduction, and the layering is clean: types stay neutral, runtimes do zero accumulation/differencing, projection is numeric-only with `redacted: false`, and prose assembly now lives solely in the Feishu display layer. I independently verified build/test for the affected packages locally (`feishu-channel`, both runtimes, `@excitedjs/dreamux`) — all green.

However, there is one behavior regression against the PR's own stated contract, plus a few smaller items.

### Major

**1. Codex turns without a context window render a different line than before (`n/a` → compact count).** The old codex `usageSummary()` rendered `Context usage n/a` whenever the window was missing/non-positive. The new codex mapping emits `{usedTokens, windowTokens: null}` in exactly that case, and that wire shape is indistinguishable from claude's (whose window is *always* `null` and whose historical rendering is the compact used count), so the single channel renderer now prints e.g. `Context usage 14.5k` for codex. This contradicts the task's own acceptance criterion ("unchanged from the previous build for equal inputs (percentage, compact count, and `n/a` cases)") and non-goal ("the rendered line is preserved verbatim"). It also slipped through because the load-bearing codex `n/a` assertion was rewritten into a shape assertion. Details inline. Recommended minimal fix (preserves output, no new wire concept): codex emits `context: null` unless both `last.totalTokens` and a *positive* `modelContextWindow` are present.

### Minor

**2. Same-thread stale-snapshot carryover contradicts the new verification record.** Pre-existing behavior, but this PR's `verification.md` newly claims both runtimes stay silent when the terminal carries no usage; for codex, turn B failing on the same thread before any `tokenUsage/updated` re-emits turn A's snapshot under `turnB:usage`.

**3. Compatibility claim is broader than the code supports.** An older **Core** receiving the new kind from an independently-upgraded provider returns `undefined` from `projectedActivity()` (no exhaustive default) and publishes an event with `activity: undefined`; the claim should be scoped to older channel consumers behind an upgraded Core.

**4. Nothing tests that `token.usage` actually reaches a card.** The only channel test exercises the pure formatter; deleting the `case 'token.usage'` dispatch arm compiles (no `never` exhaustiveness) and all tests stay green while the line silently vanishes.

### Nits

- The second new projection test (`cot-projection-privacy.test.ts`) is not exact-object: `expect(usage.kind === 'token.usage' && usage.redacted).toBe(false)` also passes when `kind` is wrong. Use a whole-object `toEqual` like the first test.
- `.agents/domains/channel.md` (the "what enters a card" paragraph) now omits the usage line that the channel builds itself; one sentence would keep the KB complete.
- Considered and **not** blocking: typing `redacted` as literal `false`/dropping it. Uniform shape across the activity union is defensible; leave as-is.

Overall the change is close; once the codex no-window parity is resolved (option (a) or an explicit operator ruling plus doc updates) and the test/doc gaps are closed, this is good to merge.

YourWildDad:
Thanks for the careful review — all six findings are addressed in d8fdb60. Local reruns: build, typecheck:tests, lint, and the full non-live suites (1,076 dreamux / 51 codex / 238 claude-code / 636 feishu-channel), plus `check.sh`.

**Major — codex no-window changed `n/a` into a used count**
Took the minimal fix suggested. Codex now omits `context` entirely unless both a used footprint and a *positive* window exist (`turn-manager.ts`, `tokenUsageActivity`), so `{usedTokens, windowTokens: null}` is no longer on codex's wire and its historical line stays `Context usage n/a`. The distinction between "structurally lacks a window" (claude → compact count) and "didn't supply one this turn" (codex → n/a) is carried by context presence, no new field. Restored and strengthened the byte-preservation assertion as exact objects in two cases — `modelContextWindow: null` and `modelContextWindow: 0` ("treats a zero window like a missing one"). The design doc's rendering/compatibility sections now state this explicitly.

**Minor 1 — stale snapshot re-emitted on the next turn**
The snapshot is consumed at the terminal (`this.tokenUsage = null` right after `tokenUsageActivity`), since a tokenUsage notification always precedes its own turn's terminal. New test "does not repeat the previous turn snapshot when the next same-thread turn ends without an update" asserts the exact kind sequence `assistant.message, token.usage, turn.ended, turn.ended` and the absence of `turn-2:usage`. The inaccurate silence claim in the task verification record is corrected.

**Minor 2 — older Core projected an unknown kind to `activity: undefined`**
`projectedActivity` now returns `TeammateActivity | null`; the `default` arm is a compile-time `never` check and, at run time, drops the fact while the caller logs `"Dropping an activity this Core build cannot project"` with dispatcher/agent/kind. New test drives a `token.usage.v2` cast and asserts zero publications plus the warning. Compatibility wording in the design doc is narrowed to the two real boundaries (older channel listener never subscribes; older Core drops with a warning) and notes that mixed provider/Core versions are defensive in this monorepo, not a supported matrix.

**Minor 3 — nothing proved the kind reached a card**
`feishu-cot-delivery.test.ts` now runs two full sessions asserting both rendered strings — `Context usage 50% | Token usage: total=28.6k input=28.6k output=69` and the `n/a` counterpart — actually land on the open COT card. The adapter switch also has a compile-time exhaustive `default`, so deleting the dispatch arm now fails both the tests and the build.

**Nit 1 — projection test asserted two weak predicates**
Replaced with a whole-object `toEqual({ kind, event_id, input_tokens, output_tokens, context: null, redacted: false })`, so a wrong kind or re-keyed field fails it.

**Nit 2 — channel knowledge doc missed the usage line**
The channel domain page's card paragraph now names "the token-usage line the Channel renders itself from a `token.usage` activity" alongside text/tool content.

Re-requesting review.

YourWildDad:
Follow-up on **Minor 2** (1d2c3a3) after an operator review of the fix: the run-time drop arm defended a deployment that does not exist. `@excitedjs/dreamux` pins all runtimes and the built-in channel via `workspace:*`, and the runtimes/channel pin `@excitedjs/dreamux-types` the same way — everything ships in one monorepo release, so a deployed process cannot contain a provider and a Core compiled against different activity unions.

So I removed the run-time `default`/warning/null-return and its test; the projection switch keeps a compile-time `never` exhaustiveness check (`return exhaustive` in the default), and the Feishu adapter's `default` remains compile-time only. The only independently-versioned consumers that can actually exist are external `npm:` channel packages (loader exists, no provider yet), which ignore an unknown tagged member in their own code against their pinned types — nothing this Core needs to detect. Design doc §5/§6 records the ruling and that exact boundary.

Non-live suites: 1,075 dreamux (the removed drop test accounts for −1), 51 codex, 238 claude-code, 636 feishu-channel; build/typecheck:tests/lint clean. The other five findings' fixes are unchanged.

ryanxiang7 (approved):
## Re-review: APPROVE

All six findings from the first round are resolved across d8fdb60 + 1d2c3a3. I re-verified each fix against the code (not just the reply), re-ran the local gates, and checked CI.

### Findings, point by point

1. **Major — codex no-window parity: fixed.** `tokenUsageActivity` now emits `context: null` unless both `last.totalTokens` and a *positive* `modelContextWindow` are present, so codex's historical `n/a` line is restored while claude's structurally-windowless compact count stays distinct — carried by context presence, no new wire field. Exact-object assertions cover both `null` and `0` windows. The full-session delivery test additionally pins the `n/a` line landing on the card. Byte-parity claim now holds for percentage, compact-count, and `n/a`.
2. **Stale same-thread snapshot: fixed.** The snapshot is consumed (`this.tokenUsage = null`) synchronously after the activity is built, before emit — so clearing cannot lose the just-built fact, and a same-thread turn ending without an update emits nothing. New test asserts the exact sequence and the absence of `turn-2:usage`. Clearing happens in a synchronous event callback, so there is no new race; under the protocol's ordering (a turn's `tokenUsage/updated` precedes its terminal) no live update is lost.
3. **Unknown-kind compatibility: fixed per operator ruling.** I independently confirmed the premise: `@excitedjs/dreamux` pins both runtimes, the channel, and `dreamux-types` via `workspace:*`, so a process cannot pair an upgraded provider with an older Core. The remaining compile-time `never` exhaustiveness in both the projection switch and the Feishu adapter is the right guard, and the design doc §5/§6 now records both the ruling and the only real lagging boundary (independently installed external `npm:` channel packages). Removing the runtime drop/warning arm was correct, not a weakening.
4. **Activity-to-card coverage: fixed.** `feishu-cot-delivery.test.ts` drives a full session for both the percentage and `n/a` lines, asserting order and card membership; deleting the adapter dispatch arm now fails both the suite and (via `never`) the build.
5. **Projection assertion nit: fixed.** The null-context case is now a whole-object `toEqual`.
6. **Channel KB nit: fixed.** The "what enters a card" paragraph names the channel-rendered usage line; product catalog and provider-runtime domain also now state the codex/claude context distinction correctly.

### Verification I ran

- `rush build` / `typecheck:tests` / `lint` for the five packages: clean.
- `rush test` for both runtimes, feishu-channel, and `@excitedjs/dreamux`: pass (warnings are the pre-existing real-pipe claude exit-admission tests with no claude binary in the sandbox).
- The changed test files directly: projection privacy 25/25; cot delivery + token usage 27/27.
- `.agents/scripts/check.sh`: task records and reachability pass.
- Head commit CI: all 9 checks green (Ubuntu + macOS rush suites, change-file, gitleaks, KB, internal-content, shellcheck, author metadata).

Layering remains clean and the task records honestly document the corrections, including the rejected intermediate runtime-drop design. No new findings. Approving; merge remains for the operator to authorize separately.

Issue #443, opened 2026-09-17
Give the Dispatcher Agent its own Commands (dispatcher.submit / dispatcher.interrupt)


YourWildDad:
Solution review for the task record `.agents/tasks/architecture/add-dispatcher-submit-command/` (requirement and final solution live there; this Issue is the review surface).

## Requirement (converged)

- Add `dispatcher.submit` and `dispatcher.interrupt` Core Commands. Addressing stays caller context (`dispatcher_id`), never payload.
- `team.submit` and `team.interrupt` require `team_name`; omitting it is `BAD_REQUEST`.
- Feishu: an unbound conversation, a bound conversation whose Team is rejected (`TEAM_NOT_FOUND` / `TEAM_CLOSED`), and a document-comment cold open deliver through `dispatcher.submit` (recipient unchanged: the Dispatcher Agent).
- Feishu `/stop` in an unbound conversation uses `dispatcher.interrupt`; bound conversations keep `team.interrupt`.
- When Collaboration Space automatic provisioning does not deliver the message to its Team, the Channel no longer falls back to the Dispatcher Agent; it replies a failure notice under the triggering message.
- Non-goals: no CLI verb, no per-adapter exposure (the new Commands are reachable over `admin.sock` too), no change to normal provisioning, bound-route reconciliation, or document-comment subscriptions.

## Summary

The Dispatcher Agent gets its own two Commands in the Dispatcher namespace, the
Team Commands stop addressing it, and the Feishu Channel follows: every
Dispatcher-addressed submission goes through `dispatcher.submit`, `/stop` in an
unbound conversation goes through `dispatcher.interrupt`, and a Collaboration
Space provisioning run that does not reach its Team answers the triggering
message with a failure notice instead of falling back to the Dispatcher.

What is removed: the optional-addressing branch in `team.submit` and
`team.interrupt`, the `teamId === null` branch in `DispatcherService.interrupt`,
the provisioning fallback in `FeishuChannel.deliver`, and the production-dead
`service/channel-submission.ts` adapter. What is added: two Command
definitions, one shared submission-payload reader both submit Commands use, one
notice send in the Feishu inbound path.

### `dispatcher.submit` and `dispatcher.interrupt`

Owner: `packages/dreamux/src/service/dispatchers/commands.ts`, beside
`dispatcher.list` / `status` / `start`. The module header changes from
"lifecycle Commands" to "the Dispatcher namespace's Commands": lifecycle plus the
[...]

PR #444, opened 2026-09-18
Give the Dispatcher Agent its own Commands

YourWildDad:
The Dispatcher Agent was reached by omitting `team_name` from a Team Command, so
one payload named two different recipients and a caller that forgot the name
silently woke the Dispatcher. It now has its own two Commands, and the Team
Commands address only Teams.

Requirement and solution: `.agents/tasks/architecture/add-dispatcher-submit-command/`.
Solution review: #443.

## Core

- `dispatcher.submit` submits one turn to the addressed Dispatcher's own Agent;
  `dispatcher.interrupt` interrupts it. Both live in the Dispatcher namespace and
  address their target through the caller context, so neither reads a
  `dispatcher_id` from its payload, and both are reachable over `admin.sock` and
  a Channel's in-process port like every other Command.
- `team.submit` and `team.interrupt` require `team_name`. A missing name is
  answered `BAD_REQUEST` by schema validation before any Agent is reached.
- `DispatcherService.interrupt(teamId | null)` splits into `interruptAgent()` and
  `interruptTeamLeader(teamId)`, mirroring the submit pair.
- The Channel-facing submit payload — safe attribute names, non-empty text, the
  optional trailing reminder, the bounded `source_id`, the fixed `channel`
  provenance — is read in one place, `service/channel-submission.ts`, by both
  Commands. That module previously held an adapter no production path called.
- `dreamux-types` exports `DispatcherSubmitCommand`; `TeamSubmitCommand` extends
  it with a required `team_name` and the optional `intent`.

## Feishu Channel

- An unbound conversation, the single fallback after a bound Team proved it
  admitted nothing, and a document-comment cold open invoke `dispatcher.submit`.
  `/stop` in an unbound conversation invokes `dispatcher.interrupt`.
- A Collaboration Space message whose Team could not be provisioned is no longer
  delivered to the Dispatcher Agent. The Channel replies under the triggering
  message with a fixed notice — `Could not start a Team for this conversation.
  The reason is in the Dreamux log.` — and submits nothing. The notice carries no
  reason because raw failure text can name this host's absolute state paths; the
  reason stays on the delivery log line, which is unchanged. `failed`,
  `ambiguous`, and `error` keep today's handling: logged, no notice.
- A settled question-card answer travels the same delivery entry point. When its
  topic plans provisioning and that run produces no recipient, the answer is
  dropped with its log line rather than notified or re-routed: handing an
  already-given answer to another recipient risks a second turn for one answer.

## Verification

`rush build`, `rush lint`, `rush test`, `rush typecheck:tests`, and
`.agents/scripts/check.sh` pass. Live Codex tests ran; `DREAMUX_SKIP_LIVE_CODEX`
was not used. Acceptance evidence per criterion is in the task's
`verification.md`.

Three ordinary `minor` change notes. No persisted file changes shape, so nothing
here blocks an upgrade.


packages/dreamux/src/service/team-collection/commands.ts
```diff
@@ -209,85 +203,35 @@ export function teamCommands(host: CoreCommandHost): readonly AnyCoreCommand[] {
...
-              // No external submission advances the Dispatcher Agent. Who
-              // waits for a leader's completion is a property of the
-              // operation, not of the adapter that carried it: an Agent
-              // handing work to a Team says so explicitly on the Team MCP
-              // delegate, while an external caller — Channel or `admin.sock`
-              // — is answered by the TeamLeader on its own Channel.
-              deliverCompletionToDispatcher: false,
-            });
+      const admission = await dispatcher.submitToTeamLeader({
+        ...channelSubmitInput(input.command),
```
ryanxiang7, line 225: **Bug: `team.submit` silently drops the optional `intent` payload.**

On `next`, this execute forwarded it explicitly (`shared` had `...(command.intent !== undefined ? { intent: command.intent } : {})`). After the refactor:

- `parseChannelSubmission` returns only `DispatcherSubmitCommand` fields, and `channelSubmitInput` projects only `source/attrs/text/reminder/sourceId` — by design, the shared reader is not supposed to know `intent`;
- but the `team.submit`-specific composition here never adds it back. The spread only gets `teamId` and `deliverCompletionToDispatcher`.

The type checker cannot catch this because `TeammateSubmitInput.intent` is optional. At runtime the payload is accepted and answered `submitted`, while the value never reaches `TeammateService.admitToRuntime` (`teammate-service/index.ts:296-298`, `state.updateIntent(...)`), so the TeamLeader's durable recovery subject is no longer updated by this Command. The MCP delegate path (`mcp-delegate.ts:177-181`) still forwards it, so the two surfaces now diverge.

This also contradicts the reviewed design, which states unchanged behavior ("`team.submit` composes it with `team_name` and `intent`" in the shared-reader section, and "Unchanged: `team.submit` `intent`"). The new type test even constructs a payload with `intent: 'review this'`, but no test drives it through the Command to `submitToTeamLeader`, which is why the regression slipped through.

Suggested fix at the composition seam:

```ts
const admission = await dispatcher.submitToTeamLeader({
  ...channelSubmitInput(input.command),
  teamId: input.command.team_name,
  ...(input.command.intent !== undefined
    ? { intent: input.command.intent }
    : {}),
  deliverCompletionToDispatcher: false,
});
```

plus an adapter-level assertion that a `team.submit` carrying `intent` reaches the fake as the same value (and that `dispatcher.submit` never sets it).

packages/channel/feishu-channel/src/feishu-submit.ts
```diff
@@ -92,10 +92,10 @@ export type FeishuSubmitOutcome =
...
     }
   /**
-   * A proven pre-admission rejection: Core resolved the Team and refused
-   * before creating anything. Because it proves no turn was accepted, it lets
-   * the Channel drop the stale row and deliver the message once to the
-   * Dispatcher Agent instead.
+   * A proven pre-admission rejection of a named Team: Core resolved the Team
+   * and refused before creating anything. Because it proves no turn was
+   * accepted, the delivery path drops the stale row and delivers the message
+   * once to the Dispatcher Agent instead.
```
ryanxiang7: Nit: this variant comment is now context-dependent. A `rejected` outcome falls back to the Dispatcher Agent only on the `bound` plan; on the `provision` plan (a fresh Team refusing its first submission) it is answered in place with the failure notice and reconciles nothing — `feishu-channel-session.test.ts` ('a rejected submission to the just-provisioned Team …') pins that. Consider wording it as 'the `bound` delivery path drops the stale row …' so a reader cannot conclude every rejection falls back.

ryanxiang7 (changes requested):

ryanxiang7 (approved):
第二轮复审通过（Approval）。

本轮 delta 只有声明的四个文件，均已核对：

- `team.submit` execute 在组合处补回 `intent` 条件透传（commands.ts:227-232），写法与 Team MCP delegate 一致；缺省保持缺键；`dispatcher.submit` 不涉及。下游 `TeammateService.admitToRuntime` → `state.updateIntent` 链路恢复，两个入口不再分叉。
- 新增的命令层用例双适配器覆盖：带 intent 的调用在 `submitToTeamLeader` 收到原值且带 `source: 'channel'`、`deliverCompletionToDispatcher: false`；省略时断言无 `intent` 键；`submitToAgent` 零调用。我本地做了独立红绿验证：临时撤掉修复该用例如期失败，恢复后整个测试文件 20 个用例全过。
- nit 已修：`rejected` 注释明确限定为 bound 计划才清路由回落，provision 计划是就地提示、不和解路由。
- verification.md 已如实补记本轮结论；CI 九项全绿。

第一轮核对的其余结论维持不变：deliver 三分支出口完整、notice 触发条件正好是 unsubmitted/rejected、共享 reader 的校验搬迁一字不差、测试整体加强。

packages/dreamux-types/src/team.ts
```diff
@@ -142,6 +136,17 @@ export interface TeamSubmitCommand {
...
+/**
+ * Submit one turn to a Team's TeamLeader.
+ *
+ * A Team Command names exactly one Team; a turn meant for the Dispatcher Agent
+ * uses `dispatcher.submit` instead.
+ */
+export interface TeamSubmitCommand extends DispatcherSubmitCommand {
+  readonly team_name: string;
+  readonly intent?: string;
+}
```
YourWildDad: 这个继承名字很难理解，父类不应该包含dispatcher 这个单词了，应该改成更中性的名字

packages/dreamux/src/service/dispatcher-service/index.ts
```diff
@@ -514,18 +514,15 @@ export class DispatcherService {
...
+
+  /** Interrupt one Team's leader. */
+  interruptTeamLeader(teamId: string) {
     return this.admitOperation(async () =>
-      teamId === null
-        ? this.mustAgent().interrupt()
-        : (await this.teams.open(teamId)).interruptLeader(),
-    );
+      (await this.teams.open(teamId)).interruptLeader());
   }
```
YourWildDad, line 526: 这个玩意太蠢了。为什么一定要在 dispatcherServices 里透传到 teamleader 的 teammate？

Issue #448, opened 2026-09-19
Solution review: transactional store for every runtime store, and agents config Commands


YourWildDad:
This Issue is the review surface for the final technical solution of the task [Add runtime config Commands](https://github.com/excitedjs/dreamux/blob/45e2f0fbc12be908655380293c56077db1edd35e/.agents/tasks/architecture/add-runtime-config-commands/README.md). The body below is [`technical-design/final.md`](https://github.com/excitedjs/dreamux/blob/45e2f0fbc12be908655380293c56077db1edd35e/.agents/tasks/architecture/add-runtime-config-commands/technical-design/final.md) as pushed on branch `dreamux/add-runtime-config-commands`. That file is authoritative; if review changes anything, the file changes first and this Issue is updated after it.

Inputs to review against:

- [Requirement](https://github.com/excitedjs/dreamux/blob/45e2f0fbc12be908655380293c56077db1edd35e/.agents/tasks/architecture/add-runtime-config-commands/requirement.md) (converged)
- [Operator rulings ledger](https://github.com/excitedjs/dreamux/blob/45e2f0fbc12be908655380293c56077db1edd35e/.agents/tasks/architecture/add-runtime-config-commands/rulings.md) (verbatim)
- Independent proposals with their cross-review: [A](https://github.com/excitedjs/dreamux/blob/45e2f0fbc12be908655380293c56077db1edd35e/.agents/tasks/architecture/add-runtime-config-commands/technical-design/proposals/solution-a.md), [B](https://github.com/excitedjs/dreamux/blob/45e2f0fbc12be908655380293c56077db1edd35e/.agents/tasks/architecture/add-runtime-config-commands/technical-design/proposals/solution-b.md), [C](https://github.com/excitedjs/dreamux/blob/45e2f0fbc12be908655380293c56077db1edd35e/.agents/tasks/architecture/add-runtime-config-commands/technical-design/proposals/solution-c.md)

Nothing is implemented yet. Development starts only after this review and the operator's approval.

---

# Final technical solution

The single authoritative solution for this task. It merges the three
independent proposals and their cross-review
([A](https://github.com/excitedjs/dreamux/blob/45e2f0fbc12be908655380293c56077db1edd35e/.agents/tasks/architecture/add-runtime-config-commands/technical-design/proposals/solution-a.md),
[...]

PR #450, opened 2026-09-20
Land the two paused architecture task records

### LINKED ISSUE #449: Land the two paused architecture task records on next
## Why

Two architecture tasks are paused by operator decision, and everything they produced lives only on the branch `dreamux/add-runtime-config-commands`. `next` has none of it. That is fine for a day and bad for a month: the branch drifts, and anyone picking either task up cannot see the requirement they would be building against.

Nothing here is product code. The change is the two task records and the parent index line.

## What lands

**`architecture/add-runtime-config-commands`** — state `blocked`, blocked on development authorization.

- A converged requirement: one transactional store kind every persisted runtime store moves onto, plus a Config Service exposing the `agents` section through two Core Commands.
- `rulings.md`, the operator's decisions quoted rather than paraphrased.
- Three independent solution proposals with a cross-review round.
- `technical-design/final.md`, the adjudicated single solution, reviewed on #448 with the one accepted finding applied.

**`architecture/promote-runtime-model-and-effort`** — state `blocked`, blocked on choosing a solution path.

- A final requirement: `defaultModel` and `defaultEffort` as `agents[]` envelope keys, a neutral capability so Core can carry and change a model and an effort, per-entity persistence in the identity record, and `/model` and `/effort` commands that list, validate, switch and record.
- It also supersedes two lines of #430: the Codex keyword now carries an injected sentence as its own input item, and matching moves from substring to word boundary. The requirement records why #430's reason still holds.

## What this does not do

It does not authorize development on either task, and it merges no implementation. Both records say so themselves: a `blocked` state with a named blocker is the honest state for work that is paused, and the only states that can land on the trunk are `intake`, `blocked` and `done`.

## Checks

`.agents/scripts/check.sh` passes: 50 task records, 299 KB files reachable.

YourWildDad:
Lands the records of two paused architecture tasks on `next`. Documentation only — no product code, no test, no configuration.

Both records are in a trunk state with a named blocker, because both tasks are paused by operator decision:

- `architecture/add-runtime-config-commands` — `blocked` on development authorization. Carries the converged requirement, the rulings ledger, three proposals with their cross-review, and the adjudicated `technical-design/final.md` reviewed on #448.
- `architecture/promote-runtime-model-and-effort` — `blocked` on the solution path. Carries the final requirement for first-class `defaultModel` / `defaultEffort`, the neutral capability, per-entity persistence, and the `/model` and `/effort` commands, including the part that supersedes two lines of #430.

This merges no implementation and authorizes none. #448 stays open: it is that task's solution review surface, and that task is not delivered.

`.agents/scripts/check.sh` passes (50 task records, 299 KB files reachable).

Closes #449


ryanxiang7 (approved):
## Approved

Docs-only review against the two points raised.

**Trunk-state wording.** Both records are `blocked` with a named blocker that stays true after the merge:
- `add-runtime-config-commands` quotes the operator's "暂不开发" (2026-09-20) verbatim, names the unblocking event (a new development-authorization decision), and keeps Development approval / Delivery at "Not granted" / "Not opened" — no in-flight state lands on trunk.
- `promote-runtime-model-and-effort` quotes "先停在这里，今天不开发", again with the exact unblocking event (operator resumes and picks a solution path).
The three 09-21 → 09-20 corrections are consistent across the two READMEs and the rulings ledger heading.

**The #430 supersession.** The rationale holds against current source:
- Codex matches `/ultrathink/i` as a bare case-insensitive substring (`packages/agent-runtime/codex/src/reasoning-effort.ts:26`) and the keyword selects only the turn's effort parameter — nothing is appended to the submission, exactly the two lines of #430 the requirement names.
- The change keeps #430's actual reason (never rewrite the operator's task text) intact: the Claude Code sentence is carried as a separate input item while the assembled submission stays byte for byte, so superseding "nothing is appended" does not overturn its rationale — and the requirement says so explicitly rather than silently drifting.
- `\bultrathink\b` still covers the mixed CJK/English case #430 user story 3 cared about (a CJK character is a non-word boundary); the narrowed forms (`ultrathinking`, `x_ultrathink_y`) are named. The parts of #430 that survive (same rule for all submission sources, no scan of history/files/tool results) are restated as non-goals.

Also verified: `.agents/scripts/check.sh` passes at this head (50 task records, 299 reachable files), all nine CI checks are green, `Closes #449` is correct, and keeping #448 open is correct since that task is undelivered.

PR #451, opened 2026-09-20
Align the Codex ultrathink keyword with Claude Code

YourWildDad:
Aligns the Codex `ultrathink` keyword with the way Claude Code handles the same
word, which is requirement behavior 7 of
[promote model and effort](https://github.com/excitedjs/dreamux/blob/next/.agents/tasks/architecture/promote-runtime-model-and-effort/requirement.md).
The rest of that requirement — `defaultModel`, `defaultEffort`, `/model` and
`/effort` — is not in this change and has no solution yet.

## What changes

- A submission that matches now carries a second input item holding the
  sentence Claude Code injects for this keyword. The submission itself is
  untouched: it is still assembled and sent byte for byte, and the hint travels
  beside it rather than inside it. Codex has no reminder channel — `UserInput`
  is text, images, audio, skills and mentions — so a separate text item is the
  only shape available.
- Matching narrows from the bare substring `/ultrathink/i` to the word boundary
  `/\bultrathink\b/i`. `请ultrathink一下` still matches; `ultrathinking` and
  `prefixULTRATHINKsuffix` no longer do.

Issue #430 chose the substring and chose to append nothing, on the grounds that
the operator's task text must never be rewritten. That reason still holds and
this change respects it — the submission is not edited — but both specifics
were superseded by an operator ruling recorded in the task.

## Shape of the change

`CodexReasoningEffort` owns the whole keyword: the match, the sentence, and the
metadata lookup behind the effort. `effortFor` now answers what one submission
asks for — its effort, and the hint when it is marked — so the rule is still
evaluated once per submission in one place. `submitTurnStart` takes the turn's
text items instead of a single body, which is what lets a turn carry two.

No item type Dreamux renders as activity covers user input, so the added item
does not surface on a card.

## Verification

`rush build`, `rush lint`, `rush test`, `rush typecheck:tests` and
`.agents/scripts/check.sh` all pass. The narrowing is covered by new
non-matching cases; `prefixULTRATHINKsuffix` asserted the opposite before this
change and now asserts that it is left to Codex. No live Codex run exercised
the added input item.

The product catalog entry, the provider-runtime section and the package README
are rewritten in the same change, with an ordinary change note — nothing here
blocks an upgrade.


ryanxiang7 (approved):
## Approved

Reviewed the two flagged areas plus the wiring around them.

**`effortFor` → effort + hint; `submitTurnStart` takes the text items.**
- The keyword has one home: the match, the sentence constant, and the metadata lookup all live in `CodexReasoningEffort`; `effortFor` is still called exactly once per submission from a single call site (`turn-manager.ts`), so the rule is evaluated once, not duplicated at the send site.
- The empty-object early return preserves today's semantics exactly: an unmarked cold submission that previously sent `effort: undefined` still sends no `effort` parameter (the test pins `not.toHaveProperty('effort')`), rather than accidentally sending `effort: undefined`.
- `[text, hint]` ordering is right (task first, instruction about reading it second), and the single-item one-shot helper (`runTurn`, test-only) is updated consistently. Every submission path — including busy appended input — goes through this one `submit()`, so the hint cannot be missed on one path.
- The added item cannot surface on a card or in history: `itemActivity` projects only `agentMessage`, `contextCompaction`, and tool calls; user input items are not a rendered activity kind. That matches the design note, verified in code.

**The word-boundary narrowing and the flipped case.**
- `prefixULTRATHINKsuffix` moved from the matching list to the non-matching list — it is still asserted, with the inverted verdict, rather than deleted. The operator ruling "改成词边界，和 claude 一致" is quoted in the requirement (2026-09-20) and recorded as a deliberate narrowing of #430's user story 3, so this is the requirement changing, not a load-bearing assertion weakened for a green run.
- The narrowing is well-targeted: `请ultrathink一下` / `(ultrathink)` stay matching (CJK and punctuation are non-word boundaries), and the genuinely absorbed forms (`ultrathinking`, `x_ultrathink_y`) stop, each with a case. Case-insensitivity is retained and covered.

Also checked:
- The hint text in the test is spelled out independently, so a reworded constant fails instead of matching itself.
- Change note is an ordinary `minor` with no `BREAKING:` — correct, no persisted shape/path changes; catalog/provider/README all rewritten in the same change; maintenance skill correctly untouched (no config/state moved).
- The task record returns to `blocked` for behaviors 1–6 and names this slice delivered, with the live-codex coverage limitation stated plainly.
- All nine CI checks green and `.agents/scripts/check.sh` passes (50 records, 300 files) at this head.

The recorded coverage limit (wire shape asserted against the fake client only, no live Codex) is the right thing to flag and does not block: the added item is a plain second `{type:'text'}` element of the same `input` array the live tests already exercise, and the failure mode a live run would catch — Codex rejecting the array — is a protocol error that would surface on the first marked submission.

